PR #125, opened 2025-11-30
fix: Ignore 'dbe' in typos check

afadesigns:
Adds  to  to ignore the Double Bracket Expression abbreviation which was triggering false positives.


redteamx (approved):
LGTM.

PR #175, opened 2026-03-30
feat: Add ZC1169 — prefer cp+chmod over install -m

afadesigns:
ZC1169: Flag install -m, suggest cp+chmod. Version 0.1.66 (166 katas)


redteamx (approved):
Reviewed and approved.

PR #305, opened 2026-04-07
feat: Add ZC1278 — use arithmetic expansion instead of expr

afadesigns:
## Summary
- Adds ZC1278: detects `expr` usage and suggests Zsh's native `$(( ))` arithmetic expansion
- Severity: Style
- Version: 0.2.75 (275 katas)

## Test plan
- [x] `go test ./...` passes


redteamx (approved):
LGTM.

PR #306, opened 2026-04-07
feat: Add ZC1279 — use realpath instead of readlink -f

afadesigns:
## Summary
- Adds ZC1279: detects `readlink -f` and suggests `realpath` for better cross-platform portability
- `readlink -f` is not available on macOS without coreutils
- Severity: Info
- Version: 0.2.76 (276 katas)

## Test plan
- [x] `go test ./...` passes


redteamx (approved):
LGTM.

PR #307, opened 2026-04-07
feat: Add ZC1280 — use Zsh ${var:e} instead of cut for file extension

afadesigns:
## Summary
- Adds ZC1280: detects `cut -d. -f2` pattern for extracting file extensions
- Suggests Zsh's `:e` parameter expansion modifier as a built-in alternative
- Severity: Style
- Version: 0.2.77 (277 katas)

## Test plan
- [x] `go test ./...` passes


redteamx (approved):
LGTM.

PR #308, opened 2026-04-07
feat: Add ZC1281 — use sort -u instead of sort | uniq

afadesigns:
## Summary
- Adds ZC1281: detects bare `uniq` usage and suggests `sort -u` for combined sort+dedup
- Skips detection when `uniq` has flags like `-c`, `-d`, `-D` that provide functionality beyond dedup
- Severity: Style
- Version: 0.2.78 (278 katas)

## Test plan
- [x] `go test ./...` passes

### DIFF
--- a/pkg/katas/katatests/zc1281_test.go
+++ b/pkg/katas/katatests/zc1281_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #310, opened 2026-04-10
feat: Support XDG Base Directory Specification

### LINKED ISSUE #309: [FEAT] Support XDG Base Directory Specification
### Request Type

Other

### Description

I propose that the paths searched for the `.zshellcheckrc` file be ammended to the following order:

1. `${XDG_CONFIG_HOME:-${HOME}/.config}/zshellcheck/config.yml` is loaded (if it exists) and it's `disabled_katas` array is appended to the default set.
1. `${HOME}/.zshellcheckrc` is loaded (if it exists) and it's `disabled_katas` array is appended to the current set.
1. `./.zshellcheckrc` is loaded (if it exists) and it's `disabled_katas` array is appended to the current set.

```zsh
echo "disabled_katas: [ZC1049]" > ~/.config/zshellcheck/config.yml

# style: [ZC1049] Prefer functions over aliases.
alias cat=bat
```

```zsh
echo "disabled_katas: [ZC1049]" > ~/.config/zshellcheck/config.yml

# absolutely fine
alias cat=bat
```

### Rationale

Currently, the `.zshellcheckrc` file is only respected in the current directory, with no command line flag to specify an alternative path. Many (most?) CLI tools look for their dotfiles in the user's home directory, as well as the current directory, although this leads to a messy `$HOME`. Using the widely supported[^1] XDG standard, would allow the user to have a single file that has rules they might want to disable, and then they wouldn't need to have an identical `.zshellcheckrc` in each project, or be forced to run `zshellcheck` in specific directories to have the correct set of katas.

[^1]: https://wiki.archlinux.org/title/XDG_Base_Directory#Supported

### Additional Context

In full disclosure, while shellcheck[^2] does support the XDG standard, they use the `~/.config/shellcheckrc` path instead of the more idiomatic `~/.config/shellcheck/shellcheckrc`. I've suggested the path `~/.config/zshellcheck/config.yml` but as this is a shellcheck-adjacent project, you may want to consider the shorter `~/.config/zshellcheckrc` instead, although it means that the `~/.config/` dir gets polluted just like `$HOME` would be. My preference (and common practice) is to have a separate dir under `~/.config/`, but as I said that is a departure from shellcheck.

[^2]: https://github.com/koalaman/shellcheck/wiki/Directive#shellcheckrc-file

Edit: It has occured to me that `~/.config/zshellcheck/config.yaml` is probably a better path to avoid the stuttering `zshellcheck/zshellcheck` bit.

norwd:
## Description

This PR adds support for the XDG Base Directory Spec.

1. `~/.config/zshellcheck/config.yml` is loaded if it exists
1. `~/.zshellcheckrc` is merged if it exists
1. `./.zshellcheckrc` is merged if it exists

The implementation is backwards compatible, the existing tests still pass.

Note, I noticed when running the `integration_test.zsh` script that any katas disabled in the `config.yml` file will show as failed. In may be good to add a `--norc` flag to be used in cases like this, although this is not an issue in CI or on machines where no custom files exist.

Fixes #309

## Type of change

- [x] `feat`: New feature (non-breaking change which adds functionality)
- [ ] `fix`: Bug fix (non-breaking change which fixes an issue)
- [ ] `docs`: Documentation update
- [ ] `chore`: Maintenance (deps, build, etc.)
- [ ] `refactor`: Code restructuring
- [ ] `test`: Adding missing tests

## Checklist

- [x] I have read the [CONTRIBUTING.md](CONTRIBUTING.md) guide.
- [x] I have read the [DEVELOPMENT.md](DEVELOPMENT.md) guide.
- [x] **Linting**: My code passes `go fmt` and `go vet`.
- [ ] **Tests**: I have added tests for my changes (especially for new Katas).
- [x] **Integration**: I have run `./tests/integration_test.zsh` and it passes.
- [x] **Documentation**: I have updated relevant documentation (if applicable).

## For New Katas (If Applicable)

_**N/A**_

- [ ] Added `pkg/katas/zcXXXX.go`
- [ ] Registered in `pkg/katas/katas.go`
- [ ] Added tests in `pkg/katas/katatests/zcXXXX_test.go`
- [ ] Verified that the ID (`ZCXXXX`) is unique and sequential.

### DIFF
--- a/cmd/zshellcheck/main.go
+++ b/cmd/zshellcheck/main.go
@@ -77,9 +77,6 @@ func run() int {
 		fmt.Fprint(os.Stderr, config.Banner)
 	}
 
-	// Start with default config
-	config := config.DefaultConfig()
-
 	configFileXdgConfigPath, _ := xdg.ConfigFile("zshellcheck/config.yaml")
 	configFileHomePath := filepath.Join(xdg.Home, ".zshellcheckrc")
 	config, err := loadConfig(configFileXdgConfigPath, configFileHomePath, ".zshellcheckrc")
--- a/cmd/zshellcheck/main.go
+++ b/cmd/zshellcheck/main.go
@@ -77,7 +77,7 @@ func run() int {
 		fmt.Fprint(os.Stderr, config.Banner)
 	}
 
-	configFileXdgConfigPath, _ := xdg.ConfigFile("zshellcheck/config.yaml")
+	configFileXdgConfigPath, _ := xdg.SearchConfigFile("zshellcheck/config.yaml")
 	configFileHomePath := filepath.Join(xdg.Home, ".zshellcheckrc")
 	config, err := loadConfig(configFileXdgConfigPath, configFileHomePath, ".zshellcheckrc")
 	if err != nil {
--- a/cmd/zshellcheck/main.go
+++ b/cmd/zshellcheck/main.go
@@ -77,7 +77,12 @@ func run() int {
 		fmt.Fprint(os.Stderr, config.Banner)
 	}
 
-	configFileXdgConfigPath, _ := xdg.SearchConfigFile("zshellcheck/config.yaml")
+	// Pick either config.yml or config.yaml, but not both
+	configFileXdgConfigPath, err := xdg.SearchConfigFile("zshellcheck/config.yml")
+	if err != nil {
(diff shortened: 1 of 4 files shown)


afadesigns:
Superseded by #1034, which rebases this branch onto main and adds the `go mod tidy`, empty-path guard, and 3 unit tests the original review asked for. Thanks @norwd for the initial implementation — the authorship chain is preserved on the new PR.

PR #312, opened 2026-04-13
feat: Add ZC1282 — use Zsh ${var:r} instead of sed to remove file extension

afadesigns:
## Summary
- Adds ZC1282: detects `sed 's/\.[^.]*$//'` patterns for stripping file extensions
- Recommends using Zsh built-in `${var:r}` modifier instead
- Severity: style

## Test plan
- [x] Unit tests pass for valid and invalid cases
- [x] Full test suite passes
- [x] Lint clean

### DIFF
--- a/pkg/katas/katatests/zc1282_test.go
+++ b/pkg/katas/katatests/zc1282_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #313, opened 2026-04-13
feat: Add ZC1283 — use setopt instead of set -o for Zsh options

afadesigns:
## Summary
- Adds ZC1283: detects `set -o <option>` and recommends native `setopt` instead
- Zsh-specific idiom — `setopt`/`unsetopt` are the native Zsh builtins
- Severity: style

## Test plan
- [x] Full test suite passes
- [x] Lint clean


redteamx (approved):
LGTM.

PR #314, opened 2026-04-13
feat: Add ZC1284 — use Zsh ${(s:sep:)var} instead of cut -d for splitting

afadesigns:
## Summary
- Adds ZC1284: detects `cut -d<delim> -f<N>` patterns (excluding `-d.` which is ZC1280)
- Recommends Zsh native `${(s:sep:)var}` parameter expansion flag for splitting
- Severity: style

## Test plan
- [x] Full test suite passes
- [x] Lint clean

### DIFF
--- a/pkg/katas/katatests/zc1284_test.go
+++ b/pkg/katas/katatests/zc1284_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #315, opened 2026-04-13
feat: Add ZC1285 — use Zsh ${(o)array} for sorting instead of sort

afadesigns:
## Summary
- Adds ZC1285: detects bare `sort <file>` without complex flags
- Recommends Zsh native `${(o)array}` parameter flag for in-shell sorting
- Severity: style

## Test plan
- [x] Full test suite passes
- [x] Lint clean


redteamx (approved):
LGTM.

PR #316, opened 2026-04-13
feat: Add ZC1286 — use Zsh ${array:#pattern} instead of grep -v

afadesigns:
## Summary
- Adds ZC1286: detects `grep -v` usage for filtering
- Recommends Zsh native `${array:#pattern}` parameter expansion
- Severity: style

## Test plan
- [x] Full test suite passes
- [x] Lint clean


redteamx (approved):
LGTM.

PR #317, opened 2026-04-13
feat: Add ZC1287 — use Zsh ${(V)var} instead of cat -v

afadesigns:
## Summary
- Adds ZC1287: detects `cat -v` and `cat -A` for displaying control characters
- Recommends Zsh native `${(V)var}` parameter expansion flag
- Severity: style

## Test plan
- [x] Full test suite passes
- [x] Lint clean


redteamx (approved):
LGTM.

PR #318, opened 2026-04-13
feat: Add ZC1288 — use typeset instead of declare in Zsh scripts

afadesigns:
## Summary
- Adds ZC1288: detects `declare` usage in Zsh scripts
- Recommends native `typeset` as the idiomatic Zsh builtin
- Severity: style

## Test plan
- [x] Full test suite passes
- [x] Lint clean


redteamx (approved):
LGTM.

PR #319, opened 2026-04-13
feat: Add ZC1289 — use Zsh ${(u)array} for unique elements instead of sort -u

afadesigns:
## Summary
- Adds ZC1289: detects `sort -u` without other complex flags
- Recommends Zsh native `${(u)array}` parameter flag for deduplication
- Preserves original order unlike `sort -u`
- Severity: style

## Test plan
- [x] Full test suite passes
- [x] Lint clean


redteamx (approved):
LGTM.

PR #321, opened 2026-04-13
feat: Add ZC1291 — use Zsh ${(O)array} for reverse sorting instead of sort -r

afadesigns:
## Summary
- Adds ZC1291: detects `sort -r` without other complex flags
- Recommends Zsh native `${(O)array}` parameter flag for descending sort
- Severity: style

## Test plan
- [x] Full test suite passes
- [x] Lint clean


redteamx (approved):
LGTM.

PR #322, opened 2026-04-13
feat: Add ZC1292 — use Zsh ${var//old/new} instead of tr for char translation

afadesigns:
## Summary
- Adds ZC1292: detects `tr` with single-character arguments
- Recommends Zsh native `${var//old/new}` parameter expansion
- Does not conflict with ZC1277 which handles case conversion ranges
- Severity: style

## Test plan
- [x] Full test suite passes
- [x] Lint clean


redteamx (approved):
LGTM.

PR #323, opened 2026-04-13
feat: Add ZC1293 — use [[ ]] instead of test command in Zsh

afadesigns:
## Summary
- Adds ZC1293: detects `test` command usage
- Recommends Zsh native `[[ ]]` conditional syntax
- `[[ ]]` supports pattern matching, regex, and doesn't need variable quoting
- Severity: style

## Test plan
- [x] Full test suite passes
- [x] Lint clean


redteamx (approved):
LGTM.

PR #324, opened 2026-04-13
feat: Add ZC1294 — use bindkey instead of bind for Zsh key bindings

afadesigns:
## Summary
- Adds ZC1294: detects Bash `bind` command usage in Zsh scripts
- Recommends Zsh native `bindkey` for ZLE key bindings
- Severity: warning (bind will fail in pure Zsh)

## Test plan
- [x] Full test suite passes
- [x] Lint clean


redteamx (approved):
LGTM.

PR #325, opened 2026-04-13
feat: Add ZC1295 — use vared instead of read -e for interactive editing

afadesigns:
## Summary
- Adds ZC1295: detects `read -e` (Bash readline editing flag)
- Recommends Zsh native `vared` for interactive variable editing with ZLE
- Severity: style

## Test plan
- [x] Unit tests pass
- [x] Full test suite passes
- [x] Lint clean


redteamx (approved):
LGTM.

PR #326, opened 2026-04-13
feat: Add ZC1296 — avoid shopt in Zsh, use setopt/unsetopt instead

afadesigns:
## Summary
- Adds ZC1296: detects `shopt` (Bash-only builtin) in Zsh scripts
- Recommends `setopt`/`unsetopt` as the native Zsh equivalents
- Severity: warning (shopt does not exist in Zsh)

## Test plan
- [x] Unit tests pass
- [x] Full test suite passes
- [x] Lint clean

### DIFF
--- a/pkg/katas/katatests/zc1296_test.go
+++ b/pkg/katas/katatests/zc1296_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #328, opened 2026-04-13
feat: Add ZC1298 — avoid $FUNCNAME in Zsh, use $funcstack

afadesigns:
## Summary
- Adds ZC1298: detects `$FUNCNAME` usage in Zsh scripts
- Recommends `$funcstack` as the native Zsh call stack array
- Severity: warning (FUNCNAME is undefined in Zsh)

## Test plan
- [x] Unit tests pass
- [x] Full test suite passes
- [x] Lint clean

### DIFF
--- a/pkg/katas/katatests/zc1298_test.go
+++ b/pkg/katas/katatests/zc1298_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #348, opened 2026-04-16
fix: infrastructure hardening — fix release workflow, install.sh, security scans, 95% coverage

### LINKED ISSUE #311: [BUG] The install.sh script fails when Go is not already installed
### Describe the bug

As far as I can tell, this is because the `/repos/afadesigns/zshellcheck/releases/latest` endpoint returns 404 (due to there being no releases, only tags). The `cleanup` function is run because of the trap, but the error message is not shown as the `grep` is failing, while the error message is only run if `LATEST_TAG` is empty.

https://github.com/afadesigns/zshellcheck/blob/f85704f3f574880d634cb3ccb215a74d4f291a95/install.sh#L207-L216

### Reproduction Code

```zsh
echo "N/A"
```

### Actual Output

```text
echo "N/A"
```

### Expected Output

At minimum the script should correctly report the error, currently it fails silently and is difficult to debug.

```sh
$ ./install.sh
Failed to fetch latest release info from GitHub.
```

Better yet, the script should install `main` until a major release is reached.

### ZShellCheck Version

f85704f3f574880d634cb3ccb215a74d4f291a95

### Operating System

macOS

### Additional Context

The easiest way to reproduce this is to run `curl -sSL https://github.com/afadesigns/zshellcheck/raw/refs/heads/main/install.sh | sh -x` in a clean VM / Docker / GitHub Action.

afadesigns:
## Summary

Infrastructure hardening pass ahead of the public beta. Four focused commits:

1. **Fix release workflow** — install syft + cosign, include man page/completions/CHANGELOG in archives
2. **Sync version docs** — README, USER_GUIDE, ROADMAP all updated from stale v0.1.66/166 to v0.3.35/335
3. **Fix install.sh** (closes #311) — graceful HTTP error handling, /tags fallback when no releases exist
4. **Security stack** — new SBOM / OSV-Scanner / Trivy workflows, nightly fuzz with deeper budget, codecov tightened from 80% to 95%

## Why now

- Every Release workflow run since goreleaser was configured has failed silently because syft was never installed. This PR fixes that.
- Users hitting install.sh today get no error message when the installer can't find a release — just a silent exit. Now they get specific guidance.
- Security posture for the upcoming v0.3.35 public beta needs OSV-Scanner + Trivy + fuzz-nightly for Scorecard ≥9.5 and SLSA L3 practices.

## Test plan

- [x] `go test ./...` — all packages green locally
- [x] `golangci-lint` — clean
- [x] Trace scan — zero matches in tracked tree
- [x] Local `install.sh` dry run — graceful fallback to /tags verified
- [ ] CI checks green on this PR
- [ ] Next tag push successfully produces a release with attached binaries + SBOMs + provenance


redteamx (approved):
LGTM.

redteamx (approved):
LGTM.

redteamx (approved):
LGTM.

PR #351, opened 2026-04-16
fix: install.sh substring checksum match picks up sibling files

afadesigns:
## Summary

Post-release smoke test of v0.3.35 exposed a checksum verification failure:

```
sha256sum: zshellcheck_Linux_x86_64.tar.gz.sbom.json: No such file or directory
Checksum verification failed!
```

The release archive layout now includes per-archive siblings (.sbom.json, .sig, .pem) and the installer's `grep "$FILENAME" checksums.txt` selects those lines too because the filename is a substring prefix.

## Fix

Match the exact filename at field 2 of the checksum line via `awk`. No false-positive matches, no over-selection, and more robust to future sibling additions.

## Test plan

- [x] Manual: `bash install.sh -y -v v0.3.35` now succeeds (verified locally against the tagged release archives)
- [ ] CI green

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -285,16 +285,20 @@ if [ "$BUILD_SUCCESS" = false ]; then
 
     # Download Binary
     if curl -sL -o "$FILENAME" "$URL"; then
-        # Verify
+        # Verify — match the exact filename at end-of-line so sibling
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #352, opened 2026-04-16
feat: ZC1339 — use `${#${(f)var}}` instead of `wc -l` for line count

afadesigns:
ZC1339 — Use Zsh `${#${(f)var}}` instead of `wc -l` for line count

What: flags `wc -l` invocations with no file argument (so it reads from stdin in a pipeline).
Why: Zsh's `(f)` parameter flag splits on newlines; `${#array}` counts elements. The combination counts lines natively without spawning an external process.
Fix suggestion: `${#${(f)var}}`
Severity: Style

Completes the `wc` family alongside ZC1081 (`-c` → `${#var}`) and ZC1185 (`-w` → `${#${(z)var}}`).

### DIFF
--- a/pkg/katas/katatests/zc1339_test.go
+++ b/pkg/katas/katatests/zc1339_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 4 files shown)


redteamx (approved):
LGTM.

redteamx (approved):
LGTM.

PR #354, opened 2026-04-16
feat: ZC1341 — use Zsh `*(.x)` glob qualifier instead of `find -executable`

afadesigns:
ZC1341 — Use Zsh `*(.x)` glob qualifier instead of `find -executable`

What: flags any `find ... -executable` invocation.
Why: Zsh's glob qualifier `*(.x)` restricts expansion to regular files (`.`) that have the executable bit (`x`), covering the common use case without an external process. For recursive search use `**/*(.x)` (with extended_glob).
Fix suggestion: `for f in *(.x); do ...; done` or `scripts=(**/*(.x))`
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1341_test.go
+++ b/pkg/katas/katatests/zc1341_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #355, opened 2026-04-17
feat: ZC1342 — use Zsh `*(L0)` glob qualifier instead of `find -empty`

afadesigns:
ZC1342 — Use Zsh `*(L0)` glob qualifier instead of `find -empty`

What: flags any `find ... -empty` invocation.
Why: Zsh's `*(L0)` glob qualifier matches files whose length is zero; combined with `.` it restricts to regular files, and with `/` to directories. Same selection, no external process.
Fix suggestion: `*(L0)` (any), `*(.L0)` (regular files), `*(/L0)` (directories).
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1342_test.go
+++ b/pkg/katas/katatests/zc1342_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #356, opened 2026-04-17
feat: ZC1343 — use Zsh glob age qualifiers instead of `find -mtime`

afadesigns:
ZC1343 — Use Zsh `*(m±N)` glob qualifier instead of `find -mtime N`

What: flags `find` with age predicates (`-mtime`, `-mmin`, `-atime`, `-amin`, `-ctime`, `-cmin`).
Why: Zsh glob qualifiers encode the same predicates (`*(mN)`, `*(m+N)`, `*(m-N)` for modification in days; `*(M±N)` for minutes; `*(a±N)` access; `*(c±N)` status-change) without shelling out.
Fix suggestion: `files=(**/*(m-7))` (modified in last 7 days), `old=(**/*(m+30))` (older than 30), etc.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1343_test.go
+++ b/pkg/katas/katatests/zc1343_test.go
@@ -1,0 +1,65 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #357, opened 2026-04-17
feat: ZC1344 — use Zsh `*(L±Nk)` glob qualifier instead of `find -size`

afadesigns:
ZC1344 — Use Zsh `*(L±Nk)` glob qualifier instead of `find -size`

What: flags any `find ... -size` invocation.
Why: Zsh's `*(LN)` glob qualifier matches files by size; unit suffix (`k`, `m`, `p`) gives byte granularity; `+`/`-` prefixes for larger/smaller. Same expressive power as `find -size` without an external process.
Fix suggestion: `big=(**/*(L+10m))` (larger than 10 MB), `tiny=(**/*(L-1k))` (smaller than 1 KB).
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1344_test.go
+++ b/pkg/katas/katatests/zc1344_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #358, opened 2026-04-17
feat: ZC1345 — use Zsh `*(f:mode:)` glob qualifier instead of `find -perm`

afadesigns:
ZC1345 — Use Zsh `*(f:mode:)` glob qualifier instead of `find -perm`

What: flags any `find ... -perm` invocation.
Why: Zsh glob qualifier `*(f:mode:)` accepts octal (`*(f:0755:)`) or symbolic (`*(f:u+x:)`) mode expressions and is evaluated entirely in the shell.
Fix suggestion: `bins=(**/*(f:u+x:))` (has owner-executable bit), `secret=(**/*(f:0600:))`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1345_test.go
+++ b/pkg/katas/katatests/zc1345_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #359, opened 2026-04-17
feat: ZC1346 — use Zsh `*(u:name:)` glob qualifier instead of `find -user`

afadesigns:
ZC1346 — Use Zsh `*(u:name:)` glob qualifier instead of `find -user`

What: flags `find` with `-user`, `-uid`, or `-nouser`.
Why: Zsh glob qualifiers encode ownership: `*(u:name:)` by name, `*(u+uid)` by numeric uid, `*(U)` for current user. No external process.
Fix suggestion: `mine=(**/*(U))`, `alice=(**/*(u:alice:))`, `numeric=(**/*(u+1000))`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1346_test.go
+++ b/pkg/katas/katatests/zc1346_test.go
@@ -1,0 +1,65 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #360, opened 2026-04-17
feat: ZC1347 — use Zsh `*(g:name:)` glob qualifier instead of `find -group`

afadesigns:
ZC1347 — Use Zsh `*(g:name:)` glob qualifier instead of `find -group`

What: flags `find` with `-group`, `-gid`, or `-nogroup`.
Why: Zsh glob qualifiers encode group ownership: `*(g:name:)`, `*(g+gid)`, `*(G)` for current user's primary group.
Fix suggestion: `mygroup=(**/*(G))`, `wheel=(**/*(g:wheel:))`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1347_test.go
+++ b/pkg/katas/katatests/zc1347_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #361, opened 2026-04-17
feat: ZC1348 — use Zsh glob type qualifiers instead of `find -type`

afadesigns:
ZC1348 — Use Zsh glob type qualifiers instead of `find -type`

What: flags `find` with `-type`.
Why: Zsh's glob qualifiers map directly to file-type predicates: `*(/)` directories, `*(.)` regular files, `*(@)` symlinks, `*(=)` sockets, `*(p)` named pipes, `*(*)` executable regular files, `*(%)` block/char devices.
Fix suggestion: `files=(**/*(.))`, `dirs=(**/*(/))`, `links=(**/*(@))`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1348_test.go
+++ b/pkg/katas/katatests/zc1348_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #362, opened 2026-04-17
feat: ZC1349 — use `${#var}` instead of `expr length` for string length

afadesigns:
ZC1349 — Use `${#var}` instead of `expr length "$var"` for string length

What: flags `expr length ...` invocations.
Why: `${#var}` returns string length natively without spawning `expr`. Works in Zsh and POSIX alike.
Fix suggestion: `len=${#var}`.
Severity: Style


redteamx (approved):
LGTM.

PR #363, opened 2026-04-17
feat: ZC1350 — use `${str:pos:len}` instead of `expr substr`

afadesigns:
ZC1350 — Use `${str:pos:len}` instead of `expr substr` for substring extraction

What: flags `expr substr ...` invocations.
Why: Parameter expansion `${str:pos:len}` does the same work natively. Supports negative `pos` for from-end offsets; omit `:len` for "to end of string".
Fix suggestion: `chunk=${s:1:5}`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1350_test.go
+++ b/pkg/katas/katatests/zc1350_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #364, opened 2026-04-17
feat: ZC1351 — use `[[ =~ ]]` instead of `expr match`/`expr index`

afadesigns:
ZC1351 — Use `[[ $str =~ pattern ]]` for regex instead of `expr match` / `expr index`

What: flags `expr match` and `expr index` invocations.
Why: `[[ $str =~ pattern ]]` evaluates regex natively in Zsh and populates `$match`, `$MATCH`, `$mbegin`, `$mend` arrays.
Fix suggestion: `[[ $str =~ '(foo)([0-9]+)' ]] && echo "${match[1]} ${match[2]}"`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1351_test.go
+++ b/pkg/katas/katatests/zc1351_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #365, opened 2026-04-17
feat: ZC1352 — avoid `xargs -I{}` — use Zsh `for` loop

afadesigns:
ZC1352 — Avoid `xargs -I{}` — use a Zsh `for` loop for per-item substitution

What: flags `xargs -I` / `xargs --replace` invocations.
Why: Per-item substitution with `-I{}` runs a fresh process per item, and variable mutations inside the command never reach the outer shell. A Zsh for loop over `${(f)"$(cmd)"}` keeps everything in-process.
Fix suggestion: `for x in ${(f)"$(cmd)"}; do ...; done`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1352_test.go
+++ b/pkg/katas/katatests/zc1352_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

redteamx (approved):
LGTM.

PR #366, opened 2026-04-17
feat: ZC1353 — avoid `printf -v` — Bash-only, use `print -v` or command substitution

afadesigns:
ZC1353 — Avoid `printf -v` — use `print -v` or command substitution

What: flags `printf -v ...` invocations.
Why: `printf -v var fmt args` is a Bash-ism; POSIX printf silently ignores `-v`, producing surprising behavior. Zsh's `print -v var -rf fmt args` or plain `var=$(printf fmt args)` both work.
Fix suggestion: `print -v line -rf '%d' 42` or `line=$(printf '%d' 42)`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1353_test.go
+++ b/pkg/katas/katatests/zc1353_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #367, opened 2026-04-17
feat: ZC1354 — use `whence -w`/`whence -a`/`whence -p` instead of `type -t`/`-a`/`-P`

afadesigns:
ZC1354 — Use Zsh `whence -w` instead of Bash-specific `type -t`

What: flags `type -t`, `type -a`, `type -P` invocations.
Why: These `type` flags are Bash extensions. Zsh provides the same information via `whence -w` (category), `whence -a` (all matches), `whence -p` (path only) — portable Zsh, no external fallback needed.
Fix suggestion: `whence -w grep` instead of `type -t grep`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1354_test.go
+++ b/pkg/katas/katatests/zc1354_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #368, opened 2026-04-17
feat: ZC1355 — use `print -r` instead of `echo -E` for raw output

afadesigns:
ZC1355 — Use `print -r` instead of `echo -E` for raw output

What: flags `echo -E ...` invocations.
Why: `echo -E` (disable backslash interpretation) is a Bash-ism; POSIX echo silently ignores `-E`. Zsh's `print -r` is the idiomatic raw-printer and works in all cases.
Fix suggestion: `print -r -- "$line"`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1355_test.go
+++ b/pkg/katas/katatests/zc1355_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #369, opened 2026-04-17
feat: ZC1356 — use `read -A` (Zsh) instead of `read -a` (Bash)

afadesigns:
ZC1356 — Use `read -A` instead of `read -a` for array read

What: flags `read -a` invocations.
Why: Zsh's `read -A` (uppercase) reads into an array; `read -a` has different semantics in Zsh than in Bash, producing wrong results when Bash scripts are run under Zsh without adjustment.
Fix suggestion: `read -A arr <<< "$input"`.
Severity: Error (produces wrong behavior, not a style nit)

### DIFF
--- a/pkg/katas/katatests/zc1356_test.go
+++ b/pkg/katas/katatests/zc1356_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #370, opened 2026-04-17
feat: ZC1357 — use Zsh `${(q)var}` instead of `printf '%q'`

afadesigns:
ZC1357 — Use Zsh `${(q)var}` instead of `printf '%q'` for shell-quoting

What: flags `printf` with a format string containing `%q`.
Why: Bash's `%q` format specifier is a Bash extension; POSIX printf ignores it. Zsh's `${(q)var}` parameter flag produces shell-quoted output natively, with `${(qq)}`, `${(qqq)}`, `${(qqqq)}` for single-, double-, and `$'...'` styles.
Fix suggestion: `quoted=${(q)line}`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1357_test.go
+++ b/pkg/katas/katatests/zc1357_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #371, opened 2026-04-17
feat: ZC1358 — use `${PWD:P}` instead of `pwd -P` for canonical path

afadesigns:
ZC1358 — Use `${PWD:P}` instead of `pwd -P` for physical current directory

What: flags `pwd -P` invocations.
Why: Zsh's `P` filename modifier resolves symlinks and returns the canonical path. `${PWD:P}` does the same as `pwd -P` without spawning the external.
Fix suggestion: `canonical=${PWD:P}`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1358_test.go
+++ b/pkg/katas/katatests/zc1358_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #372, opened 2026-04-17
feat: ZC1359 — use Zsh `$groups` instead of `id -Gn`/`-G`/`-gn`/`-g`

afadesigns:
ZC1359 — Use Zsh `$groups` associative array instead of `id -Gn`

What: flags `id -Gn`, `-G`, `-gn`, `-g`.
Why: Zsh's `zsh/parameter` module exposes `$groups` (names→gids), `$GID` (primary gid), and `$USERNAME`/`$LOGNAME` for user info. No external `id` call needed.
Fix suggestion: `zmodload zsh/parameter; print -l ${(k)groups}` or `echo $GID`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1359_test.go
+++ b/pkg/katas/katatests/zc1359_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #373, opened 2026-04-17
feat: ZC1360 — use Zsh `*(OL)` glob qualifier instead of `ls -S`

afadesigns:
ZC1360 — Use Zsh `*(OL)` glob qualifier instead of `ls -S` for size-ordered listing

What: flags `ls -S`, `-Sr`, `-lS`, `-lSr`.
Why: Zsh glob qualifier `*(OL)` orders by size descending, `*(oL)` ascending. Combined with `[N]` subscript to pick N-th item. No external `ls` spawn.
Fix suggestion: `biggest=(*(OL[1]))`, `smallest=(*(oL[1]))`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1360_test.go
+++ b/pkg/katas/katatests/zc1360_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #374, opened 2026-04-17
feat: ZC1361 — avoid `awk 'NR==N'` — use Zsh array subscript

afadesigns:
ZC1361 — Avoid `awk 'NR==N'` — use Zsh array subscript on `${(f)...}`

What: flags `awk` programs containing `NR==N` (pick N-th line).
Why: Zsh can split on newlines natively with `${(f)"$(<file)"}`, producing an array that is directly indexable. Skips awk process startup for a trivial operation.
Fix suggestion: `lines=(${(f)"$(<file)"}); print $lines[5]`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1361_test.go
+++ b/pkg/katas/katatests/zc1361_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #375, opened 2026-04-17
feat: ZC1362 — use `[[ -o option ]]` instead of `test -o option`

afadesigns:
ZC1362 — Use `[[ -o option ]]` instead of `test -o option` for Zsh option checks

What: flags `test -o` / `[ -o` invocations.
Why: In Zsh, `[[ -o option ]]` queries a shell option. The external `test` builtin interprets `-o` as logical OR — producing wrong results or syntax errors for option checks.
Fix suggestion: `[[ -o noglob ]] && ...`
Severity: Info (semantic gotcha when porting from `test` expressions)

### DIFF
--- a/pkg/katas/katatests/zc1362_test.go
+++ b/pkg/katas/katatests/zc1362_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #376, opened 2026-04-17
feat: ZC1363 — use Zsh `*(e:...:)` eval qualifier instead of `find -newer`

afadesigns:
ZC1363 — Use Zsh `*(e:...:)` eval qualifier instead of `find -newer`/`-older`

What: flags `find` with `-newer`, `-anewer`, `-cnewer`, `-newerXY`.
Why: Zsh's `*(e:'expr':)` evaluates an arbitrary expression per match — covers the `-newer` family (`[[ $REPLY -nt ref ]]`), `-older` (`-ot`), and more complex custom predicates without spawning find.
Fix suggestion: `files=(**/*(e:'[[ $REPLY -nt reference ]]':))`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1363_test.go
+++ b/pkg/katas/katatests/zc1363_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #377, opened 2026-04-17
feat: ZC1364 — use Zsh `${var:pos:len}` instead of `cut -c`

afadesigns:
ZC1364 — Use Zsh `${var:pos:len}` instead of `cut -c` for character ranges

What: flags `cut -c ...` and `cut -c1-5` attached-form invocations.
Why: `${var:pos:len}` (zero-indexed position, length count) extracts the same substring natively. No external `cut` process for character-range selection.
Fix suggestion: `chunk=${line:0:5}` replaces `echo "$line" | cut -c 1-5`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1364_test.go
+++ b/pkg/katas/katatests/zc1364_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #378, opened 2026-04-17
feat: ZC1365 — use Zsh `zstat` module instead of `stat -c`

afadesigns:
ZC1365 — Use Zsh `zstat` module instead of `stat -c` for file metadata

What: flags `stat -c`, `stat --format`, `stat --printf`.
Why: Zsh's `zsh/stat` module (command `zstat`) exposes every stat(2) field natively. `zstat -H meta file` populates an associative array keyed by field name.
Fix suggestion: `zmodload zsh/stat; zstat -H meta file; echo "${meta[size]}"`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1365_test.go
+++ b/pkg/katas/katatests/zc1365_test.go
@@ -31,8 +31,8 @@ func TestZC1365(t *testing.T) {
 			},
 		},
 		{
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

redteamx (approved):
LGTM.

PR #379, opened 2026-04-17
feat: ZC1366 — use Zsh `limit` instead of POSIX `ulimit`

afadesigns:
ZC1366 — Use Zsh `limit` instead of POSIX `ulimit` for idiomatic resource queries

What: flags `ulimit` invocations.
Why: Zsh provides `limit` as the native resource-query builtin; output is human-readable (`cputime 10 seconds` vs `-t 10`) and accepts `unlimited` as a value.
Fix suggestion: `limit cputime 10` instead of `ulimit -t 10`. Keep `ulimit` only when the script runs under Bash too.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1366_test.go
+++ b/pkg/katas/katatests/zc1366_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #380, opened 2026-04-17
feat: ZC1367 — use Zsh `strftime` instead of Bash `printf '%(fmt)T'`

afadesigns:
ZC1367 — Use Zsh `strftime` instead of Bash `printf '%(fmt)T'`

What: flags `printf` with a `%(fmt)T` format specifier.
Why: `printf '%(fmt)T'` is a Bash 4.2+ extension. Zsh's `zsh/datetime` module provides `strftime fmt seconds` natively — more readable and portable across Zsh versions.
Fix suggestion: `zmodload zsh/datetime; strftime '%Y-%m-%d' $EPOCHSECONDS`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1367_test.go
+++ b/pkg/katas/katatests/zc1367_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #381, opened 2026-04-17
feat: ZC1368 — avoid `sh -c`/`bash -c` inside Zsh scripts

afadesigns:
ZC1368 — Avoid `sh -c` / `bash -c` inside a Zsh script

What: flags `sh -c` and `bash -c` invocations.
Why: Spawning `sh`/`bash` inside a Zsh script loses access to Zsh arrays, associative arrays, and Zsh-specific syntax. A Zsh function call or inline command is usually better.
Fix suggestion: inline as a function, or use `zsh -c` when a fresh shell is required.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1368_test.go
+++ b/pkg/katas/katatests/zc1368_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #382, opened 2026-04-17
feat: ZC1369 — use Zsh `${(V)var}` instead of `od -c`

afadesigns:
ZC1369 — Prefer Zsh `${(V)var}` over `od -c` for visible-char rendering

What: flags `od -c` and `od -C` invocations.
Why: Zsh's `${(V)var}` renders non-printable characters in visible form (e.g. `\n` for newline) directly from a variable, without spawning `od`.
Fix suggestion: `print -r -- ${(V)line}`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1369_test.go
+++ b/pkg/katas/katatests/zc1369_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #383, opened 2026-04-17
feat: ZC1370 — prefer Zsh `repeat N { ... }` over `yes str | head`

afadesigns:
ZC1370 — Prefer Zsh `repeat N { ... }` over `yes str | head -n N`

What: flags `yes ...` invocations.
Why: `yes str | head -n N` is a common idiom for N copies of a line. Zsh's `repeat N { print str }` achieves the same with no external process and no pipe.
Fix suggestion: `repeat 5 { print banana }`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1370_test.go
+++ b/pkg/katas/katatests/zc1370_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #384, opened 2026-04-17
feat: ZC1371 — use Zsh array `:t` modifier instead of `basename -a`

afadesigns:
ZC1371 — Use Zsh array `:t` modifier instead of `basename -a` for bulk path stripping

What: flags `basename -a ...` and `basename --multiple`.
Why: Zsh's `${array:t}` modifier extracts the tail component from every element in one go, inside the shell.
Fix suggestion: `names=(${paths:t})`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1371_test.go
+++ b/pkg/katas/katatests/zc1371_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #385, opened 2026-04-17
feat: ZC1372 — use Zsh `zmv` autoload instead of `rename`/`rename.ul`

afadesigns:
ZC1372 — Use Zsh `zmv` autoload function instead of `rename`/`rename.ul`/`prename`

What: flags `rename`, `rename.ul`, `prename` invocations.
Why: Zsh's `zmv` (autoloaded via `autoload -Uz zmv`) batch-renames files using glob patterns with capture groups.
Fix suggestion: `autoload -Uz zmv; zmv '(*).txt' '\$1.md'`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1372_test.go
+++ b/pkg/katas/katatests/zc1372_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #386, opened 2026-04-17
feat: ZC1373 — use Zsh `${(0)var}` NUL-split flag instead of `env -0`

afadesigns:
ZC1373 — Use Zsh `${(0)var}` flag for NUL-split parsing instead of `env -0`

What: flags `env -0` / `env --null` invocations.
Why: Zsh's `${(0)var}` parameter flag splits a string on NUL into an array natively. Reading `/proc/*/environ` or similar NUL-delimited sources no longer needs `env -0 | xargs -0` pipelines.
Fix suggestion: `envs=(${(0)"$(<file)"}) ; print -l $envs`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1373_test.go
+++ b/pkg/katas/katatests/zc1373_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #387, opened 2026-04-17
feat: ZC1374 — clarify $FUNCNEST semantics differ between Bash and Zsh

afadesigns:
ZC1374 — Avoid reading `\$FUNCNEST` as depth — Zsh uses it only as a limit

What: flags `echo`/`print`/`printf` with `\$FUNCNEST` / `\${FUNCNEST}` as an argument.
Why: In Bash, `\$FUNCNEST` is both the configured limit and the current depth. In Zsh, `\$FUNCNEST` is **only** the limit; current nesting depth is `\${#funcstack}`. Code that reads `\$FUNCNEST` expecting depth produces wrong results under Zsh.
Fix suggestion: `print \${#funcstack}` for current depth, keep `\$FUNCNEST` only when setting the limit.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1374_test.go
+++ b/pkg/katas/katatests/zc1374_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #388, opened 2026-04-17
feat: ZC1375 — use `set --` instead of `shift $#` to clear positional args

afadesigns:
ZC1375 — Use `set --` (empty) instead of `shift \$#` to clear positional arguments

What: flags `shift \$#` and `shift \${#}` invocations.
Why: `shift \$#` discards all positional args indirectly. `set --` clears them atomically and reads clearer.
Fix suggestion: `set --` replaces `shift \$#` in any context.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1375_test.go
+++ b/pkg/katas/katatests/zc1375_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #389, opened 2026-04-17
feat: ZC1375 — use `[[ -t fd ]]` instead of `tty -s`

afadesigns:
ZC1375 — Use `[[ -t fd ]]` instead of `tty -s` for tty-check

What: flags `tty -s` invocations.
Why: `tty -s` returns 0 iff stdin is a terminal. Zsh's `[[ -t 0 ]]` test works in-shell with any file descriptor (0 stdin, 1 stdout, 2 stderr) — no subprocess.
Fix suggestion: `[[ -t 1 ]] && colored_output || plain_output`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1375_test.go
+++ b/pkg/katas/katatests/zc1375_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #390, opened 2026-04-17
feat: ZC1376 — warn on $BASH_XTRACEFD (Bash-only)

afadesigns:
ZC1376 — Avoid \`\$BASH_XTRACEFD\` — Bash-only, use \`exec {fd}>file\` + \`setopt XTRACE\`

What: flags references to \`\$BASH_XTRACEFD\` / \`\${BASH_XTRACEFD}\`.
Why: Bash redirects \`set -x\` output to the fd named by this variable. Zsh silently ignores the variable — the trace still goes to stderr.
Fix suggestion: \`exec {fd}>trace.log; exec 2>&\$fd; setopt XTRACE\` for equivalent behavior in Zsh.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1376_test.go
+++ b/pkg/katas/katatests/zc1376_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #391, opened 2026-04-17
feat: ZC1377 — avoid $BASH_ALIASES (Bash-only)

afadesigns:
ZC1377 — Avoid \`\$BASH_ALIASES\` — use Zsh \`\$aliases\`

What: flags references to \`\$BASH_ALIASES\` / \`\${BASH_ALIASES}\`.
Why: \`\$BASH_ALIASES\` is Bash-only. In Zsh the equivalent associative array is \`\$aliases\` (lowercase, built into the \`zsh/parameter\` module).
Fix suggestion: \`print -l \${(kv)aliases}\`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1377_test.go
+++ b/pkg/katas/katatests/zc1377_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #392, opened 2026-04-17
feat: ZC1378 — use Zsh lowercase $dirstack instead of Bash $DIRSTACK

afadesigns:
ZC1378 — Avoid uppercase \`\$DIRSTACK\` — Zsh uses lowercase \`\$dirstack\`

What: flags references to \`\$DIRSTACK\` / \`\${DIRSTACK}\`.
Why: Bash \`\$DIRSTACK\` is the pushd/popd directory stack. In Zsh the equivalent array is lowercase \`\$dirstack\` (zsh/parameter module). Using the uppercase name reads an unrelated (usually empty) variable.
Fix suggestion: \`print -l \$dirstack\`.
Severity: Error (reading wrong variable silently)

### DIFF
--- a/pkg/katas/katatests/zc1378_test.go
+++ b/pkg/katas/katatests/zc1378_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

redteamx (approved):
LGTM.

PR #394, opened 2026-04-17
feat: ZC1380 — use Zsh $HISTORY_IGNORE instead of $HISTIGNORE

afadesigns:
ZC1380 — Avoid \`\$HISTIGNORE\` — use Zsh \`\$HISTORY_IGNORE\`

What: flags references to \`\$HISTIGNORE\`.
Why: Bash filter-variable name is \`HISTIGNORE\` (no underscore). Zsh's equivalent is \`HISTORY_IGNORE\` (underscored). Setting the Bash name in Zsh is a silent no-op.
Fix suggestion: \`HISTORY_IGNORE='(ls|cd|pwd)'\`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1380_test.go
+++ b/pkg/katas/katatests/zc1380_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #397, opened 2026-04-17
feat: ZC1383 — use Zsh $TIMEFMT instead of Bash $TIMEFORMAT

afadesigns:
ZC1383 — Avoid \`\$TIMEFORMAT\` — Zsh uses \`\$TIMEFMT\`

What: flags references to \`\$TIMEFORMAT\`.
Why: Bash reads \`\$TIMEFORMAT\` to format the \`time\` builtin's output. Zsh uses the shorter name \`\$TIMEFMT\`. Setting \`TIMEFORMAT\` in Zsh has no effect.
Fix suggestion: \`TIMEFMT='%J  %U user  %S system  %P cpu  %*E total'\`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1383_test.go
+++ b/pkg/katas/katatests/zc1383_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #399, opened 2026-04-17
feat: ZC1385 — warn on $PS0 (Bash-only)

afadesigns:
ZC1385 — Avoid \`\$PS0\` — Bash-only; Zsh uses \`preexec\` hook

What: flags references to \`\$PS0\`.
Why: Bash 4.4+ expands \`\$PS0\` between reading and executing a command. Zsh has no such variable; the equivalent is a \`preexec\` function.
Fix suggestion: \`preexec() { echo "running: \$1"; }\` or \`add-zsh-hook preexec my_preexec\`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1385_test.go
+++ b/pkg/katas/katatests/zc1385_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #400, opened 2026-04-17
feat: ZC1386 — warn on $FIGNORE (Bash-only)

afadesigns:
ZC1386 — Avoid \`\$FIGNORE\` — Zsh uses compsys tag patterns

What: flags references to \`\$FIGNORE\`.
Why: Bash filters filename completion via \`\$FIGNORE\` (colon-separated suffix list). Zsh's compsys uses \`zstyle ':completion:*' ignored-patterns\` for the same effect.
Fix suggestion: \`zstyle ':completion:*' ignored-patterns '*.o *.pyc'\`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1386_test.go
+++ b/pkg/katas/katatests/zc1386_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #401, opened 2026-04-17
feat: ZC1387 — use Zsh $options instead of Bash $SHELLOPTS

afadesigns:
ZC1387 — Avoid \`\$SHELLOPTS\` — Zsh uses \`\$options\` associative array

What: flags references to \`\$SHELLOPTS\`.
Why: Bash's colon-separated \`\$SHELLOPTS\` enumerates enabled \`set\` options. Zsh exposes the same (plus \`setopt\` toggles) via the \`\$options\` assoc array.
Fix suggestion: \`print -l \${(kv)options}\` or \`[[ \${options[xtrace]} == on ]]\`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1387_test.go
+++ b/pkg/katas/katatests/zc1387_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #402, opened 2026-04-17
feat: ZC1388 — use Zsh $mailpath instead of Bash $MAILPATH

afadesigns:
ZC1388 — Use Zsh lowercase \`\$mailpath\` array instead of colon-separated \`\$MAILPATH\`

What: flags references to \`\$MAILPATH\` (uppercase) — both \`\$VAR\` and assignment form.
Why: Bash stores mailpath as a colon-separated string. Zsh uses lowercase \`\$mailpath\` as a typed array. Setting the Bash name in Zsh is ignored.
Fix suggestion: \`mailpath=(~/Mail/inbox'?New mail in inbox')\`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1388_test.go
+++ b/pkg/katas/katatests/zc1388_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #403, opened 2026-04-17
feat: ZC1389 — warn on $HOSTFILE (Bash-only)

afadesigns:
ZC1389 — Avoid \`\$HOSTFILE\` — Zsh uses \`\$hosts\` array

What: flags references to \`\$HOSTFILE\`.
Why: Bash reads \`\$HOSTFILE\` for hostname completion. Zsh uses the \`\$hosts\` array (lowercase) populated from \`/etc/hosts\` and \`~/.hosts\` by default.
Fix suggestion: \`hosts+=(\$(awk '!/^#/ && \$2 { print \$2 }' /etc/hosts))\`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1389_test.go
+++ b/pkg/katas/katatests/zc1389_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #404, opened 2026-04-17
feat: ZC1390 — $GROUPS is scalar in Zsh, not array

afadesigns:
ZC1390 — Avoid \`\$GROUPS[@]\` — Zsh \`\$GROUPS\` is a scalar, not an array

What: flags \`\${GROUPS[...]}\` / \`\$GROUPS[...]\` subscript forms.
Why: Bash \`\$GROUPS\` is an array of all group IDs. Zsh \`\$GROUPS\` is a scalar holding only the primary GID. Subscripting it as an array produces wrong results.
Fix suggestion: \`zmodload zsh/parameter; print -l \${(k)groups}\` for all groups by name.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1390_test.go
+++ b/pkg/katas/katatests/zc1390_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #405, opened 2026-04-17
feat: ZC1391 — prefer Zsh `(( ${+VAR} ))` over Bash `[[ -v VAR ]]`

afadesigns:
ZC1391 — Avoid \`[[ -v VAR ]]\` for Bash set-check — use Zsh \`(( \${+VAR} ))\`

What: flags \`test -v VAR\` / \`[ -v VAR ]\` / \`[[ -v VAR ]]\`.
Why: Bash 4.2+ added \`-v\` to test if a variable is set. Zsh's canonical form is \`(( \${+VAR} ))\` which returns 1 when set and 0 when unset — reliable across Zsh versions.
Fix suggestion: \`(( \${+VAR} )) && use_it\`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1391_test.go
+++ b/pkg/katas/katatests/zc1391_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #406, opened 2026-04-17
feat: ZC1392 — $CHILD_MAX is Bash-only, use limit/ulimit in Zsh

afadesigns:
ZC1392 — Avoid \`\$CHILD_MAX\` — Bash-only

What: flags references to \`\$CHILD_MAX\`.
Why: Bash exposes \`\$CHILD_MAX\` (max remembered exited children). Zsh does not. For process-count limits use \`limit -s maxproc\` or \`ulimit -u\` — semantics differ but are closest available.
Severity: Info

### DIFF
--- a/pkg/katas/katatests/zc1392_test.go
+++ b/pkg/katas/katatests/zc1392_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #407, opened 2026-04-17
feat: ZC1393 — warn on $SRANDOM (Bash 5.1+ only)

afadesigns:
ZC1393 — Avoid \`\$SRANDOM\` — Bash 5.1+ only, Zsh has no direct equivalent

What: flags references to \`\$SRANDOM\`.
Why: Bash 5.1 added \`\$SRANDOM\` for cryptographically secure 32-bit random values. Zsh has no equivalent; use \`/dev/urandom\` or \`openssl rand\`.
Fix suggestion: \`n=\$(od -N4 -An -tu4 /dev/urandom | tr -d ' ')\`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1393_test.go
+++ b/pkg/katas/katatests/zc1393_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #408, opened 2026-04-17
feat: ZC1394 — warn on $BASH (use Zsh $ZSH_NAME)

afadesigns:
ZC1394 — Avoid \`\$BASH\` — Zsh uses \`\$ZSH_NAME\`

What: flags standalone \`\$BASH\` references (not \`\$BASH_*\` names).
Why: Bash populates \`\$BASH\` with the interpreter path. Zsh exposes the interpreter name via \`\$ZSH_NAME\`; \`\$BASH\` is unset.
Fix suggestion: \`[[ \$ZSH_NAME == zsh ]] && ...\`.
Severity: Info

### DIFF
--- a/pkg/katas/katatests/zc1394_test.go
+++ b/pkg/katas/katatests/zc1394_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #409, opened 2026-04-17
feat: ZC1395 — warn on `wait -n` (Bash 4.3+ only)

afadesigns:
ZC1395 — Avoid \`wait -n\` — Bash 4.3+ only

What: flags \`wait -n\`.
Why: Bash 4.3+ added \`-n\` to wait for any one job to finish. Zsh's \`wait\` does not accept \`-n\`. Loop on individual PIDs for any-child semantics.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1395_test.go
+++ b/pkg/katas/katatests/zc1395_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #410, opened 2026-04-17
feat: ZC1396 — warn on `unset -n` (Bash nameref)

afadesigns:
ZC1396 — Avoid \`unset -n\` — Bash nameref operation not honored by Zsh

What: flags \`unset -n\`.
Why: Bash's \`-n\` unsets the nameref itself rather than the target. Zsh lacks nameref support; \`-n\` is error/unintended-unset territory.
Fix suggestion: \`unset -v var\` (variable) or \`unset -f fn\` (function) explicitly.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1396_test.go
+++ b/pkg/katas/katatests/zc1396_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #411, opened 2026-04-17
feat: ZC1397 — warn on Bash $COMP_TYPE/$COMP_KEY

afadesigns:
ZC1397 — Avoid Bash \`\$COMP_TYPE\`/\`\$COMP_KEY\`/\`\$COMP_WORDBREAKS\`

What: flags references to Bash completion-context variables.
Why: These exist in Bash programmable completion only. Zsh's compsys uses the \`\$compstate\` associative array.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1397_test.go
+++ b/pkg/katas/katatests/zc1397_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #413, opened 2026-04-17
feat: ZC1399 — use Zsh $signals array instead of `kill -l`

afadesigns:
ZC1399 — Use Zsh \`\$signals\` array instead of \`kill -l\` for signal enumeration

What: flags \`kill -l\`.
Why: Zsh's \`zsh/parameter\` module exposes \`\$signals\` — an array of signal names. \`print -l \$signals\` produces the same list without spawning \`kill\`.
Fix suggestion: \`zmodload zsh/parameter; print -l \$signals\`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1399_test.go
+++ b/pkg/katas/katatests/zc1399_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #414, opened 2026-04-17
feat: ZC1400 — use Zsh $CPUTYPE instead of parsing $HOSTTYPE

afadesigns:
ZC1400 — Use Zsh \`\$CPUTYPE\` for architecture detection instead of parsing \`\$HOSTTYPE\`

What: flags \`cut\`/\`awk\`/\`sed\` invocations that reference \`\$HOSTTYPE\`.
Why: Zsh pre-splits the compound \`\$HOSTTYPE\` into \`\$CPUTYPE\` (architecture) and \`\$OSTYPE\` (OS). No text-splitting external needed.
Fix suggestion: \`arch=\$CPUTYPE\` instead of \`arch=\$(cut -d- -f1 <<<\$HOSTTYPE)\`.
Severity: Info

### DIFF
--- a/pkg/katas/katatests/zc1400_test.go
+++ b/pkg/katas/katatests/zc1400_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #415, opened 2026-04-17
feat: ZC1401 — use Zsh $VENDOR instead of parsing $MACHTYPE

afadesigns:
ZC1401 — Prefer Zsh \`\$VENDOR\` over parsing \`\$MACHTYPE\` for vendor detection

What: flags \`cut\`/\`awk\`/\`sed\` invocations referencing \`\$MACHTYPE\`.
Why: Zsh pre-parses the \`MACHTYPE\` triple into \`\$CPUTYPE\`/\`\$VENDOR\`/\`\$OSTYPE\`. Direct access is cleaner than string splitting.
Fix suggestion: \`vendor=\$VENDOR\`.
Severity: Info

### DIFF
--- a/pkg/katas/katatests/zc1401_test.go
+++ b/pkg/katas/katatests/zc1401_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #416, opened 2026-04-17
feat: ZC1402 — use Zsh strftime instead of `date -d @N`

afadesigns:
ZC1402 — Avoid \`date -d @seconds\` — use Zsh \`strftime\` for epoch formatting

What: flags \`date\` with \`-d\` / \`--date\` / \`--date=...\`.
Why: \`-d\` / \`@\` form is GNU-specific (BSD \`date\` uses \`-r\`). Zsh's \`strftime\` is portable and avoids the external.
Fix suggestion: \`zmodload zsh/datetime; strftime '%Y-%m-%d' 1700000000\`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1402_test.go
+++ b/pkg/katas/katatests/zc1402_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #418, opened 2026-04-17
feat: ZC1404 — warn on $BASH_CMDS (use Zsh $commands)

afadesigns:
ZC1404 — Avoid \`\$BASH_CMDS\` — use Zsh \`\$commands\`

What: flags references to \`\$BASH_CMDS\`.
Why: Bash hash table as assoc array. Zsh equivalent is \`\$commands\` from \`zsh/parameter\`.
Fix suggestion: \`zmodload zsh/parameter; print -l \${(k)commands}\`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1404_test.go
+++ b/pkg/katas/katatests/zc1404_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #419, opened 2026-04-17
feat: ZC1405 — use (unset VAR; cmd) instead of `env -u VAR`

afadesigns:
ZC1405 — Avoid \`env -u VAR cmd\` — use Zsh \`(unset VAR; cmd)\` subshell

What: flags \`env -u\` invocations.
Why: \`(unset VAR; cmd)\` is in-shell, auto-scoped, and avoids an external \`env\` process.
Severity: Style


redteamx (approved):
LGTM.

PR #420, opened 2026-04-17
feat: ZC1406 — prefer Zsh `zargs -P` over `xargs -P`

afadesigns:
ZC1406 — Prefer Zsh \`zargs -P N\` over \`xargs -P N\` for parallel execution

What: flags \`xargs -P\` / \`xargs --max-procs\`.
Why: \`zargs\` (autoload -Uz zargs) is the Zsh-native xargs replacement — supports \`-P N\` for parallelism and keeps Zsh functions and variables in scope.
Fix suggestion: \`autoload -Uz zargs; zargs -P 4 -- item1 item2 -- my_zsh_fn\`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1406_test.go
+++ b/pkg/katas/katatests/zc1406_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #421, opened 2026-04-17
feat: ZC1407 — avoid /dev/tcp (Bash-only), use zsh/net/tcp

afadesigns:
ZC1407 — Avoid \`/dev/tcp/...\` — use Zsh \`zsh/net/tcp\` module

What: flags any path containing \`/dev/tcp/\` or \`/dev/udp/\`.
Why: These virtual files are a Bash-specific extension. Zsh provides \`zsh/net/tcp\` with \`ztcp\` builtin for real TCP operations.
Fix suggestion: \`zmodload zsh/net/tcp; ztcp host port; print -u \$REPLY "GET / HTTP/1.0\\r\\n\\r\\n"\`.
Severity: Error (portability bug)

### DIFF
--- a/pkg/katas/katatests/zc1407_test.go
+++ b/pkg/katas/katatests/zc1407_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #422, opened 2026-04-17
feat: ZC1408 — warn on $BASH_FUNC_* exported-function envvar

afadesigns:
ZC1408 — Avoid \`\$BASH_FUNC_*\` — Bash-specific exported-function envvar

What: flags references to \`\$BASH_FUNC_...\` envvars.
Why: Bash exports functions into environment variables with this prefix. Zsh does not consume the format — function export must go through \`\$FPATH\` and \`autoload\`.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1408_test.go
+++ b/pkg/katas/katatests/zc1408_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #423, opened 2026-04-17
feat: ZC1409 — warn on `test -N` (Bash-only)

afadesigns:
ZC1409 — Avoid \`[ -N file ]\` / \`test -N\` — Bash extension

What: flags \`test -N\` / \`[ -N ]\`.
Why: Bash-only modified-since-read test. Zsh has no \`-N\` — use \`zstat\` for explicit mtime/atime comparison.
Severity: Info


redteamx (approved):
LGTM.

PR #425, opened 2026-04-17
feat: ZC1410 — compopt is Bash-only (use zstyle)

afadesigns:
ZC1410 — `compopt` tweaks Bash completion options. Zsh compsys uses `zstyle` for tuning. Severity: Error


redteamx (approved):
LGTM.

PR #426, opened 2026-04-17
feat: ZC1411 — use Zsh `disable` instead of Bash `enable -n`

afadesigns:
ZC1411 — Use Zsh `disable name` instead of Bash `enable -n name`. Zsh has a dedicated `disable` builtin. Re-enable with `enable`. Severity: Style


redteamx (approved):
LGTM.

PR #427, opened 2026-04-17
feat: ZC1412 — $COMPREPLY is Bash-only (use Zsh compadd)

afadesigns:
ZC1412 — $COMPREPLY is Bash completion output array. Zsh compsys uses `compadd` builtin. Severity: Error


redteamx (approved):
LGTM.

PR #428, opened 2026-04-17
feat: ZC1413 — use Zsh `whence -p` instead of `hash -t`

afadesigns:
ZC1413 — `whence -p cmd` prints absolute path regardless of hash state, more reliable than `hash -t`. Severity: Style


redteamx (approved):
LGTM.

PR #429, opened 2026-04-17
feat: ZC1414 — warn on `hash -d` (opposite semantics)

afadesigns:
ZC1414 — `hash -d` deletes in Bash, defines named dir in Zsh. Use `unhash` for removal in Zsh. Severity: Error


redteamx (approved):
LGTM.

PR #430, opened 2026-04-17
feat: ZC1415 — prefer Zsh TRAPZERR over `trap 'cmd' ERR`

afadesigns:
ZC1415 — Zsh's `TRAPZERR() { ... }` named function is more idiomatic than `trap` string for ERR handling. Severity: Info


redteamx (approved):
LGTM.

PR #431, opened 2026-04-17
feat: ZC1416 — use Zsh preexec instead of `trap 'cmd' DEBUG`

afadesigns:
ZC1416 — Zsh's DEBUG trap doesn't fire like Bash's. Use `preexec` function. Severity: Warning


redteamx (approved):
LGTM.

PR #432, opened 2026-04-17
feat: ZC1417 — prefer Zsh TRAPRETURN over `trap 'cmd' RETURN`

afadesigns:
ZC1417 — Zsh's named `TRAPRETURN` function is idiomatic. Severity: Info


redteamx (approved):
LGTM.

PR #433, opened 2026-04-17
feat: ZC1418 — use Zsh `limit -h`/`-s` instead of `ulimit -H`/`-S`

afadesigns:
ZC1418 — Zsh's `limit` builtin uses lowercase -h/-s and produces human-readable output. Severity: Style


redteamx (approved):
LGTM.

PR #437, opened 2026-04-17
feat: ZC1421 — warn on `chpasswd` without -e

afadesigns:
ZC1421 — `chpasswd` without `-e` accepts plaintext passwords, visible via `ps`/logs/env. Use hashed input or a credentials store. Severity: Error


redteamx (approved):
LGTM.

PR #438, opened 2026-04-17
feat: ZC1422 — warn on `sudo -S` (password-via-stdin)

afadesigns:
ZC1422 — `sudo -S` enables `echo pw | sudo -S` patterns that put plaintext in process tree. Use askpass, NOPASSWD, or pkexec. Severity: Error


redteamx (approved):
LGTM.

PR #439, opened 2026-04-17
feat: ZC1423 — warn on iptables -F / nft flush (lockout risk)

afadesigns:
ZC1423 — Flushing firewall rules on a remote host can lock you out if default policy is DROP. Save + rollback timer recommended. Severity: Warning


redteamx (approved):
LGTM.

PR #443, opened 2026-04-17
feat: ZC1427 — warn on `nc -e` (reverse-shell pattern)

afadesigns:
ZC1427 — `nc -e cmd` is a reverse-shell pattern. Remove from scripts unless audited. Severity: Error


redteamx (approved):
LGTM.

PR #446, opened 2026-04-17
feat: ZC1430 — prefer Zsh `sched` over `at`/`batch`

afadesigns:
ZC1430 — Zsh's `zsh/sched` module schedules commands in-shell, no atd dependency. Severity: Style


redteamx (approved):
LGTM.

PR #453, opened 2026-04-17
feat: ZC1436 — `sysctl -w` is ephemeral

afadesigns:
ZC1436 — `sysctl -w` doesn't survive reboot. Persist via /etc/sysctl.d/. Severity: Info


redteamx (approved):
LGTM.

PR #455, opened 2026-04-17
feat: ZC1438 — warn on `systemctl mask`

afadesigns:
ZC1438 — `systemctl mask` symlinks unit to /dev/null, blocks manual start. Document unmask path. Severity: Warning

100-kata session milestone reached.


redteamx (approved):
LGTM.

PR #460, opened 2026-04-17
feat: ZC1443 — warn on `terraform destroy` without `-target`

afadesigns:
ZC1443 — `terraform destroy` without -target wipes everything in state. Severity: Warning


redteamx (approved):
LGTM.

PR #463, opened 2026-04-17
feat: ZC1446 — warn on aws s3 bulk-delete

afadesigns:
ZC1446 — aws s3 rm --recursive / rb --force can wipe buckets. Use versioning + --dryrun. Severity: Error


redteamx (approved):
LGTM.

PR #464, opened 2026-04-17
feat: ZC1447 — prefer iproute2 over net-tools

afadesigns:
ZC1447 — net-tools deprecated, not installed by default on modern distros. Use ip/ss. Severity: Style


redteamx (approved):
LGTM.

PR #466, opened 2026-04-17
feat: ZC1449 — warn on dnf/yum install without -y

afadesigns:
ZC1449 — dnf/yum install without -y hangs on prompt in scripts. Severity: Warning


redteamx (approved):
LGTM.

PR #468, opened 2026-04-17
feat: ZC1451 — warn on `pip install` without --user/venv

afadesigns:
ZC1451 — System-wide pip install pollutes system Python / fails under PEP 668. Use venv or --user. Severity: Warning


redteamx (approved):
LGTM.

PR #475, opened 2026-04-17
feat: ZC1458 — warn on `docker run --user root/0`

afadesigns:
ZC1458 — Explicit root inside container = host root on escape. Use non-root USER. Severity: Warning

120-kata session milestone.


redteamx (approved):
LGTM.

PR #476, opened 2026-04-17
feat: ZC1459 — warn on docker run --cap-add dangerous capabilities

afadesigns:
## Summary
- Flags `docker run --cap-add=SYS_ADMIN` (or SYS_PTRACE / SYS_MODULE / SYS_RAWIO / NET_ADMIN / DAC_READ_SEARCH / ALL) which breaks the container security boundary
- Handles both `--cap-add=VALUE` and `--cap-add VALUE` forms

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.4.56 (456 katas)

### DIFF
--- a/pkg/katas/katatests/zc1459_test.go
+++ b/pkg/katas/katatests/zc1459_test.go
@@ -1,0 +1,65 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #478, opened 2026-04-17
feat: ZC1461 — avoid docker run --pid=host (shared host PID ns)

afadesigns:
## Summary
- Flags `docker run --pid=host` / `docker run --pid host` — container can see and signal every host process (including init)
- Ignores container-scoped forms like `--pid=container:abc`

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.4.58 (458 katas)


redteamx (approved):
LGTM.

PR #480, opened 2026-04-17
feat: ZC1463 — avoid docker run --userns=host (disables user-ns remap)

afadesigns:
## Summary
- Flags `docker run --userns=host` / `--userns host` — disables UID remap, makes UID 0 in container == UID 0 on host
- Ignores container-scoped values (`keep-id`, `private`, etc.)
- Moves `mke2fs` typos entry from `extend-words` to `extend-identifiers` so the `mke` digit-prefix no longer flags as typo of `make`

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.4.60 (460 katas)

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -1,5 +1,7 @@
 [default.extend-words]
 # DoubleBracketExpression abbreviated as dbe in AST node names
 dbe = "dbe"
+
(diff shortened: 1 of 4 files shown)


redteamx (approved):
LGTM.

PR #481, opened 2026-04-17
feat: ZC1464 — warn on iptables -F / -P INPUT ACCEPT (firewall weakening)

afadesigns:
## Summary
- Flags `iptables -F` (flush all) and `iptables -P INPUT/FORWARD ACCEPT` (default-accept policy)
- Covers `iptables`, `ip6tables`, and `nft`
- Rules with explicit chain scoping (`iptables -F INPUT`) could be legitimate — out of scope for now

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.4.61 (461 katas)

### DIFF
--- a/pkg/katas/katatests/zc1464_test.go
+++ b/pkg/katas/katatests/zc1464_test.go
@@ -1,0 +1,75 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #482, opened 2026-04-17
feat: ZC1465 — warn on setenforce 0 (disables SELinux enforcement)

afadesigns:
## Summary
- Flags `setenforce 0` / `setenforce Permissive` — disables SELinux enforcing mode host-wide
- Suggestion: use `audit2allow` for the specific AVC instead of going permissive

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.4.62 (462 katas)


redteamx (approved):
LGTM.

PR #483, opened 2026-04-17
feat: ZC1466 — warn on disabling the host firewall

afadesigns:
## Summary
- Flags `ufw disable` — hard off for UFW
- Flags `systemctl stop|disable|mask` for `firewalld`, `ufw`, `nftables`, `iptables` (with or without `.service`)
- Leaves `systemctl start|enable` alone

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.4.63 (463 katas)

### DIFF
--- a/pkg/katas/katatests/zc1466_test.go
+++ b/pkg/katas/katatests/zc1466_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #488, opened 2026-04-17
feat: ZC1471 — error on kubectl/helm/oc --insecure-skip-tls-verify

afadesigns:
## Summary
- Flags `kubectl|helm|oc` with `--insecure-skip-tls-verify` (flag-after-subcommand) or `--insecure-skip-tls-verify=true`
- Parser limitation: flag-before-subcommand is swallowed as command name, so that form is out of scope
- Severity: Error — cluster secrets leak to MITM

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.4.68 (468 katas)

### DIFF
--- a/pkg/katas/katatests/zc1471_test.go
+++ b/pkg/katas/katatests/zc1471_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #493, opened 2026-04-17
feat: ZC1476 — warn on apt-key add/adv (deprecated, global trust)

afadesigns:
## Summary
- Flags `apt-key add` and `apt-key adv` (including `--recv-keys`)
- apt-key was deprecated in APT 2.2, removed in 2.5 — keys land in a global keyring that signs every repo
- Suggest `/etc/apt/keyrings/<vendor>.gpg` + `signed-by=` on the sources.list entry

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.4.73 (473 katas)

### DIFF
--- a/pkg/katas/katatests/zc1475_test.go
+++ b/pkg/katas/katatests/zc1475_test.go
@@ -1,0 +1,65 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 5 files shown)


redteamx (approved):
LGTM.

PR #494, opened 2026-04-17
feat: ZC1477 — warn on printf "$var" (format-string attack)

afadesigns:
## Summary
- Flags `printf "$var"`, `printf "$(cmd)"`, `printf "$x text"` — any unquoted variable in format position
- Leaves single-quoted format strings alone (no interpolation possible)
- Suggest `printf '%s\n' "$var"`

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.4.74 (474 katas)

### DIFF
--- a/pkg/katas/katatests/zc1477_test.go
+++ b/pkg/katas/katatests/zc1477_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #495, opened 2026-04-17
feat: ZC1478 — warn on mktemp -u (TOCTOU race)

afadesigns:
## Summary
- Flags `mktemp -u` (allocates name without creating the file → TOCTOU)
- Suggests dropping `-u` or using `mktemp -d` for a directory
- Parser limitation: `--dry-run` is swallowed as command name, so only the `-u` short form is in scope

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.4.75 (475 katas)

### DIFF
--- a/pkg/katas/katatests/zc1478_test.go
+++ b/pkg/katas/katatests/zc1478_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #500, opened 2026-04-17
feat: ZC1483 — warn on pip install --break-system-packages (PEP 668 bypass)

afadesigns:
## Summary
- Flags `pip|pip3|pipx install --break-system-packages`
- Bypasses PEP 668 externally-managed guard → collides with apt/dnf-managed Python files
- Suggest venv / pipx / uv / poetry instead
- Severity: Warning

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.4.80 (480 katas)

### DIFF
--- a/pkg/katas/katatests/zc1483_test.go
+++ b/pkg/katas/katatests/zc1483_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #504, opened 2026-04-17
feat: ZC1487 — warn on history -c (Bash-ism / post-compromise tactic)

afadesigns:
## Summary
- Flags `history -c` and `history -d` — clears history in Bash; silently no-ops in Zsh
- Dual relevance: post-compromise anti-forensics in Bash, programmer-surprise bug in Zsh
- Severity: Warning

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.4.84 (484 katas)

### DIFF
--- a/pkg/katas/katatests/zc1487_test.go
+++ b/pkg/katas/katatests/zc1487_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #511, opened 2026-04-17
feat: ZC1494 — warn on tcpdump -w without -Z (pcap as root)

afadesigns:
## Summary
- Flags `tcpdump -w <file>` without matching `-Z <user>`
- Without -Z, capture file is owned by root and filter runs with root privileges
- Suggest `-Z tcpdump` (or dedicated capture user)
- Severity: Warning

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.4.91 (491 katas)

### DIFF
--- a/pkg/katas/katatests/zc1494_test.go
+++ b/pkg/katas/katatests/zc1494_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #515, opened 2026-04-17
feat: ZC1498 — warn on mount -o remount,rw / (read-only root bypass)

afadesigns:
## Summary
- Flags `mount -o remount,rw /` / `rw,remount /` (and `/root`, `/boot`)
- Breaks rollback guarantees on immutable distros (Silverblue, Chrome OS, appliance images)
- Suggest systemd-sysext or ostree admin deploy
- Severity: Warning

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.4.95 (495 katas)

### DIFF
--- a/pkg/katas/katatests/zc1498_test.go
+++ b/pkg/katas/katatests/zc1498_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #519, opened 2026-04-17
feat: ZC1502 — warn on grep "$var" without -- (flag injection)

afadesigns:
## Summary
- Flags `grep|egrep|fgrep|rg|ag "$var" file` (or unquoted `$var`) without `--` end-of-flags marker
- Attacker-controlled `$var` starting with `-` becomes a flag (e.g. `-f /etc/shadow`)
- Suggest `grep -- "$var" file`
- Severity: Warning

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.4.99 (499 katas)

### DIFF
--- a/pkg/katas/katatests/zc1502_test.go
+++ b/pkg/katas/katatests/zc1502_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #524, opened 2026-04-17
feat: ZC1507 — warn on rsync -a / -l without --safe-links (symlink escape)

afadesigns:
## Summary
- Flags `rsync` with `-a` / `-l` / `--links` but no `--safe-links` / `--copy-unsafe-links` / `--no-links` / `--munge-links`
- Default symlink preservation allows traversal when a symlink points outside the source tree
- Parser limitation: `--archive` long form is swallowed upstream
- Severity: Warning

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.5.4 (504 katas)

### DIFF
--- a/pkg/katas/katatests/zc1507_test.go
+++ b/pkg/katas/katatests/zc1507_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #540, opened 2026-04-17
feat: ZC1523 — error on tar -C / (extract into root)

afadesigns:
## Summary
- Flags `tar|bsdtar -C /` / `--directory /`
- Malicious archive paths like `etc/pam.d/sshd` would overwrite system files
- Suggest stage-inspect-copy workflow
- Severity: Error

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.5.20 (520 katas)

### DIFF
--- a/pkg/katas/katatests/zc1523_test.go
+++ b/pkg/katas/katatests/zc1523_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #545, opened 2026-04-17
feat: ZC1527 — warn on crontab - (overwrites user cron from stdin)

afadesigns:
## Summary
- Flags `crontab -` and `crontab -u <user> -`
- Wholesale replacement of user crontab — drops existing rows, common persistence one-liner
- Suggest /etc/cron.d/ or read-modify-write
- Severity: Warning

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.5.24 (524 katas)

### DIFF
--- a/pkg/katas/katatests/zc1527_test.go
+++ b/pkg/katas/katatests/zc1527_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #551, opened 2026-04-17
feat: ZC1533 — warn on setsid <cmd> (detaches from TTY, escapes supervision)

afadesigns:
## Summary
- Flags any `setsid <cmd>` invocation
- Detaches from TTY/session — SIGHUP won't reach child, falls off job table
- Legit for daemonising but systemd does it better; also persistence pattern
- Severity: Warning

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.5.30 (530 katas)

### DIFF
--- a/pkg/katas/katatests/zc1533_test.go
+++ b/pkg/katas/katatests/zc1533_test.go
@@ -1,0 +1,48 @@
+package katas
+
+import (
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #571, opened 2026-04-17
feat: ZC1553 — style: use Zsh ${(U)var} instead of tr case-class

afadesigns:
## Summary
- Flags `tr '[:lower:]' '[:upper:]'` / `tr a-z A-Z` (and reverse direction)
- Zsh parameter-expansion flags `${(U)var}` / `${(L)var}` avoid the fork/exec
- Also sidesteps BusyBox/macOS `tr` portability quirks
- Severity: Style

## Test plan
- [x] `go test ./...`
- [x] Version bumped to 0.5.50 (550 katas)

### DIFF
--- a/pkg/katas/katatests/zc1553_test.go
+++ b/pkg/katas/katatests/zc1553_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #606, opened 2026-04-17
feat: ZC1588 — flag `nsenter -t 1` host-namespace container escape

afadesigns:
ZC1588 — Error on `nsenter --target 1` — joins host init namespaces (container escape)

What: flags `nsenter -t 1` / `nsenter -t1` / `nsenter --target=1` — attaching the caller to the namespaces of pid 1.
Why: inside a privileged container pid 1 is the host init. Joining its mount/pid/net/uts/ipc namespaces drops the caller onto the host with the container's capabilities intact — the canonical escape primitive.
Fix suggestion: use `docker exec` / `kubectl exec` for legitimate debugging, from the host side rather than pivoting out of a guest.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1588_test.go
+++ b/pkg/katas/katatests/zc1588_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #607, opened 2026-04-17
feat: ZC1589 — flag `trap 'set -x' ERR/RETURN/EXIT` xtrace-from-trap leak

afadesigns:
ZC1589 — Warn on `trap 'set -x' ERR/RETURN/EXIT/ZERR` — trace hook leaks env to stderr

What: flags traps that install `set -x` / `set -o xtrace` / `set -v` on ERR, RETURN, EXIT, or ZERR.
Why: once the trap fires, every subsequent expansion hits stderr with environment vars substituted — API tokens, passwords, signed URLs. In CI that stderr ends up in build logs.
Fix suggestion: scope `set -x` to a `set -x ... set +x` block, or replace with `trap 'safe_dump' ERR` that only prints non-sensitive state.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1589_test.go
+++ b/pkg/katas/katatests/zc1589_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #608, opened 2026-04-17
fix: move `ede` / `Hashi` to typos extend-words

afadesigns:
`extend-identifiers` only matches whole-identifier tokens. `ede` inside `des-ede3-cbc` and `Hashi` inside `HashiCorp` are word-fragments after typos' tokenization, so they still trip Check Spelling. Moving them to `extend-words` allows substring matching.

## Test plan
- [x] `./typos .` clean locally


redteamx (approved):
LGTM.

PR #610, opened 2026-04-17
feat: ZC1591 — prefer `print -l` / `${(F)array}` over `printf '%s\n' "${a[@]}"`

afadesigns:
ZC1591 — Style: use Zsh `print -l` / `${(F)array}` for one-per-line array printing

What: flags `printf '%s\n' "${array[@]}"` and `printf '%s' "${a[*]}"` — the Bash idiom for one-element-per-line.
Why: Zsh has shorter, safer options. `print -l -r -- "${array[@]}"` prints one raw element per line. `${(F)array}` joins with newlines inline. Both avoid format-string surprises if an element ever contains `%`.
Fix suggestion: swap `printf '%s\n' "${array[@]}"` for `print -l -r -- "${array[@]}"`; use `${(F)array}` in command substitutions.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1591_test.go
+++ b/pkg/katas/katatests/zc1591_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #611, opened 2026-04-17
feat: ZC1592 — flag `faillock --reset` / `pam_tally2 -r` lockout clearing

afadesigns:
ZC1592 — Warn on `faillock --reset` / `pam_tally2 -r` — clears PAM failed-auth counter

What: flags `faillock ... --reset` and `pam_tally2 -r` / `pam_tally2 --reset`.
Why: both tools zero the PAM counter that locks an account after too many failed logins. Automated resets also erase evidence of an ongoing brute-force attempt — intrusion detection relies on those counters for alerting.
Fix suggestion: do not automate resets. If you must, snapshot the prior count first and page security on every invocation.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1592_test.go
+++ b/pkg/katas/katatests/zc1592_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #612, opened 2026-04-17
feat: ZC1593 — flag `blkdiscard` full-device TRIM wipe

afadesigns:
ZC1593 — Error on `blkdiscard` — TRIM/DISCARD across the whole device (data loss)

What: flags every `blkdiscard` invocation with a target argument.
Why: the command tells the SSD controller to invalidate every block in range. Modern drives make the data unrecoverable the moment the controller acknowledges — even forensic recovery cannot pull it back.
Fix suggestion: gate the command behind interactive operator confirmation, not shell flow control; never leave it on a codepath reachable from a typo or untrusted input.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1593_test.go
+++ b/pkg/katas/katatests/zc1593_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #613, opened 2026-04-17
feat: ZC1594 — flag `docker run --security-opt=systempaths=unconfined`

afadesigns:
ZC1594 — Warn on `docker/podman run --security-opt=systempaths=unconfined` — unhides host kernel knobs

What: flags `docker run` / `docker create` / `podman run` / `podman create` invocations that set `--security-opt=systempaths=unconfined`.
Why: the default `systempaths=all` masks `/proc/sys`, `/proc/sysrq-trigger`, `/sys/firmware`, etc. from the container. Unhiding them lets a compromise inside the container panic the host via `sysrq-trigger` or re-tune kernel policy via `/proc/sys/kernel/*`.
Fix suggestion: keep the default masking and, if a specific kernel tunable is genuinely needed, bind-mount only that single path into the container.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1594_test.go
+++ b/pkg/katas/katatests/zc1594_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #614, opened 2026-04-17
feat: ZC1595 — flag `setfacl -m u:nobody:... / o::rwx` ACL bypass

afadesigns:
ZC1595 — Warn on `setfacl -m u:nobody:... / o::rwx` — ACL grants that bypass `chmod` scrutiny

What: flags `setfacl -m` arguments granting to `u:nobody:`, `g:nobody:`, `u:nogroup:`, `g:nogroup:`, or world (`o::rwx` / `o::rw`).
Why: ACL entries live outside the mode bits that `chmod` / `ls -l` / `stat -c %a` surface. Review scripts checking `stat -c %a` miss them entirely, and granting to the `nobody` daemon-fallback account is almost always a misconfiguration.
Fix suggestion: use `chmod` for world perms. For named users, name the real account — not `nobody` — and give only the minimum required perm.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1595_test.go
+++ b/pkg/katas/katatests/zc1595_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #615, opened 2026-04-17
feat: ZC1596 — flag `emulate sh/bash/ksh` without `-L` global option flip

afadesigns:
ZC1596 — Style: `emulate sh/bash/ksh` without `-L` flips options for the whole shell

What: flags `emulate sh` / `emulate bash` / `emulate ksh` / `emulate csh` without the `-L` flag (scope-to-function).
Why: without `-L`, `emulate` rewrites option state for the entire shell. After that line, Zsh-specific constructs (`${(F)arr}`, 1-indexed arrays, glob qualifiers) error or silently behave differently.
Fix suggestion: wrap the call in a function and use `emulate -L MODE`, or rename the script to `.sh` if Zsh features are not needed.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1596_test.go
+++ b/pkg/katas/katatests/zc1596_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #616, opened 2026-04-17
feat: ZC1597 — flag `systemd-run -p User=root` transient-unit privilege escalation

afadesigns:
ZC1597 — Warn on `systemd-run -p User=root` — launches arbitrary command with root privileges

What: flags `systemd-run -p User=root` / `systemd-run -p User=0`.
Why: systemd-run submits a transient unit. `User=root` runs it as root — bypassing the `sudo` audit path in `/var/log/auth.log`. On hosts where the caller's polkit / dbus rules permit the call, this is privilege escalation by another name.
Fix suggestion: use explicit `sudo` so the invocation is logged, or pre-provision a dedicated systemd unit that names the exact command it can run.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1597_test.go
+++ b/pkg/katas/katatests/zc1597_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #618, opened 2026-04-17
feat: ZC1599 — flag `ldconfig -f PATH` with non-`/etc/` loader config

afadesigns:
ZC1599 — Warn on `ldconfig -f PATH` with non-`/etc/` config — attacker-writable loader cache

What: flags `ldconfig -f <path>` where `<path>` does not start with `/etc/` (and is not a shell variable `$X`).
Why: `ldconfig -f` rebuilds `/etc/ld.so.cache` using the given config. If the file sits in `/tmp`, `/var/tmp`, `$HOME`, or any attacker-writable path, the attacker can add an `include` line pointing at their library directory and every subsequent executable on the host loads their `.so` first.
Fix suggestion: keep config under `/etc/ld.so.conf.d/` with root ownership, and run `ldconfig` with no `-f`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1599_test.go
+++ b/pkg/katas/katatests/zc1599_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #619, opened 2026-04-17
feat: ZC1600 — flag bare `chroot DIR CMD` keeping uid 0 inside the jail

afadesigns:
ZC1600 — Warn on bare `chroot DIR CMD` — missing `--userspec=` keeps uid 0 inside the jail

What: flags every `chroot DIR CMD` invocation. ( When `--userspec=` is used, the parser re-interprets the tokens and this kata never fires — only the bare form hits. )
Why: `chroot` changes the filesystem root but does not drop privileges. The caller is almost always root, so without `--userspec=USER:GROUP` the inner command still runs as uid 0. It can write anywhere in the tree and — given bind-mounted `/proc`, `/sys`, or device nodes — escape.
Fix suggestion: pass `--userspec=USER:GROUP`, or drop to `bwrap` / `firejail` for user-namespace isolation.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1600_test.go
+++ b/pkg/katas/katatests/zc1600_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #620, opened 2026-04-17
feat: ZC1601 — flag `ethtool -s $IF wol` enabling Wake-on-LAN

afadesigns:
ZC1601 — Warn on `ethtool -s $IF wol <g|u|m|b|a|p|s|f>` — enables remote Wake-on-LAN

What: flags `ethtool -s $IF wol X` where X is any enable bit (`g` magic, `u` unicast, `m` multicast, `b` broadcast, `a` ARP, `p` phy, `s` SecureOn, `f` filter). `wol d` (disable) is valid.
Why: WoL powers the host from sleep / soft-off when a matching packet reaches the NIC. The wake path is firmware-level and fires before the kernel boots or firewall rules load — nothing filters the wake packet.
Fix suggestion: keep `wol d` unless a documented operational need requires one of the wake bits.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1601_test.go
+++ b/pkg/katas/katatests/zc1601_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #621, opened 2026-04-17
feat: ZC1602 — flag `setopt KSH_ARRAYS / SH_WORD_SPLIT` global semantics flip

afadesigns:
ZC1602 — Warn on `setopt KSH_ARRAYS` / `SH_WORD_SPLIT` — flips Zsh core semantics shell-wide

What: flags `setopt` with the args `KSH_ARRAYS`, `SH_WORD_SPLIT`, or their lowercase / no-underscore variants (Zsh accepts `ksharrays`, `shwordsplit`, etc.).
Why: `KSH_ARRAYS` makes arrays 0-indexed, breaking every Zsh access that uses `[1]` for the first element. `SH_WORD_SPLIT` makes unquoted `$var` word-split on `IFS`, breaking Zsh's default "one token per reference" promise. Either setting applied globally silently breaks pre-existing code from that line on.
Fix suggestion: scope inside a function with `emulate -L ksh` or `emulate -L sh` so the calling shell keeps Zsh semantics.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1602_test.go
+++ b/pkg/katas/katatests/zc1602_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1602(t *testing.T) {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #622, opened 2026-04-17
feat: ZC1603 — flag `gdb -p PID` / `ltrace -p PID` live-attach secret exposure

afadesigns:
ZC1603 — Warn on `gdb -p PID` / `ltrace -p PID` — live attach reads target memory

What: flags `gdb -p` / `ltrace -p` invocations in scripts.
Why: both tools attach via ptrace and hand the caller full read/write access to registers, heap, stack, open fds, and every environment variable of the target. Credentials, session tokens, and TLS keys are all readable. From a root-run script this extracts everything a user's process holds.
Fix suggestion: keep production scripts out of the debugger. If post-mortem diagnostics are needed, run `coredumpctl` against a captured core file instead.
Severity: Warning

Milestone: this PR takes the project to 600 katas (v0.6.0).

### DIFF
--- a/pkg/katas/katatests/zc1603_test.go
+++ b/pkg/katas/katatests/zc1603_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #623, opened 2026-04-17
feat: ZC1604 — flag `source <glob>` sourcing every matched file

afadesigns:
ZC1604 — Warn on `source <glob>` / `. <glob>` — loads every match; one bad file = code exec

What: flags `source` and `.` commands whose first argument contains glob metacharacters (`*`, `?`, `[`).
Why: glob-sourcing (e.g. `source /etc/profile.d/*.sh`) loads every file that matches. One attacker-writable file anywhere in the glob yields arbitrary code execution as the caller, with the caller's privileges.
Fix suggestion: list the files explicitly so review can enumerate what gets loaded. If a directory of drop-ins is required, keep it root-owned and audit ownership at install time.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1604_test.go
+++ b/pkg/katas/katatests/zc1604_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #624, opened 2026-04-17
feat: ZC1605 — flag `debugfs -w DEV` write-mode fs debugger bypass

afadesigns:
ZC1605 — Error on `debugfs -w DEV` — write-mode filesystem debugger bypasses journal

What: flags `debugfs -w` invocations.
Why: `-w` opens the filesystem in write mode, sidestepping the kernel's normal write path. The journal does not see the changes, filesystem locks are ignored, and inodes / blocks can be edited directly. On a mounted filesystem this corrupts state silently.
Fix suggestion: keep `debugfs -w` out of scripts — use it interactively from a rescue environment only.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1605_test.go
+++ b/pkg/katas/katatests/zc1605_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #625, opened 2026-04-17
feat: ZC1606 — flag `mkdir -m` / `install -m` world-write without sticky

afadesigns:
ZC1606 — Warn on `mkdir -m NNN` / `install -m NNN` with world-write bit (no sticky)

What: flags `mkdir -m NNN` / `install -m NNN` with a 3-digit octal mode whose last digit is `2`, `3`, `6`, or `7`. `1777` and other sticky-prefixed modes are exempt.
Why: world-writable directories / files without the sticky bit are TOCTOU symlink-attack ground. An attacker can drop a symlink with the expected output name before the script writes, redirecting the write wherever they choose.
Fix suggestion: use `-m 700` / `-m 750` and scope access by group or ACL. Use `-m 1777` only for genuine shared-sticky temp dirs.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1606_test.go
+++ b/pkg/katas/katatests/zc1606_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #626, opened 2026-04-17
feat: ZC1607 — flag `git config safe.directory '*'` CVE-2022-24765 bypass

afadesigns:
ZC1607 — Warn on `git config safe.directory '*'` — disables CVE-2022-24765 protection

What: flags `git config ... safe.directory '*'` and the equivalent `git -c safe.directory=*` single-shot form.
Why: `safe.directory` is git's guard against fake repos planted by another uid. Setting it to `*` trusts every directory on the host — a planted `/tmp/evil/.git` with a malicious `core.fsmonitor` hook executes arbitrary code the next time any user runs `git status` near that path.
Fix suggestion: list the specific paths that need cross-owner git access, or fix the underlying ownership mismatch.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1607_test.go
+++ b/pkg/katas/katatests/zc1607_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #627, opened 2026-04-17
feat: ZC1608 — flag `find -exec sh -c '... {} ...'` filename injection

afadesigns:
ZC1608 — Warn on `find -exec sh -c '... {} ...'` — filename in quoted script is injectable

What: flags `find -exec sh -c`, `find -exec bash -c`, `find -exec zsh -c`, `find -execdir sh -c`, etc. when the quoted script argument contains `{}`.
Why: substituting `{}` directly into the shell script lets filenames with metacharacters break out. A file named `$(rm -rf ~)` triggers command substitution; a file named `foo; curl evil` chains a second command.
Fix suggestion: pass `{}` as a positional argument: `find -exec sh -c 'grep pat "$1"' _ {} \;` so the filename becomes `$1`, not source.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1608_test.go
+++ b/pkg/katas/katatests/zc1608_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #628, opened 2026-04-17
feat: ZC1609 — flag `aa-disable` / `aa-complain` / `apparmor_parser -R` profile removal

afadesigns:
ZC1609 — Warn on `aa-disable` / `aa-complain` / `apparmor_parser -R` — disables AppArmor enforcement

What: flags `aa-disable` (with target), `aa-complain` (with target), and `apparmor_parser -R` / `--remove`.
Why: `aa-disable` unloads the profile entirely. `aa-complain` flips it from enforce to complain — violations are logged but allowed. `apparmor_parser -R` removes the profile from the running kernel. In each case the confined process loses its MAC restrictions.
Fix suggestion: keep the profile enforcing. If a specific rule is wrong, edit the profile file and run `apparmor_parser -r /etc/apparmor.d/<profile>` to reload it.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1609_test.go
+++ b/pkg/katas/katatests/zc1609_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #629, opened 2026-04-17
feat: ZC1610 — flag `curl -o /etc/...` / `wget -O /etc/...` direct system-path download

afadesigns:
ZC1610 — Warn on `curl -o /etc/...` / `wget -O /etc/...` — direct download to a system path

What: flags `curl -o`, `curl --output`, `wget -O`, `wget --output-document` that target paths under `/etc/`, `/usr/`, `/bin/`, `/sbin/`, `/lib/`, `/lib64/`, `/opt/`.
Why: writing the HTTP response body straight into a system path skips every integrity check. A compromised URL or MITM replaces the file with attacker content; the next command over picks it up.
Fix suggestion: download to a temp file, verify signature / checksum, then `install -m 0644 tmpfile /etc/...` into place. For system packages, prefer the package manager.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1610_test.go
+++ b/pkg/katas/katatests/zc1610_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #630, opened 2026-04-17
feat: ZC1611 — prefer Zsh `${(U)var}` / `${(L)var}` over Bash `${var^^}` / `${var,,}`

afadesigns:
ZC1611 — Style: `${var^^}` / `${var,,}` — prefer Zsh `${(U)var}` / `${(L)var}`

What: flags arguments containing the Bash 4+ case-change syntax `${var^^}` / `${var,,}` — detected when the substring `^^}` or `,,}` appears inside a `${...}` expansion.
Why: Zsh accepts the `^^ / ,,` forms for compatibility, but the native flag form `${(U)var}` / `${(L)var}` is what Zsh documents and composes with other `(X)var` flags. Consistent Zsh style reads better.
Fix suggestion: replace with `${(U)var}` for uppercase and `${(L)var}` for lowercase.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1611_test.go
+++ b/pkg/katas/katatests/zc1611_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #631, opened 2026-04-17
feat: ZC1612 — flag `sysctl -w` disabling kernel hardening knobs

afadesigns:
ZC1612 — Warn on `sysctl -w` disabling kernel hardening knobs

What: flags `sysctl -w ... KEY=VAL` (or `sysctl KEY=VAL`) where the key/value pair loosens one of:
- `kernel.yama.ptrace_scope=0` — any process can ptrace any other.
- `kernel.kptr_restrict=0` — kernel pointers leak via `/proc/kallsyms`.
- `kernel.dmesg_restrict=0` — unprivileged dmesg access.
- `kernel.unprivileged_bpf_disabled=0` — unprivileged BPF loading.
- `net.core.bpf_jit_harden=0` — BPF JIT-spray mitigations off.
- `kernel.perf_event_paranoid=-1` / `=0` — unprivileged perf access.

Why: each is a defense-in-depth layer. Loosening them removes an isolation guarantee the kernel documents.
Fix suggestion: leave the default unless there is a measured performance or debugging need that outweighs the security trade-off.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1612_test.go
+++ b/pkg/katas/katatests/zc1612_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1612(t *testing.T) {
+	tests := []struct {
+		name     string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #632, opened 2026-04-17
feat: ZC1613 — flag `cat/grep/head` reading SSH private-key files

afadesigns:
ZC1613 — Warn on reading SSH private-key files with `cat` / `less` / `grep` / `head`

What: flags text-tool reads (`cat`, `less`, `more`, `head`, `tail`, `wc`, `grep`, `awk`, `sed`, `cut`) of SSH host keys under `/etc/ssh/ssh_host_*_key` and user keys under `*/.ssh/id_{rsa,ed25519,ecdsa,dsa}`.
Why: piping a private key through a text tool copies key material into the process and — if stdout is redirected or piped — into logs, backups, or terminal scrollback. Host keys impersonate the server; user keys impersonate the user.
Fix suggestion: use `ssh-keygen -l -f KEY` for fingerprint / metadata. Pass the key path directly to the consumer (`ssh -i`, `git -c core.sshCommand`) rather than staging it through a shell tool.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1613_test.go
+++ b/pkg/katas/katatests/zc1613_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #633, opened 2026-04-18
feat: ZC1614 — flag `expect` script containing literal password/passphrase

afadesigns:
ZC1614 — Error on `expect` script containing `password` / `passphrase`

What: flags any `expect` invocation whose argument list contains the substring `password` or `passphrase`.
Why: `expect -c '...'` puts the full scripted dialog on the command line. The credential lands in `ps`, `/proc/<pid>/cmdline`, shell history, and audit logs — readable by any local user who can list processes.
Fix suggestion: use key-based authentication (SSH keys, GSSAPI) where possible. If credential feeding is genuinely unavoidable, keep the secret in a protected file the expect script reads, not in the argv.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1614_test.go
+++ b/pkg/katas/katatests/zc1614_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #634, opened 2026-04-18
feat: ZC1615 — prefer Zsh `$EPOCHREALTIME` over `date "+%s.%N"`

afadesigns:
ZC1615 — Style: use Zsh `$EPOCHREALTIME` / `$epochtime` instead of `date "+%s.%N"`

What: flags `date ...` invocations whose format string contains both `%s` (epoch seconds) and `%N` (nanoseconds).
Why: Zsh's `zsh/datetime` module exposes `$EPOCHREALTIME` (scalar with fractional seconds) and `$epochtime` (two-element array). Both come from `clock_gettime(CLOCK_REALTIME)` without forking `date` — on a hot path, dramatically faster.
Fix suggestion: `zmodload zsh/datetime` once, then use `$EPOCHREALTIME` for the fractional epoch or `$epochtime[1]`/`$epochtime[2]` for seconds and nanoseconds.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1615_test.go
+++ b/pkg/katas/katatests/zc1615_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #635, opened 2026-04-18
feat: ZC1616 — flag `fsfreeze -f` without guaranteed unfreeze

afadesigns:
ZC1616 — Warn on `fsfreeze -f MOUNTPOINT` — filesystem stays frozen until `-u` runs

What: flags every `fsfreeze -f` invocation.
Why: `-f` blocks every write on the mountpoint until `fsfreeze -u` thaws it. The intended use is a short window around a hypervisor or LVM snapshot, but if the script errors between the freeze and the unfreeze the filesystem stays frozen forever. On the root fs, a reboot is often the only way out.
Fix suggestion: pair every freeze with `trap 'fsfreeze -u MOUNTPOINT' EXIT` so the thaw fires even on failure, and keep the frozen window under a few seconds.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1616_test.go
+++ b/pkg/katas/katatests/zc1616_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #636, opened 2026-04-18
feat: ZC1617 — flag `xargs -P 0` unbounded parallelism risk

afadesigns:
ZC1617 — Warn on `xargs -P 0` — unbounded parallelism risks CPU / fd / memory exhaustion

What: flags `xargs -P 0` and `xargs -P0`.
Why: `-P 0` tells xargs to spawn as many concurrent children as input lines. On any non-trivial input the count blows past `RLIMIT_NPROC`, saturates the downstream tool's fd limit, or drives the host OOM.
Fix suggestion: pick an explicit cap — `xargs -P $(nproc)` for CPU-bound work, `-P 4..8` for I/O-bound — so the failure mode is predictable.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1617_test.go
+++ b/pkg/katas/katatests/zc1617_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #637, opened 2026-04-18
feat: ZC1618 — flag `git commit/push --no-verify` hook bypass

afadesigns:
ZC1618 — Warn on `git commit --no-verify` / `git push --no-verify` — bypasses hooks

What: flags `git commit --no-verify`, `git push --no-verify`, and `git commit -n` (the alias).
Why: `--no-verify` skips pre-commit, commit-msg, and pre-push hooks. Those are where projects run linting, type-checking, unit tests, and secret scanning. Skipping them ships code the project's own automation would have rejected.
Fix suggestion: reserve the flag for emergencies and follow up with a commit that passes the hooks. Scripts should never need `--no-verify`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1618_test.go
+++ b/pkg/katas/katatests/zc1618_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #638, opened 2026-04-18
feat: ZC1619 — flag `mount -t nfs/cifs/smb/sshfs` missing `nosuid,nodev`

afadesigns:
ZC1619 — Warn on `mount -t nfs/cifs/smb/sshfs` missing `nosuid` or `nodev`

What: flags `mount -t {nfs,nfs4,cifs,smbfs,smb3,sshfs}` where the `-o` options don't include both `nosuid` and `nodev`.
Why: network filesystems present files whose mode bits come from a remote server. Without `nosuid`, a hostile server plants a setuid-root binary that the client honors. Without `nodev`, the server plants device nodes the kernel trusts.
Fix suggestion: always mount network shares with `nosuid,nodev` in `-o`. Add `noexec` unless the export is meant to hold executables.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1619_test.go
+++ b/pkg/katas/katatests/zc1619_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #639, opened 2026-04-18
feat: ZC1620 — flag `tee /etc/sudoers` bypassing `visudo -cf` validation

afadesigns:
ZC1620 — Error on `tee /etc/sudoers` / `/etc/sudoers.d/*` — writes without `visudo -cf`

What: flags `tee` invocations whose output path is `/etc/sudoers` or anything under `/etc/sudoers.d/`.
Why: `tee` writes stdin verbatim. A typo in a sudoers rule (stray comma, missing `ALL`, unclosed alias) leaves the file unparseable. The next sudo refuses to load it; on most systems that means nobody can become root until someone boots from rescue.
Fix suggestion: pipe through `visudo -cf /dev/stdin` first, or stage in a temp file, validate with `visudo -cf`, then atomically `mv` into `/etc/sudoers.d/`.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1620_test.go
+++ b/pkg/katas/katatests/zc1620_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #640, opened 2026-04-18
feat: ZC1621 — flag `tmux -S /tmp/SOCKET` shared socket hijack risk

afadesigns:
ZC1621 — Warn on `tmux -S /tmp/SOCKET` — shared-path socket invites session hijack

What: flags `tmux -S PATH` where `PATH` starts with `/tmp/` or `/var/tmp/`.
Why: `-S PATH` overrides the default 0700 `$XDG_RUNTIME_DIR/tmux-$UID/` location. `/tmp/` and `/var/tmp/` are world-traversable; if the socket is created with loose perms any local user who can read it can `tmux -S PATH attach` and see or drive the session — keystrokes, output, arbitrary commands in the attached pane.
Fix suggestion: keep the socket in `$XDG_RUNTIME_DIR` or another 0700-scoped parent dir.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1621_test.go
+++ b/pkg/katas/katatests/zc1621_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #641, opened 2026-04-18
feat: ZC1622 — prefer Zsh `${(X)var}` flags over Bash `${var@U/L/Q}` operators

afadesigns:
ZC1622 — Style: `${var@U/L/Q/...}` — prefer Zsh `${(U)var}` / `${(L)var}` / `${(Q)var}` flags

What: flags arguments whose `${...}` contains the Bash 5 operator suffix `@U`, `@L`, `@Q`, `@E`, `@A`, `@K`, or `@a`.
Why: Zsh 5.9+ has compatibility for most of these, but the idiomatic Zsh form is the `(X)var` flag — `${(U)var}` uppercase, `${(L)var}` lowercase, `${(Q)var}` unquote, `${(k)var}` keys, `${(t)var}` type, `${(e)var}` re-evaluate. The flag form composes (`${(Uf)str}` works) and reads consistently across Zsh docs.
Fix suggestion: replace `${var@U}` with `${(U)var}` and so on.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1622_test.go
+++ b/pkg/katas/katatests/zc1622_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1622(t *testing.T) {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #642, opened 2026-04-18
feat: ZC1623 — flag `kill -STOP PID` without guaranteed `-CONT`

afadesigns:
ZC1623 — Warn on `kill -STOP PID` / `pkill -STOP` — target halts until `kill -CONT` runs

What: flags `kill -STOP`, `kill -SIGSTOP`, `kill -19`, `kill -s STOP`, and the same forms on `pkill`.
Why: SIGSTOP halts the target process until SIGCONT arrives. If the script fails, is killed, or exits before the resume, the target stays paused indefinitely — consuming memory, holding locks, blocking its dependents.
Fix suggestion: pair every STOP with `trap "kill -CONT $PID" EXIT` so the resume fires even on failure. Consider `kill -TSTP` when the target handles tstop signals.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1623_test.go
+++ b/pkg/katas/katatests/zc1623_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #643, opened 2026-04-18
feat: ZC1624 — flag `az login -p` / `--password` credential in process list

afadesigns:
ZC1624 — Error on `az login -p` / `--password` — service-principal secret in process list

What: flags `az login -p SECRET` and `az login --password SECRET`.
Why: the password passes as an argv element. The expanded value lands in `ps`, `/proc/<pid>/cmdline`, shell history, and audit logs — readable by any local user who can list processes.
Fix suggestion: use federated-token OIDC (`--federated-token`), managed identity on the host, or interactive device-code flow. If a password must be used, set `AZURE_PASSWORD` via a protected env var and call plain `az login --service-principal`.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1624_test.go
+++ b/pkg/katas/katatests/zc1624_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #644, opened 2026-04-18
feat: ZC1625 — flag `rm --no-preserve-root` safeguard bypass

afadesigns:
ZC1625 — Error on `rm --no-preserve-root` — disables GNU rm safeguard against `rm -rf /`

What: flags `rm ... --no-preserve-root ...`.
Why: GNU `rm` refuses to remove `/` by default (the `--preserve-root` safeguard since coreutils 8.4). `--no-preserve-root` explicitly disables that check, so `rm -rf /` actually recurses and wipes the filesystem.
Fix suggestion: drop the flag. If a specific path needs deletion, list it explicitly and leave the safeguard in place.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1625_test.go
+++ b/pkg/katas/katatests/zc1625_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #645, opened 2026-04-18
feat: ZC1626 — flag `helm --set` secret key exposing value in process list

afadesigns:
ZC1626 — Error on `helm install/upgrade --set KEY=VALUE` with secret-shaped key

What: flags `helm install|upgrade|template --set KEY=VALUE` or `--set-string KEY=VALUE` where KEY contains one of `password`, `passwd`, `secret`, `token`, `apikey`, `api_key`, `access_key`, `private_key` (dotted names like `db.password` also matched).
Why: `--set` puts the full `KEY=VALUE` pair on the command line. The expanded secret value ends up in `ps`, `/proc/<pid>/cmdline`, shell history, and audit logs.
Fix suggestion: pass secrets via a protected values file (`helm install -f /secure/values.yaml`) or `--set-file KEY=PATH` so helm reads the value from PATH at apply time.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1626_test.go
+++ b/pkg/katas/katatests/zc1626_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #646, opened 2026-04-18
feat: ZC1627 — flag `crontab /tmp/FILE` TOCTOU cron injection

afadesigns:
ZC1627 — Warn on `crontab /tmp/FILE` — attacker-writable path installed as a user's cron

What: flags `crontab PATH` (including `crontab -u USER PATH`) where PATH starts with `/tmp/` or `/var/tmp/`.
Why: `crontab` reads whatever PATH contains at the moment it opens the file. A concurrent local user can substitute the file between the script's write and crontab's read, injecting their own cron rules.
Fix suggestion: stage the file in a 0700-scoped dir (`$XDG_RUNTIME_DIR/` or `mktemp -d`), or pipe the content via `crontab -` so it's never on disk.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1627_test.go
+++ b/pkg/katas/katatests/zc1627_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #647, opened 2026-04-18
feat: ZC1628 — flag `insmod` / `modprobe -f` kernel-module bypass

afadesigns:
ZC1628 — Warn on `insmod` / `modprobe -f` — loads modules bypassing blacklist / signature checks

What: flags `insmod PATH.ko` (any invocation with an arg) and `modprobe -f`.
Why: `insmod` loads a module directly, skipping depmod and `/etc/modprobe.d/*.conf` blacklists. `modprobe -f` tells modprobe to ignore version-magic and kernel-mismatch checks. Either path lets a disabled or mismatched module into the kernel — crash, privilege escalation, or full compromise.
Fix suggestion: use plain `modprobe MODNAME` so system policy and signature verification run; fix any underlying version mismatch rather than masking it.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1628_test.go
+++ b/pkg/katas/katatests/zc1628_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #648, opened 2026-04-18
feat: ZC1629 — flag `rsync --rsync-path='sudo rsync'` hidden privilege escalation

afadesigns:
ZC1629 — Warn on `rsync --rsync-path='sudo rsync'` — remote privilege escalation

What: flags `rsync --rsync-path=VALUE` where VALUE contains `sudo`, `doas`, or `pkexec`.
Why: `--rsync-path` is normally used to pick a non-default remote rsync binary. Setting it to `sudo rsync` makes the remote side run rsync as root — a privilege-escalation hidden inside what looks like a path override.
Fix suggestion: provision a scoped sudoers rule on the remote host that names the specific rsync invocation, and keep the path explicit (`--rsync-path=/usr/bin/rsync`).
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1629_test.go
+++ b/pkg/katas/katatests/zc1629_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #649, opened 2026-04-18
feat: ZC1630 — flag `php -S 0.0.0.0:PORT` dev server on all interfaces

afadesigns:
ZC1630 — Warn on `php -S 0.0.0.0:PORT` — PHP dev server exposes CWD to all interfaces

What: flags `php -S HOST:PORT` where HOST is `0.0.0.0`, `*`, or `[::]` (all interfaces, IPv4 or IPv6).
Why: PHP's built-in dev server serves files from the working directory with no auth, no TLS, and minimal access logging. The PHP docs explicitly warn against production use. Public-interface binds expose the CWD to everyone on the network.
Fix suggestion: bind to `127.0.0.1:PORT` locally; put nginx / caddy in front for external exposure.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1630_test.go
+++ b/pkg/katas/katatests/zc1630_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #651, opened 2026-04-18
feat: ZC1632 — flag `shred` unreliable on journaled / CoW filesystems

afadesigns:
ZC1632 — Warn on `shred` — unreliable on journaled / CoW filesystems (ext4, btrfs, zfs)

What: flags every `shred` invocation.
Why: `shred` assumes in-place overwrites (ext2 behavior). On ext4 with journaling, overwrite passes go through the journal and may not hit the original blocks. On CoW filesystems (btrfs, zfs, xfs with reflink) the overwrite lands in fresh blocks and leaves old content intact. `shred`'s own man page warns about this.
Fix suggestion: use full-disk encryption with key destruction for file-level erasure, or retire the drive with `blkdiscard` on an SSD.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1632_test.go
+++ b/pkg/katas/katatests/zc1632_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #652, opened 2026-04-18
feat: ZC1633 — flag `gpg --passphrase SECRET` passphrase in argv

afadesigns:
ZC1633 — Error on `gpg --passphrase SECRET` — passphrase on cmdline

What: flags `gpg` / `gpg2` invocations containing the bare `--passphrase` flag.
Why: `--passphrase VALUE` passes the key passphrase as an argv element. It ends up in `ps`, `/proc/<pid>/cmdline`, shell history, and audit logs — readable by any local user.
Fix suggestion: use `--passphrase-file PATH`, `--passphrase-fd N`, or `--pinentry-mode=loopback` with the passphrase piped on stdin. Pair with `--batch` for non-interactive runs.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1633_test.go
+++ b/pkg/katas/katatests/zc1633_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #653, opened 2026-04-18
feat: ZC1634 — flag `umask` leaving world-write bit (mask inversion footgun)

afadesigns:
ZC1634 — Warn on `umask NNN` that fails to mask world-write — mask-inversion footgun

What: flags `umask NNN` (3- or 4-digit octal) whose "other" digit is `0`, `1`, `4`, or `5` — i.e., values that do not set the write-bit mask. Already-covered `000` / `0` cases (ZC1195, ZC1516) are excluded.
Why: `umask` is a mask; set bits are REMOVED from the default permission. `umask 111` feels restrictive ("no execute for anyone") but it doesn't mask the write bit, so new files default to `666` (rw-rw-rw-). Every new file becomes world-writable.
Fix suggestion: use `022` for publicly readable files (file mode 644), or `077` for secrets-handling (file mode 600). The "other" digit must be `2`/`3`/`6`/`7` to strip world-write.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1634_test.go
+++ b/pkg/katas/katatests/zc1634_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #654, opened 2026-04-18
feat: ZC1635 — flag `mysql -pSECRET` / `--password=SECRET` credential in argv

afadesigns:
ZC1635 — Error on `mysql -pSECRET` / `--password=SECRET` — password in process list

What: flags `mysql` / `mysqldump` / `mysqladmin` / `mariadb` / `mariadb-dump` / `mariadb-admin` invocations where `-p` has a concatenated value (`-pSECRET`) or `--password=SECRET` is used.
Why: both forms place the password in argv. It shows up in `ps`, `/proc/<pid>/cmdline`, shell history, and audit logs for every local user who can list processes. Bare `-p` (no arg) prompts interactively and is safe.
Fix suggestion: use `-p` with no arg for interactive prompt, `--login-path=<group>` for the MySQL login-path credentials helper, or a `~/.my.cnf` with `0600` perms and `[client] password=...`.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1635_test.go
+++ b/pkg/katas/katatests/zc1635_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #655, opened 2026-04-18
feat: ZC1636 — flag `virsh destroy DOMAIN` force VM power-off

afadesigns:
ZC1636 — Warn on `virsh destroy DOMAIN` — force-stops VM (no graceful shutdown)

What: flags `virsh destroy DOM` where `--graceful` is absent.
Why: `destroy` is the libvirt equivalent of pulling the plug. The guest OS gets no chance to flush filesystems, close network connections, or run shutdown services — filesystem corruption on any open guest file.
Fix suggestion: use `virsh shutdown DOM` (ACPI event) and wait for completion. `virsh destroy --graceful DOM` tries graceful first then forces and is not flagged.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1636_test.go
+++ b/pkg/katas/katatests/zc1636_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #656, opened 2026-04-18
feat: ZC1637 — prefer Zsh `typeset -r` over POSIX `readonly`

afadesigns:
ZC1637 — Style: prefer Zsh `typeset -r NAME=value` over POSIX `readonly NAME=value`

What: flags `readonly NAME` / `readonly NAME=value` invocations.
Why: Zsh has a single, composable declaration verb: `typeset`. `typeset -r` creates a read-only parameter and composes with other flags — `-ir` for readonly integer, `-xr` for readonly-export, `-gr` for readonly global. `readonly` works but reads as POSIX / Bash in a Zsh codebase.
Fix suggestion: use `typeset -r NAME=value` (or `local -r NAME=value` inside a function).
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1637_test.go
+++ b/pkg/katas/katatests/zc1637_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #657, opened 2026-04-18
feat: ZC1638 — flag `docker build --build-arg SECRET=VALUE` layer leak

afadesigns:
ZC1638 — Error on `docker/podman/buildah build --build-arg SECRET=VALUE` — secret baked into image layer

What: flags `docker build` / `podman build` / `buildah build` with `--build-arg KEY=VALUE` where KEY contains one of `password`, `passwd`, `secret`, `token`, `apikey`, `api_key`, `access_key`, `private_key`.
Why: `--build-arg` values land in image metadata visible via `docker history`. The value is cached in the layer forever, even if the Dockerfile only used it as a build-time env var. Everyone who pulls the image reads the secret back.
Fix suggestion: use BuildKit secrets (`--secret id=NAME,src=PATH`) or a multi-stage build where the secret stays in a discarded stage.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1638_test.go
+++ b/pkg/katas/katatests/zc1638_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #658, opened 2026-04-18
feat: ZC1639 — flag `curl -H 'Authorization: ...'` credential header in argv

afadesigns:
ZC1639 — Error on `curl -H 'Authorization: ...'` — credential header in process list

What: flags `curl` / `wget` invocations where `-H` / `--header` carries a credential-bearing header (`Authorization`, `Proxy-Authorization`, `X-Api-Key`, `Api-Key`, `X-Auth-Token`, `X-Access-Token`, `Cookie`).
Why: the header value passes as an argv element. The expanded token shows up in `ps`, `/proc/<pid>/cmdline`, shell history, and audit logs.
Fix suggestion: use `curl -H @FILE` (reads header from a 0600-protected file) or `curl --config FILE` with `header = "..."` lines so the secret stays on disk, never on the command line.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1639_test.go
+++ b/pkg/katas/katatests/zc1639_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #659, opened 2026-04-18
feat: ZC1640 — prefer Zsh `${(P)var}` over Bash `${!var}` indirect expansion

afadesigns:
ZC1640 — Style: `${!var}` Bash indirect expansion — prefer Zsh `${(P)var}`

What: flags any argument containing the substring `${!` (Bash indirect parameter reference).
Why: Zsh has `${(P)var}` as the native equivalent. It composes with other parameter-expansion flags (`${(Pf)var}` to read the indirect value split on newlines, `${(PU)var}` to uppercase it, etc.) and reads consistently across Zsh docs.
Fix suggestion: replace `${!var}` with `${(P)var}`; `${!prefix*}` / `${!array[@]}` have Zsh equivalents via the `$parameters` hash or `(k)` subscript flags.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1640_test.go
+++ b/pkg/katas/katatests/zc1640_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #660, opened 2026-04-18
feat: ZC1641 — flag `kubectl create secret --from-literal` / `--docker-password`

afadesigns:
ZC1641 — Error on `kubectl create secret --from-literal=...` / `--docker-password=...`

What: flags `kubectl create secret ...` invocations containing `--from-literal=KEY=VALUE` or `--docker-password=VALUE`.
Why: both forms put the secret content in argv. The expanded value ends up in `ps`, `/proc/<pid>/cmdline`, shell history, and audit logs.
Fix suggestion: use `--from-file=KEY=PATH` (reads from a 0600-protected file), `--from-env-file=PATH` (KEY=VALUE lines), or pipe a manifest into `kubectl apply -f -` with base64-encoded `data:` values staged on disk.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1641_test.go
+++ b/pkg/katas/katatests/zc1641_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #661, opened 2026-04-18
feat: ZC1642 — flag `tshark -w` / `dumpcap -w` without `-Z USER`

afadesigns:
ZC1642 — Warn on `tshark -w FILE` / `dumpcap -w FILE` without `-Z USER` — capture file owned by root

What: flags `tshark -w FILE` / `dumpcap -w FILE` where `-Z USER` is absent.
Why: packet captures need `CAP_NET_RAW`, so the capture process typically runs as root. Without `-Z USER` the resulting pcap is root-owned. An analyst opening the file in Wireshark — which runs helper scripts from the capture — operates on a root-owned file and may unintentionally invoke things as root.
Fix suggestion: pass `-Z USER` so tshark/dumpcap drops privileges before writing the pcap.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1642_test.go
+++ b/pkg/katas/katatests/zc1642_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #662, opened 2026-04-18
feat: ZC1643 — prefer `$(<file)` over `$(cat file)` (no fork)

afadesigns:
ZC1643 — Style: `$(cat file)` — use `$(<file)` to skip the fork / exec

What: flags any argument whose raw text contains the substring `$(cat ` (command substitution that forks `/usr/bin/cat`).
Why: `$(<FILE)` is a shell builtin — it reads FILE directly into the command-substitution buffer with no fork, no exec, no wait. In a hot path the speedup is dramatic; in cold paths it still avoids one of the most common useless-use-of-cat patterns in review feedback.
Fix suggestion: replace `$(cat FILE)` with `$(<FILE)`.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1643_test.go
+++ b/pkg/katas/katatests/zc1643_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #664, opened 2026-04-18
feat: ZC1645 — prefer `/etc/os-release` over `lsb_release`

afadesigns:
ZC1645 — Style: `lsb_release` — prefer sourcing `/etc/os-release` (no dependency, no fork)

What: flags every `lsb_release` invocation.
Why: `lsb_release` is provided by the `lsb-release` / `redhat-lsb-core` package, missing on most minimal / container images (Alpine doesn't ship it at all). Scripts that depend on it fail on stripped images. `/etc/os-release` is always present on modern systemd-based Linux.
Fix suggestion: `source /etc/os-release; print -r -- $ID $VERSION_ID` — same info, no package dependency, no fork.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1645_test.go
+++ b/pkg/katas/katatests/zc1645_test.go
@@ -1,0 +1,53 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #665, opened 2026-04-18
feat: ZC1646 — flag `btrfs check --repair` / `xfs_repair -L` last-resort fs repair

afadesigns:
ZC1646 — Warn on `btrfs check --repair` / `xfs_repair -L` — last-resort recovery, may worsen damage

What: flags `btrfs check --repair ...` and `xfs_repair -L ...`.
Why: both are destructive last-resort fs recovery. `btrfs check --repair` explicitly warns it may cause additional filesystem damage — the btrfs devs ask users to try `btrfs scrub` and read-only `btrfs check` first. `xfs_repair -L` zeroes the log, dropping any uncommitted transactions.
Fix suggestion: snapshot the underlying block device before either command, and exhaust non-destructive checks first.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1646_test.go
+++ b/pkg/katas/katatests/zc1646_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #666, opened 2026-04-18
feat: ZC1647 — flag `kubectl apply -f URL` remote manifest without digest check

afadesigns:
ZC1647 — Warn on `kubectl apply -f URL` — remote manifest applied without digest verification

What: flags `kubectl {apply,create,replace} -f https://...yaml` / `http://...yaml`.
Why: the manifest is fetched from the network and applied to the cluster. TLS (when present) verifies transport but not authorship — if the URL is compromised or the file is rewritten, the cluster picks up whatever was substituted. Plain HTTP gives the attacker direct control of the response body.
Fix suggestion: download the manifest to disk, verify a known SHA256, then `kubectl apply -f local.yaml`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1647_test.go
+++ b/pkg/katas/katatests/zc1647_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #667, opened 2026-04-18
feat: ZC1648 — flag `cp /dev/null /var/log/` / `truncate -s 0 /var/log/` audit-log wipe

afadesigns:
ZC1648 — Error on `cp /dev/null /var/log/...` / `truncate -s 0 /var/log/...` — audit-log wipe

What: flags `cp /dev/null DEST` and `truncate -s 0 DEST` where DEST starts with `/var/log/`.
Why: replacing or zeroing a log under `/var/log/` erases audit evidence: failed logins in `auth.log`, sudo use in `sudo.log`, kernel audit trail in `audit/audit.log`, tty history in `wtmp` / `btmp`. Scripts doing this during "cleanup" are either misusing logrotate or deliberately covering tracks.
Fix suggestion: use `logrotate -f /etc/logrotate.d/<unit>` for rotation, `journalctl --vacuum-time=...` for the systemd journal.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1648_test.go
+++ b/pkg/katas/katatests/zc1648_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #668, opened 2026-04-18
feat: ZC1649 — flag `openssl` cert with validity > 825 days

afadesigns:
ZC1649 — Warn on `openssl req -days N` with N > 825 — long-validity certificate

What: flags `openssl req`, `openssl x509`, and `openssl ca` invocations where `-days N` is greater than 825.
Why: CA/Browser Forum capped public TLS cert validity at 825 days in 2018; major browsers tightened it to 398 days in 2020. A 10-year cert can't be revoked effectively — once the key leaks, the attacker keeps access until natural expiry.
Fix suggestion: keep leaf / server certs under 398 days and automate rotation. Long validity is defensible for internal root CAs only.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1649_test.go
+++ b/pkg/katas/katatests/zc1649_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #669, opened 2026-04-18
feat: ZC1650 — flag `setopt RM_STAR_SILENT` / `unsetopt RM_STAR_WAIT` prompt removal

afadesigns:
ZC1650 — Warn on `setopt RM_STAR_SILENT` / `unsetopt RM_STAR_WAIT` — removes `rm *` prompt

What: flags `setopt RM_STAR_SILENT` / `unsetopt RM_STAR_WAIT` (and their case / underscore variants).
Why: Zsh's default is to pause for 10 seconds on an interactive `rm *` and confirm. Disabling the prompt removes a safety net that has saved countless home directories.
Fix suggestion: keep the default `RM_STAR_WAIT` on. For scripted deletions where the prompt is a nuisance, pass `rm` through a subshell that scopes the option: `( setopt RM_STAR_SILENT; rm -- "${files[@]}" )`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1650_test.go
+++ b/pkg/katas/katatests/zc1650_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #671, opened 2026-04-18
fix: allow openssl flag names in typos allowlist

afadesigns:
Add `passin`, `passout`, and `unparseable` to the typos extend-words list so Check Spelling stops failing on OpenSSL password-source flag docs and ZC1620 description text.


redteamx (approved):
LGTM.

PR #673, opened 2026-04-18
feat: ZC1653 — flag `$BASHPID` Bash-only parameter

afadesigns:
ZC1653 — Avoid `$BASHPID` — Bash-only; Zsh uses `$sysparams[pid]` from `zsh/system`

What: flags arguments referencing `$BASHPID` / `${BASHPID}`.
Why: `$BASHPID` returns the current subshell's PID (while `$$` returns the parent shell's PID). Zsh does not set this parameter — scripts silently read empty and misbehave.
Fix suggestion: `zmodload zsh/system` once, then read `$sysparams[pid]` — updates inside subshells the same way.
Severity: Warning

Milestone: this PR takes the project to 650 katas.

### DIFF
--- a/pkg/katas/katatests/zc1653_test.go
+++ b/pkg/katas/katatests/zc1653_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #674, opened 2026-04-18
feat: ZC1654 — flag `sysctl -p /tmp/...` TOCTOU config load

afadesigns:
ZC1654 — Warn on `sysctl -p /tmp/...` — loading kernel tunables from attacker-writable path

What: flags `sysctl -p PATH` where PATH starts with `/tmp/` or `/var/tmp/`.
Why: `sysctl -p` reads `key=value` tunables from PATH and applies them. A `/tmp/` location is world-traversable — a concurrent local user can substitute the file between the script's write and sysctl's read, injecting `kernel.core_pattern=|/tmp/evil`, `kernel.modprobe=/tmp/evil`, or disabling hardening knobs.
Fix suggestion: keep sysctl configs under `/etc/sysctl.d/` with root ownership.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1654_test.go
+++ b/pkg/katas/katatests/zc1654_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #675, opened 2026-04-18
feat: ZC1655 — flag `read -n N` Bash-only character-count semantics

afadesigns:
ZC1655 — Warn on `read -n N` — Bash reads N chars; Zsh's `-n` means "drop newline"

What: flags `read -n N var` and `read -nN var` where N is an integer.
Why: Bash's `read -n N` reads exactly N characters. Zsh's `-n` is the "don't append trailing newline to the reply" flag and takes no count — `read -n 1 var` sets `var` to the whole input line instead of one character.
Fix suggestion: use `read -k N var` in Zsh for N-character reads.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1655_test.go
+++ b/pkg/katas/katatests/zc1655_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #676, opened 2026-04-18
feat: ZC1656 — flag rsync -e ssh host-key verify bypass

afadesigns:
ZC1656 — Error on `rsync -e 'ssh -o StrictHostKeyChecking=no'` — host-key verify disabled

What: `rsync -e "ssh -o StrictHostKeyChecking=no"` (or `UserKnownHostsFile=/dev/null`) ships the file stream over an ssh transport with host-key verification disabled.
Why: Same class as ZC1479 but easier to miss in review — the flags hide inside a quoted `-e` string. A MITM on the path impersonates the remote and the rsync stream goes straight through.
Fix suggestion: Pre-provision `~/.ssh/known_hosts` (or `ssh-keyscan` under review), keep `StrictHostKeyChecking=yes`.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1656_test.go
+++ b/pkg/katas/katatests/zc1656_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #677, opened 2026-04-18
feat: ZC1657 — flag semanage permissive -a SELinux domain disable

afadesigns:
ZC1657 — Warn on `semanage permissive -a <type>` — puts SELinux domain in permissive mode

What: `semanage permissive -a DOMAIN` / `--add` adds an SELinux type to the permissive-domain list.
Why: Narrower than `setenforce 0` but still disables enforcement for the named domain (often `httpd_t`, `container_t`, `sshd_t`) and persists across reboots because it is baked into policy.
Fix suggestion: Build an explicit allow rule with `audit2allow` (or a hand-written policy module), then remove the permissive mark with `semanage permissive -d DOMAIN`.
Severity: Warning

## Test plan
- valid `semanage permissive -d httpd_t` → no violation
- valid `semanage boolean -l` → no violation
- invalid `semanage permissive -a httpd_t` → ZC1657
- invalid `semanage permissive --add sshd_t` → ZC1657

### DIFF
--- a/pkg/katas/katatests/zc1657_test.go
+++ b/pkg/katas/katatests/zc1657_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #678, opened 2026-04-18
feat: ZC1658 — flag curl -OJ server-named output filename

afadesigns:
ZC1658 — Warn on `curl -OJ` / `-J -O` — server-controlled output filename

What: `curl -J` (`--remote-header-name`) + `-O` saves the response using whatever filename the server sends in `Content-Disposition`.
Why: Server (or anything on the path that can set headers — compromised CDN, reverse proxy, MITM) picks the destination name. Path traversal is sanitized, but same-directory overwrites (e.g., stomping `.zshrc` in the current dir) are not.
Fix suggestion: Use `-o NAME` with a filename the script controls, and validate the payload before acting on it.
Severity: Warning

## Test plan
- valid `curl -O https://example.com/file` → no violation
- valid `curl -o out.bin https://example.com/file` → no violation
- invalid `curl -OJ https://example.com/file` → ZC1658
- invalid `curl -O -J https://example.com/file` → ZC1658

### DIFF
--- a/pkg/katas/katatests/zc1658_test.go
+++ b/pkg/katas/katatests/zc1658_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #679, opened 2026-04-18
feat: ZC1659 — flag fuser -k path-target broad kill

afadesigns:
ZC1659 — Warn on `fuser -k <path>` — kills every process holding the subtree open

What: `fuser -k PATH` signals (SIGKILL by default) every process with a file open anywhere under PATH.
Why: On `/`, `/var`, `/tmp`, or any mount-root this reaches sshd, cron, dbus, and the caller's own shell. On a bind-mount it hits workloads that share the host inode. Blast radius is wider than the caller usually owns.
Fix suggestion: Target specific PIDs (`kill $(pidof app)`) or ports (`fuser -k 8080/tcp`), or `systemctl stop UNIT` for services.
Severity: Warning

## Test plan
- valid `fuser -k 8080/tcp` → no violation
- valid `fuser /var/log/syslog` (no -k) → no violation
- invalid `fuser -k /mnt` → ZC1659
- invalid `fuser -kim /` → ZC1659

### DIFF
--- a/pkg/katas/katatests/zc1659_test.go
+++ b/pkg/katas/katatests/zc1659_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #681, opened 2026-04-18
feat: ZC1661 — flag curl --cacert /dev/null empty trust store

afadesigns:
ZC1661 — Error on `curl --cacert /dev/null` — empty trust store, any cert passes

What: Pointing `--cacert` or `--capath` at `/dev/null` hands curl an empty trust-anchor set.
Why: Most TLS backends (OpenSSL, wolfSSL, Schannel) treat the peer cert as valid when no issuers are configured. It is the TLS-equivalent of `--insecure` with one more keystroke of plausible deniability.
Fix suggestion: Use a real bundle (`/etc/ssl/certs/ca-certificates.crt`) or `--pinnedpubkey sha256//…` for known endpoints.
Severity: Error

## Test plan
- valid `curl URL --cacert /etc/ssl/certs/ca-certificates.crt` → no violation
- valid `curl URL -o out` → no violation
- invalid `curl URL --cacert /dev/null` → ZC1661
- invalid `curl -s URL --capath /dev/null` → ZC1661

### DIFF
--- a/pkg/katas/katatests/zc1661_test.go
+++ b/pkg/katas/katatests/zc1661_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1661(t *testing.T) {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #682, opened 2026-04-18
feat: ZC1662 — flag pkexec env VAR=VAL root-env injection

afadesigns:
ZC1662 — Error on `pkexec env VAR=VAL CMD` — controlled env crossed into the root session

What: `pkexec env VAR=VALUE CMD` invokes `/usr/bin/env` as root with a caller-controlled environment.
Why: Polkit sanitizes a small allow-list, but once `env` takes over the remaining variables (`LD_PRELOAD`, `GCONV_PATH`, `PYTHONPATH`, `XDG_RUNTIME_DIR`, `LANGUAGE`) ride straight into root. CVE-2021-4034 used the same primitive via argv[0]; `env` makes it trivial.
Fix suggestion: Set required variables in a polkit rule or via `systemd-run --user` instead.
Severity: Error

## Test plan
- valid `pkexec /usr/bin/systemctl restart unit` → no violation
- valid `pkexec /usr/bin/apt install foo` → no violation
- invalid `pkexec env DISPLAY=:0 /usr/bin/cmd` → ZC1662
- invalid `pkexec env PATH=/tmp /usr/bin/cmd` → ZC1662

### DIFF
--- a/pkg/katas/katatests/zc1662_test.go
+++ b/pkg/katas/katatests/zc1662_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1662(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #684, opened 2026-04-18
feat: ZC1664 — flag systemctl set-default rescue/emergency persistent boot

afadesigns:
ZC1664 — Error on `systemctl set-default rescue.target|emergency.target` — persistent single-user boot

What: `systemctl set-default` rewrites `/etc/systemd/system/default.target` as a symlink to the named target.
Why: `rescue.target` / `emergency.target` means every boot drops to single-user mode before networking / sshd / normal units start — remote access is lost until reverted. Unlike `systemctl isolate` (one-shot, ZC1561), this persists across reboots.
Fix suggestion: Revert with `systemctl set-default multi-user.target` (servers) or `graphical.target` (desktops).
Severity: Error

## Test plan
- valid `systemctl set-default multi-user.target` → no violation
- valid `systemctl set-default graphical.target` → no violation
- invalid `systemctl set-default rescue.target` → ZC1664
- invalid `systemctl set-default emergency.target` → ZC1664

### DIFF
--- a/pkg/katas/katatests/zc1664_test.go
+++ b/pkg/katas/katatests/zc1664_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #687, opened 2026-04-18
feat: ZC1667 — flag openssl enc without -pbkdf2 legacy KDF

afadesigns:
ZC1667 — Warn on `openssl enc` without `-pbkdf2` — legacy MD5-based key derivation

What: Without `-pbkdf2`, `openssl enc` derives the symmetric key via EVP_BytesToKey (single-round MD5 over `password || salt`).
Why: A modern GPU cracks single-MD5 passphrases at billions of guesses per second. `-aes-256-gcm` with `openssl enc` also silently drops the auth tag — so there is no AEAD path through this command at all.
Fix suggestion: Add `-pbkdf2 -iter 100000` (OpenSSL 1.1.1+). For new code, prefer `age`, `gpg --symmetric --cipher-algo AES256`, or `openssl smime`.
Severity: Warning

## Test plan
- valid `openssl enc -aes-256-cbc -pbkdf2 -iter 100000 …` → no violation
- valid `openssl req -new …` → no violation
- invalid `openssl enc -aes-256-cbc -salt -in file -out file.enc` → ZC1667
- invalid `openssl enc -aes-256-gcm -in file -out file.enc` → ZC1667

### DIFF
--- a/pkg/katas/katatests/zc1667_test.go
+++ b/pkg/katas/katatests/zc1667_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1667(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #688, opened 2026-04-18
feat: ZC1668 — flag aws iam attach-*-policy AdministratorAccess

afadesigns:
ZC1668 — Error on `aws iam attach-*-policy ... AdministratorAccess` — grants full AWS admin

What: Attaching the AWS-managed `AdministratorAccess` / `PowerUserAccess` / `IAMFullAccess` policy to a user/role/group.
Why: Grants `*:*` — IAM mutation, KMS key rotation, S3 exfiltration, root-password reset. Scripts almost never need it; usually the signal is someone replacing a scoped policy after hitting a permissions error.
Fix suggestion: Write a least-privilege inline policy (`iam put-user-policy --policy-document`) or reference a customer-managed policy with the minimum `Action`/`Resource` pairs. Admin attachment should land via change-reviewed Terraform, not a shell loop.
Severity: Error

## Test plan
- valid `aws iam attach-user-policy … ReadOnlyAccess` → no violation
- valid `aws iam create-access-key --user-name foo` → no violation
- invalid `aws iam attach-user-policy … AdministratorAccess` → ZC1668
- invalid `aws iam attach-role-policy … PowerUserAccess` → ZC1668

### DIFF
--- a/pkg/katas/katatests/zc1668_test.go
+++ b/pkg/katas/katatests/zc1668_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1668(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected []katas.Violation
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #689, opened 2026-04-18
feat: ZC1669 — flag git gc --prune=now / reflog expire recovery wipe

afadesigns:
ZC1669 — Warn on `git gc --prune=now` / `git reflog expire --expire=now` — deletes recovery window

What: Both commands bulldoze the default `gc.reflogExpire` (90 d) and `gc.pruneExpire` (2 w) grace periods, so dropped commits become unrecoverable immediately.
Why: A stray interactive rebase, a bad `git reset`, or a dropped cherry-pick usually survives because `git reflog` still points at the old SHA. With `--prune=now` / `--expire=now`, no safety net.
Fix suggestion: Keep the default cadence. Only run the destructive form to purge leaked secrets, and mirror the dropped history off-box first.
Severity: Warning

## Test plan
- valid `git gc` → no violation
- valid `git gc --auto` → no violation
- invalid `git gc --prune=now --aggressive` → ZC1669
- invalid `git reflog expire --expire=now --all` → ZC1669

### DIFF
--- a/pkg/katas/katatests/zc1669_test.go
+++ b/pkg/katas/katatests/zc1669_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1669(t *testing.T) {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #693, opened 2026-04-18
feat: ZC1673 — prefer Zsh read -s over stty -echo bracketing

afadesigns:
ZC1673 — Style: `stty -echo` around `read` — prefer Zsh `read -s`

What: `stty -echo; IFS= read -r password; stty echo` to mask password entry.
Why: A crash or SIGINT between the two `stty` calls leaves the user's terminal stuck in echo-off — silent and confusing. `read -s` (Zsh / Bash 4+) disables echo only for that one read and restores it even on interrupt, with no external forks.
Fix suggestion: `read -s password` (or `read -ks` for single-key prompt) — drop the `stty` bracketing.
Severity: Style

## Test plan
- valid `stty echo` → no violation
- valid `stty raw` → no violation
- invalid `stty -echo` → ZC1673

### DIFF
--- a/pkg/katas/katatests/zc1673_test.go
+++ b/pkg/katas/katatests/zc1673_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #694, opened 2026-04-18
feat: ZC1674 — flag docker/podman run --oom-kill-disable / --oom-score-adj abuse

afadesigns:
ZC1674 — Warn on `docker/podman run --oom-kill-disable` or `--oom-score-adj <= -500`

What: `--oom-kill-disable` (kernel OOM killer never touches the container) or `--oom-score-adj <= -500` (preferentially kill others).
Why: A memory leak inside then drives the host into OOM reclaim until sshd / journald / init itself is killed. Even at -500 the container shifts pressure onto unrelated host processes.
Fix suggestion: Cap memory with `--memory=<limit>` so the container is killed on overrun; shift heavy workloads to a dedicated node instead of rigging OOM scores.
Severity: Warning

## Test plan
- valid `docker run alpine` → no violation
- valid `docker run --oom-score-adj 0 alpine` → no violation
- invalid `docker run --oom-kill-disable alpine` → ZC1674
- invalid `podman run --oom-score-adj=-1000 alpine` → ZC1674

### DIFF
--- a/pkg/katas/katatests/zc1674_test.go
+++ b/pkg/katas/katatests/zc1674_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1674(t *testing.T) {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #695, opened 2026-04-18
feat: ZC1675 — flag Bash-only export -f / -n in Zsh scripts

afadesigns:
ZC1675 — Avoid Bash-only `export -f` / `export -n` — use Zsh `typeset -fx` / `typeset +x`

What: `export -f FUNC` (function export) and `export -n VAR` (remove-export) are Bash-only.
Why: Zsh's `export` ignores `-f` and errors on `-n` — scripts that rely on either silently break under Zsh.
Fix suggestion: Use `typeset -fx FUNC` for function export (or `autoload -Uz` from an `fpath` directory, which is better), and `typeset +x VAR` to drop the export flag.
Severity: Info

## Test plan
- valid `export VAR=value` → no violation
- valid `export PATH HOME` → no violation
- invalid `export -f my_func` → ZC1675
- invalid `export -n VAR` → ZC1675

### DIFF
--- a/pkg/katas/katatests/zc1675_test.go
+++ b/pkg/katas/katatests/zc1675_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #696, opened 2026-04-18
feat: ZC1676 — flag helm rollback --force in-flight resource kick

afadesigns:
ZC1676 — Warn on `helm rollback --force` — recreates in-flight resources, corrupts rolling updates

What: `helm rollback RELEASE N --force` asks Helm to delete and recreate any resource it cannot patch cleanly.
Why: During a mid-rollout, `--force` takes out both old and new ReplicaSets, kicks pods, bypasses `PodDisruptionBudget`, and loses in-flight requests. Rolling back to an older revision also reintroduces any CVEs the later revisions fixed.
Fix suggestion: Pin the target revision explicitly, drop `--force`, and gate the rollback behind a change-review ticket.
Severity: Warning

## Test plan
- valid `helm rollback myapp 2` → no violation
- valid `helm history myapp` → no violation
- invalid `helm rollback myapp 2 --force` → ZC1676

### DIFF
--- a/pkg/katas/katatests/zc1676_test.go
+++ b/pkg/katas/katatests/zc1676_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #697, opened 2026-04-18
feat: ZC1677 — flag trap 'set -x' DEBUG global xtrace secret leak

afadesigns:
ZC1677 — Warn on `trap 'set -x' DEBUG` — xtrace on every command leaks secrets

What: `trap 'set -x' DEBUG` runs the trap handler before every simple command, turning on xtrace for the rest of the shell.
Why: Every subsequent `curl -H 'Authorization: Bearer …'`, `mysql -p<password>`, or `aws configure set …` prints its full argv to stderr, typically into a log file or CI artifact. Same antipattern with `set -o xtrace` inside a DEBUG trap.
Fix suggestion: Trace a narrow block with `set -x … set +x` around the section you want, or use Zsh `typeset -ft FUNC` for function-level xtrace without argv-spray.
Severity: Warning

## Test plan
- valid `trap 'cleanup' EXIT` → no violation
- valid `trap 'set -x' ERR` → no violation
- invalid `trap 'set -x' DEBUG` → ZC1677
- invalid `trap "set -o xtrace" DEBUG` → ZC1677

### DIFF
--- a/pkg/katas/katatests/zc1677_test.go
+++ b/pkg/katas/katatests/zc1677_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1677(t *testing.T) {
+	tests := []struct {
+		name     string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #701, opened 2026-04-18
feat: ZC1681 — flag tar -P absolute-names extract over host paths

afadesigns:
ZC1681 — Error on `tar -P` / `--absolute-names` — archive absolute paths, can overwrite host files

What: By default GNU tar strips the leading `/` from archive paths. `-P` / `--absolute-names` disables the strip.
Why: `tar -xPf evil.tar` writes to `/etc/cron.d/evil`, `/usr/local/bin/sshd`, or any absolute path the archive mentions — a trivial host-takeover on an untrusted tarball.
Fix suggestion: Drop the flag. Extract with `-C <scratch-dir>`, audit the tree, then copy into place with `install` or `cp`.
Severity: Error

## Test plan
- valid `tar -xf archive.tar` → no violation
- valid `tar -xvzf archive.tgz` → no violation
- invalid `tar -xPf archive.tar` → ZC1681
- invalid `tar xf archive.tar -P` → ZC1681

### DIFF
--- a/pkg/katas/katatests/zc1681_test.go
+++ b/pkg/katas/katatests/zc1681_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #702, opened 2026-04-18
feat: ZC1682 — flag npm install --unsafe-perm root lifecycle script

afadesigns:
ZC1682 — Error on `npm install --unsafe-perm` — npm lifecycle scripts keep root privileges

What: `--unsafe-perm` (or `--unsafe-perm=true`) tells npm to skip its UID-drop before running `preinstall`/`install`/`postinstall` lifecycle scripts.
Why: The UID drop exists because a compromised dependency executes arbitrary code via those hooks. With `--unsafe-perm`, that code runs as the current UID — typically root during provisioning.
Fix suggestion: Build in a dedicated builder container, or `sudo -u builduser npm install` from a non-root account that already owns `node_modules/`.
Severity: Error

## Test plan
- valid `npm ci` → no violation
- valid `yarn install` → no violation
- invalid `npm install --unsafe-perm` → ZC1682
- invalid `npm install --unsafe-perm=true` → ZC1682

### DIFF
--- a/pkg/katas/katatests/zc1682_test.go
+++ b/pkg/katas/katatests/zc1682_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #703, opened 2026-04-18
feat: ZC1683 — flag npm/yarn/pnpm config set registry plaintext HTTP

afadesigns:
ZC1683 — Error on `npm/yarn/pnpm config set registry http://...` — plaintext package index

What: Pointing a JS package manager at an `http://` registry disables TLS during fetch.
Why: Any host on the path (corporate proxy, hotel Wi-Fi, compromised CDN) can rewrite tarballs mid-flight; lockfile hashes catch the rewrite only if every dep was locked pre-swap.
Fix suggestion: Pin to `https://`. If the internal registry uses a custom CA, point to it via `NODE_EXTRA_CA_CERTS` or `registry.cafile` instead of falling back to HTTP.
Severity: Error

## Test plan
- valid `npm config set registry https://registry.npmjs.org/` → no violation
- valid `npm config set strict-ssl false` (different setting) → no violation
- invalid `npm config set registry http://internal.example.com/` → ZC1683
- invalid `yarn config set registry http://internal/` → ZC1683

### DIFF
--- a/pkg/katas/katatests/zc1683_test.go
+++ b/pkg/katas/katatests/zc1683_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1683(t *testing.T) {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #705, opened 2026-04-18
feat: ZC1685 — flag sleep infinity SIGTERM-insensitive keep-alive

afadesigns:
ZC1685 — Info: `sleep infinity` — container keep-alive pattern that ignores SIGTERM

What: `sleep infinity` as a container or systemd-unit keep-alive.
Why: GNU `sleep` installs no SIGTERM handler. `docker stop` / `systemctl stop` sends SIGTERM and the process sits unresponsive until grace-period expiry, then SIGKILL. The orchestrator reports a hung stop; any cleanup in a wrapping shell's signal traps never runs.
Fix suggestion: `exec tail -f /dev/null` (signal-handles cleanly), or front the container with `tini` / `dumb-init` when PID 1 must stay.
Severity: Info

## Test plan
- valid `sleep 30` → no violation
- valid `sleep $timeout` → no violation
- invalid `sleep infinity` → ZC1685

### DIFF
--- a/pkg/katas/katatests/zc1685_test.go
+++ b/pkg/katas/katatests/zc1685_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #706, opened 2026-04-18
feat: ZC1686 — flag compinit -C / -u fpath integrity bypass

afadesigns:
ZC1686 — Warn on `compinit -C` / `compinit -u` — skips / ignores `$fpath` integrity checks

What: Zsh's completion system loads every file from `$fpath` as shell code. `compinit` normally refuses when an `$fpath` entry is writable by someone other than the current user or root.
Why: `-C` skips the security check entirely (for speed); `-u` loads the insecure files anyway. Either way, a world-writable `$fpath` dir becomes an execution primitive for any local user.
Fix suggestion: Audit `$fpath` with `compaudit`, fix ownership/permissions, then run plain `compinit` (or `compinit -d <cache>` for cache-path control).
Severity: Warning

## Test plan
- valid `compinit -d $XDG_CACHE_HOME/zcompdump` → no violation
- valid `compaudit` → no violation (different command)
- invalid `compinit -C` → ZC1686
- invalid `compinit -u` → ZC1686

### DIFF
--- a/pkg/katas/katatests/zc1686_test.go
+++ b/pkg/katas/katatests/zc1686_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1686(t *testing.T) {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #707, opened 2026-04-18
feat: ZC1687 — flag snap install --classic / --devmode sandbox weakening

afadesigns:
ZC1687 — Warn on `snap install --classic` / `--devmode` — weakens snap confinement

What: `--classic` drops the AppArmor / cgroup / seccomp sandbox entirely; `--devmode` keeps it wired but logs violations instead of blocking.
Why: Documented escape hatches for snaps that cannot fit strict confinement, but in provisioning scripts they usually mean "I could not be bothered to pick a strict snap."
Fix suggestion: Find a strict snap, install from the distro repo with proper AppArmor, or document which specific snap requires the elevation.
Severity: Warning

## Test plan
- valid `snap install hello-world` → no violation
- valid `snap list` → no violation
- invalid `snap install code --classic` → ZC1687
- invalid `snap install pkg --devmode` → ZC1687

### DIFF
--- a/pkg/katas/katatests/zc1687_test.go
+++ b/pkg/katas/katatests/zc1687_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #708, opened 2026-04-18
feat: ZC1688 — flag aws s3 sync --delete destination-wipe risk

afadesigns:
ZC1688 — Warn on `aws s3 sync --delete` — destination objects deleted when source diverges

What: `aws s3 sync SRC DST --delete` removes every object in DST that does not exist under SRC.
Why: A misspelled SRC, an empty build directory, or a stale `cd` turns the sync into a full-bucket wipe — no second confirmation, no recovery unless versioning is enabled.
Fix suggestion: Restrict deletion to the changed prefix (`aws s3 sync ./build s3://bucket/app/ --delete`), add `--dryrun` behind a gate, or enable versioning + MFA-delete before running from a pipeline.
Severity: Warning

## Test plan
- valid `aws s3 sync ./build s3://my-bucket/` → no violation
- valid `aws s3 cp file s3://bucket/` → no violation
- invalid `aws s3 sync ./build s3://my-bucket/ --delete` → ZC1688
- invalid `aws s3 sync s3://src/ s3://dst/ --delete` → ZC1688

### DIFF
--- a/pkg/katas/katatests/zc1688_test.go
+++ b/pkg/katas/katatests/zc1688_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1688(t *testing.T) {
+	tests := []struct {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #711, opened 2026-04-18
feat: ZC1691 — flag rsync --remove-source-files optimistic delete

afadesigns:
ZC1691 — Warn on `rsync --remove-source-files` — SRC deletion tied to optimistic success

What: `rsync --remove-source-files` deletes each source file after rsync reports per-file transfer success.
Why: The success check is generous: remote out-of-disk after the partial write, `--chmod` rejection, or a flaky network that drops after data bytes but before metadata can still look like success. Combined with a wrong DST, the source is gone.
Fix suggestion: Two-step flow — `rsync -a SRC DST`, verify DST (checksums / file count), then `rm` explicitly. For local moves, use `mv`.
Severity: Warning

## Test plan
- valid `rsync -av src/ dst/` → no violation
- valid `rsync -av --delete src/ dst/` → no violation
- invalid `rsync -av --remove-source-files src/ dst/` → ZC1691
- invalid `rsync -a --remove-source-files host:src dst` → ZC1691

### DIFF
--- a/pkg/katas/katatests/zc1691_test.go
+++ b/pkg/katas/katatests/zc1691_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1691(t *testing.T) {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #713, opened 2026-04-18
feat: ZC1693 — flag ionice -c 1 real-time I/O starvation risk

afadesigns:
ZC1693 — Warn on `ionice -c 1` — real-time I/O class starves every other disk consumer

What: `ionice -c 1` (real-time I/O scheduling class) promotes the child above every best-effort (class 2) and idle (class 3) task queued for the same device.
Why: A busy `rsync` / `dd` / DB backup then blocks sshd reads, journald writes, and every other process until it yields — which for sequential I/O is effectively never.
Fix suggestion: Stay on class 2 (default). Reserve class 1 for latency-critical paths launched by a scheduler that caps duration.
Severity: Warning

## Test plan
- valid `ionice -c 2 -n 4 cmd` → no violation
- valid `ionice -c 3 cmd` → no violation
- invalid `ionice -c 1 -n 0 cmd` → ZC1693
- invalid `ionice -c1 cmd` → ZC1693

### DIFF
--- a/pkg/katas/katatests/zc1693_test.go
+++ b/pkg/katas/katatests/zc1693_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #714, opened 2026-04-18
feat: ZC1694 — flag ssh -A auth-socket forwarding to remote host

afadesigns:
ZC1694 — Warn on `ssh -A` / `-o ForwardAgent=yes` — remote host can reuse local keys

What: `ssh -A` (and `-o ForwardAgent=yes`) forwards the caller's `SSH_AUTH_SOCK` into the remote session.
Why: Anyone with root on the remote — and any process that shares its uid — can read the socket and impersonate the caller against every host the local keys unlock.
Fix suggestion: Use `ssh -J jumphost target` (ProxyJump) for multi-hop access; keys stay local. Or scope a dedicated key for the remote task with `ssh-copy-id`. Keep key forwarding for interactive use on trusted hosts.
Severity: Warning

## Test plan
- valid `ssh user@host` → no violation
- valid `ssh -J bastion user@target` → no violation
- invalid `ssh -A user@host` → ZC1694
- invalid `ssh -o ForwardAgent=yes user@host` → ZC1694

### DIFF
--- a/pkg/katas/katatests/zc1694_test.go
+++ b/pkg/katas/katatests/zc1694_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1694(t *testing.T) {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #716, opened 2026-04-18
feat: ZC1696 — flag pnpm/yarn --no-frozen-lockfile / --no-immutable drift

afadesigns:
ZC1696 — Warn on `pnpm install --no-frozen-lockfile` / `yarn install --no-immutable` — CI lockfile drift

What: Both flags tell the package manager that the lockfile is merely a suggestion — any resolution change since the lockfile was written gets picked up silently.
Why: Run from CI and the artifact no longer matches the dependency graph reviewers signed off on.
Fix suggestion: `pnpm install --frozen-lockfile` (CI default) or `yarn install --immutable`. Let lockfile regeneration happen only from a dev workstation PR.
Severity: Warning

## Test plan
- valid `pnpm install --frozen-lockfile` → no violation
- valid `npm ci` → no violation
- invalid `pnpm install --no-frozen-lockfile` → ZC1696
- invalid `yarn install --no-immutable` → ZC1696

### DIFF
--- a/pkg/katas/katatests/zc1696_test.go
+++ b/pkg/katas/katatests/zc1696_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #717, opened 2026-04-18
feat: ZC1697 — flag cryptsetup --allow-discards sector-layout leak

afadesigns:
ZC1697 — Info: `cryptsetup open --allow-discards` — TRIM pass-through leaks free-sector map

What: `--allow-discards` forwards TRIM/DISCARD commands from the filesystem to the underlying SSD through dm-crypt.
Why: Real gains in performance and wear-levelling, real trade-off in deniability: an attacker with raw-device access sees which blocks are empty, fingerprinting partition layouts and defeating plausible-deniability scenarios.
Fix suggestion: Drop `--allow-discards` if offline-disk inspection is in scope; otherwise keep it but document the trade-off in the runbook.
Severity: Info

## Test plan
- valid `cryptsetup open $DISK data` → no violation
- valid `cryptsetup luksClose data` → no violation
- invalid `cryptsetup open --allow-discards $DISK data` → ZC1697
- invalid `cryptsetup luksOpen --allow-discards $DISK data` → ZC1697

### DIFF
--- a/pkg/katas/katatests/zc1697_test.go
+++ b/pkg/katas/katatests/zc1697_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1697(t *testing.T) {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #718, opened 2026-04-18
feat: ZC1698 — flag fail2ban-client unban --all / stop wipes ban list

afadesigns:
ZC1698 — Warn on `fail2ban-client unban --all` / `stop` — wipes the active brute-force ban list

What: `fail2ban-client unban --all` clears every active ban across every jail. `fail2ban-client stop` shuts the service down and flushes its rules.
Why: Either command restores network access for the exact attacker IPs fail2ban has already flagged as hostile — usually hundreds of known bots.
Fix suggestion: Target a single IP with `fail2ban-client set <jail> unbanip <ip>`, or reload a jail with `reload <jail>` when picking up new filter rules.
Severity: Warning

## Test plan
- valid `fail2ban-client status` → no violation
- valid `fail2ban-client set sshd unbanip 1.2.3.4` → no violation
- invalid `fail2ban-client unban --all` → ZC1698
- invalid `fail2ban-client stop` → ZC1698

### DIFF
--- a/pkg/katas/katatests/zc1698_test.go
+++ b/pkg/katas/katatests/zc1698_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #723, opened 2026-04-18
feat: ZC1703 — flag sysctl disabling network-hardening knobs

afadesigns:
ZC1703 — Warn on `sysctl -w` disabling network-hardening knobs

What: `net.ipv4.*` / `net.ipv6.*` knobs exist to harden against spoofing, ICMP redirect tampering, smurf amplification, source-routed packets — `rp_filter=1`, `accept_source_route=0`, `accept_redirects=0`, `send_redirects=0`, `icmp_echo_ignore_broadcasts=1`, `log_martians=1`.
Why: Flipping any to the lax value re-opens classic layer-3 attacks.
Fix suggestion: Keep the protective defaults. If a niche workload really needs relaxed filtering, scope the change per-interface with a comment explaining why.
Severity: Warning

## Test plan
- valid `sysctl -w net.ipv4.conf.all.rp_filter=1` → no violation
- valid `sysctl -w net.ipv4.tcp_syncookies=1` → no violation
- invalid `sysctl -w net.ipv4.conf.all.rp_filter=0` → ZC1703
- invalid `sysctl -w net.ipv4.conf.all.accept_source_route=1` → ZC1703

### DIFF
--- a/pkg/katas/katatests/zc1703_test.go
+++ b/pkg/katas/katatests/zc1703_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1703(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #725, opened 2026-04-18
feat: ZC1705 — flag awk -i inplace gawk-only portability footgun

afadesigns:
ZC1705 — Info: `awk -i inplace` is gawk-only — script breaks on mawk / BSD awk

What: The `inplace` extension that powers `awk -i inplace` ships only with gawk.
Why: On Alpine (default mawk), Debian-busybox, macOS, BSDs, and any container without `gawk` installed, the script aborts with `fatal: can't open extension 'inplace'`.
Fix suggestion: For portability rewrite as `awk … input > tmp && mv tmp input`. If gawk is required, document it and add `command -v gawk >/dev/null` at script top.
Severity: Info

## Test plan
- valid `awk '{print}' file` → no violation
- valid `awk -i /usr/share/awk/lib.awk '{print}' file` → no violation
- invalid `awk -i inplace '{print}' file` → ZC1705
- invalid `gawk -i inplace -v x=1 '{print x}' file` → ZC1705

### DIFF
--- a/pkg/katas/katatests/zc1705_test.go
+++ b/pkg/katas/katatests/zc1705_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #727, opened 2026-04-18
feat: ZC1707 — flag gpg --keyserver hkp:// plaintext key fetch

afadesigns:
ZC1707 — Warn on `gpg --keyserver hkp://…` — plaintext keyserver fetch

What: `hkp://` is the unencrypted HKP keyserver protocol.
Why: A MITM on the path (corporate proxy, hotel Wi-Fi, hostile router) can swap key bytes during the fetch and `gpg --recv-keys` happily imports the substitute.
Fix suggestion: Use `hkps://keys.openpgp.org` (TLS) or fetch the armored key over HTTPS and verify the fingerprint out-of-band before `gpg --import`.
Severity: Warning

## Test plan
- valid `gpg ABCD --keyserver hkps://keys.openpgp.org --recv-keys` → no violation
- valid `gpg --recv-keys ABCD` (default keyserver) → no violation
- invalid `gpg ABCD --keyserver hkp://keys.example.com --recv-keys` → ZC1707

### DIFF
--- a/pkg/katas/katatests/zc1707_test.go
+++ b/pkg/katas/katatests/zc1707_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1707(t *testing.T) {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #728, opened 2026-04-18
feat: ZC1708 — flag find -L with -delete / -exec symlink follow into unintended tree

afadesigns:
ZC1708 — Error on `find -L ... -delete` / `-exec rm` — symlink follow into unintended trees

What: `find -L` follows symlinks. Combined with `-delete` (or `-exec rm`), a symlink in the start path that points outside steers find / unlinks files in `/etc`, `/var/lib`, etc.
Why: Trivial path-escape via planted symlinks. Common in log-rotation one-liners.
Fix suggestion: Drop `-L` (default `-P` keeps symlinks as objects), add `-xdev`/`-mount`, scope with `-type f`. For log rotation prefer `logrotate`.
Severity: Error

## Test plan
- valid `find /var/log -mtime +30 -delete` → no violation
- valid `find -L /opt -name '*.bak'` → no violation
- invalid `find -L /var/log -mtime +30 -delete` → ZC1708
- invalid `find -L /var/log -mtime +30 -exec rm -f` → ZC1708

### DIFF
--- a/pkg/katas/katatests/zc1708_test.go
+++ b/pkg/katas/katatests/zc1708_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1708(t *testing.T) {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #735, opened 2026-04-18
feat: ZC1715 — flag `read -p` (Zsh coprocess clash, not a prompt)

afadesigns:
ZC1715 — `read -p` collides with Zsh coprocess semantics

What: Detect `read -p` (and combined short flags like `-rp`, `-rsp`) — Bash uses `-p` for the prompt string, Zsh uses it for "read from coprocess".
Why: In Zsh, `read -p "Prompt: " var` errors with `no coprocess` and leaves the variable empty, silently breaking the script.
Fix suggestion: Use the Zsh idiom `read "var?Prompt: "` — the `?` after the variable name introduces the prompt.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1715_test.go
+++ b/pkg/katas/katatests/zc1715_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #736, opened 2026-04-18
feat: ZC1716 — flag `uname -m` / `-p` (use $CPUTYPE / $MACHTYPE)

afadesigns:
ZC1716 — `uname -m` / `uname -p` for architecture

What: Detect `uname -m` and `uname -p` calls.
Why: Zsh maintains `$CPUTYPE` (e.g. `x86_64`) and `$MACHTYPE` (GNU triplet) populated at startup from the same `uname(2)` call. Forking `uname` for the same answer is wasted work.
Fix suggestion: Read `$CPUTYPE` (architecture) or `$MACHTYPE` (full triplet) instead.
Severity: Style

### DIFF
--- a/pkg/katas/katatests/zc1716_test.go
+++ b/pkg/katas/katatests/zc1716_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #738, opened 2026-04-18
feat: ZC1718 — flag `gh secret set --body SECRET` (secret in argv)

afadesigns:
ZC1718 — `gh secret set --body SECRET`

What: Detect `gh secret set NAME --body VALUE`, `--body=VALUE`, or `-b VALUE`.
Why: The secret value lands in argv — visible in `ps`, `/proc/<pid>/cmdline`, history, host audit logs.
Fix suggestion: Use `--body-file PATH` or `--body -` with `printf %s "$SECRET" |` piped to stdin.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1718_test.go
+++ b/pkg/katas/katatests/zc1718_test.go
@@ -1,0 +1,75 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #742, opened 2026-04-18
feat: ZC1722 — flag `ssh-keyscan HOST >> known_hosts` (TOFU bypass)

afadesigns:
ZC1722 — `ssh-keyscan HOST >> known_hosts`

What: Detect `ssh-keyscan` output redirected (`>>` or `>`) into a `known_hosts` file.
Why: Accepts whatever host key the remote serves on first contact — exactly the case the host-key check guards against. MITM on first contact wins permanently.
Fix suggestion: Pin the expected fingerprint via a side-channel and verify with `ssh-keyscan HOST | ssh-keygen -lf -` before the append.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1722_test.go
+++ b/pkg/katas/katatests/zc1722_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #744, opened 2026-04-18
feat: ZC1724 — flag `pacman -Sy <pkg>` (partial-upgrade footgun)

afadesigns:
ZC1724 — `pacman -Sy <pkg>`

What: Detect `pacman -Sy <pkg>` (or any `-Sy*` flag combination without `u`) followed by a package name.
Why: Refreshes the package DB and installs ONE package against newer metadata while every other installed package stays at its old version. Dependency closure pulls libraries newer than the rest of the system → `error while loading shared libraries`.
Fix suggestion: Use `pacman -Syu` first, then `pacman -S <pkg>`. For CI, `pacman -Syu --noconfirm <pkg>` keeps the upgrade and install atomic.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1724_test.go
+++ b/pkg/katas/katatests/zc1724_test.go
@@ -1,0 +1,51 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #748, opened 2026-04-18
feat: ZC1728 — flag `pip install --index-url http://...` (plaintext index, MITM)

afadesigns:
ZC1728 — `pip install --index-url http://...`

What: Detect `pip install --index-url`, `--extra-index-url`, or `-i` followed by an `http://` URL (joined `=` form too).
Why: Plaintext HTTP fetches let any MITM swap package metadata or wheel contents — direct code execution on the install host.
Fix suggestion: Use `https://`. On internal networks, terminate TLS at the mirror and only configure the `https://` URL.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1728_test.go
+++ b/pkg/katas/katatests/zc1728_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #749, opened 2026-04-18
feat: ZC1729 — flag `ip route flush all` / `del default` (network lockout)

afadesigns:
ZC1729 — `ip route flush all` / `ip route del default`

What: Detect `ip route flush all`, `ip route flush table ...`, and `ip route del default` (handles `ip -4`, `-6`, etc.).
Why: Wipes the default gateway — the SSH session that just ran the command loses connectivity, subsequent steps hang.
Fix suggestion: Scope the flush (`flush dev <iface>`) or use `ip route replace default via <gw>` so the new route is in place before the old one disappears.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1729_test.go
+++ b/pkg/katas/katatests/zc1729_test.go
@@ -1,0 +1,75 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #750, opened 2026-04-18
feat: ZC1730 — flag `brew install --HEAD` (unrepeatable upstream-HEAD build)

afadesigns:
ZC1730 — `brew install --HEAD <pkg>`

What: Detect `brew install`, `reinstall`, or `upgrade` with `--HEAD`.
Why: Builds the formula from the upstream source repository's HEAD branch — every run pulls a different commit, no version stability, and any compromised upstream lands directly on the host.
Fix suggestion: Pin to a stable formula release; if HEAD is genuinely required, vendor a private tap with a fixed revision.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1730_test.go
+++ b/pkg/katas/katatests/zc1730_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #751, opened 2026-04-18
feat: ZC1731 — flag `curl -d 'password=…'` / `wget --post-data` (secret in argv)

afadesigns:
ZC1731 — POST body credentials in argv

What: Detect `curl -d / --data / --data-raw / --data-urlencode / --data-binary` and `wget --post-data / --body-data` whose value contains a credential-looking key (`password`, `secret`, `token`, `apikey`, `access_key`, `private_key`).
Why: POST body lands in argv — visible in `ps`, `/proc/<pid>/cmdline`, history, CI logs. Same leak surface as inline `-u user:pass`.
Fix suggestion: Read the value from a file (`curl --data @secret.txt URL`) or pipe via stdin (`--data-binary @-`).
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1731_test.go
+++ b/pkg/katas/katatests/zc1731_test.go
@@ -1,0 +1,75 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #754, opened 2026-04-18
feat: ZC1734 — flag overwrites of `/etc/passwd|shadow|group|gshadow` (lock bypass)

afadesigns:
ZC1734 — `cp/mv/tee` overwriting user-identity files

What: Detect `cp`, `mv`, `tee`, `install`, `dd` writing to `/etc/passwd|shadow|group|gshadow`, plus `>` / `>>` redirects to those paths from any command.
Why: Bypasses the lock that `vipw` / `vigr` / `useradd` use — concurrent edits race, malformed entries lock everyone out, shadow file falls out of sync.
Fix suggestion: Edit through `vipw -e` / `vigr -e`, or mutate one entry with `useradd` / `usermod` / `passwd`.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1734_test.go
+++ b/pkg/katas/katatests/zc1734_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #755, opened 2026-04-18
feat: ZC1735 — flag `efibootmgr -B` (deletes UEFI boot entry, may brick boot)

afadesigns:
ZC1735 — `efibootmgr -B`

What: Detect `efibootmgr -B` (delete boot entry) — short form. Combined with `-b BOOTNUM` removes the specific entry.
Why: If the deleted entry is the only viable bootloader (and removable-media fallback is missing), next reboot drops at the UEFI shell — recovery needs console access.
Fix suggestion: Run `efibootmgr -v` to inspect `BootOrder`, ensure a fallback exists, prefer `efibootmgr -o NEW,ORDER` to demote rather than delete.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1735_test.go
+++ b/pkg/katas/katatests/zc1735_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #757, opened 2026-04-18
feat: ZC1737 — flag `wpa_passphrase SSID PASSWORD` (Wi-Fi pass in argv)

afadesigns:
ZC1737 — `wpa_passphrase SSID PASSWORD`

What: Detect `wpa_passphrase SSID PASSPHRASE` (two positional args).
Why: Passphrase lands in argv — visible in `ps`, `/proc/<pid>/cmdline`, history, host audit logs.
Fix suggestion: Drop the PASSWORD argument and pipe via stdin (`wpa_passphrase SSID < /run/secrets/wifi`).
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1737_test.go
+++ b/pkg/katas/katatests/zc1737_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #758, opened 2026-04-18
feat: ZC1738 — flag `aws rds delete --skip-final-snapshot` (DB lost, no recovery)

afadesigns:
ZC1738 — `aws rds delete-db-instance --skip-final-snapshot`

What: Detect `aws rds delete-db-instance` or `aws rds delete-db-cluster` paired with `--skip-final-snapshot`.
Why: RDS keeps a final snapshot on delete — the only path back from a typo'd identifier or wrong account. The flag opts out of that snapshot; the database is gone the moment the API returns.
Fix suggestion: Drop `--skip-final-snapshot`, or pass `--final-db-snapshot-identifier <name>` so the snapshot is explicit and verifiable.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1738_test.go
+++ b/pkg/katas/katatests/zc1738_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #759, opened 2026-04-18
feat: ZC1739 — flag `git submodule update --remote` (unpins from parent index)

afadesigns:
ZC1739 — `git submodule update --remote`

What: Detect `git submodule update --remote`.
Why: Fetches each submodule's tracked branch HEAD instead of the commit pinned in the parent repo's index — non-reproducible builds, supply-chain risk.
Fix suggestion: Use `git submodule update --init --recursive` (defaults to pinned commit) and bump pins via reviewed PRs.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1739_test.go
+++ b/pkg/katas/katatests/zc1739_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #760, opened 2026-04-18
feat: ZC1740 — flag `gh release upload --clobber` (silent asset overwrite)

afadesigns:
ZC1740 — `gh release upload --clobber`

What: Detect `gh release upload TAG FILE --clobber`.
Why: Silently replaces an existing asset with the same name — a re-running CI job can downgrade the user-facing download with no audit trail.
Fix suggestion: Drop `--clobber` so the second upload errors out, or version the asset name (`mytool-1.2.3-linux.tar.gz`) so each upload has a unique slot.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1740_test.go
+++ b/pkg/katas/katatests/zc1740_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #761, opened 2026-04-18
feat: ZC1741 — flag `mkpasswd PASSWORD` (cleartext password in argv)

afadesigns:
ZC1741 — `mkpasswd PASSWORD`

What: Detect `mkpasswd` (Debian whois-package) with a positional password argument, including `mkpasswd -m METHOD PASSWORD`.
Why: Cleartext password lands in argv — visible in `ps`, `/proc/<pid>/cmdline`, history, host audit logs.
Fix suggestion: Use `mkpasswd -s` and pipe the password via stdin (`printf %s "$PASSWORD" | mkpasswd -s`).
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1741_test.go
+++ b/pkg/katas/katatests/zc1741_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #762, opened 2026-04-18
feat: ZC1743 — flag `npm audit fix --force` (silent major bumps)

afadesigns:
ZC1743 — `npm audit fix --force` / `pnpm audit --fix --force`

What: Detect `npm audit fix --force` and `pnpm audit --fix --force`.
Why: The flag accepts every semver-major bump an advisory triggers, so a build can silently move to a new major that removes APIs your code calls.
Fix suggestion: Drop `--force` and triage advisories one by one; pin or vendor targets that can't take a compatible patch.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1743_test.go
+++ b/pkg/katas/katatests/zc1743_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #763, opened 2026-04-18
feat: ZC1744 — flag `kubectl port-forward --address 0.0.0.0` (pod exposed LAN-wide)

afadesigns:
ZC1744 — `kubectl port-forward --address 0.0.0.0`

What: Detect `kubectl` (or `oc`) `port-forward` with `--address 0.0.0.0` / `::` / joined `=` form.
Why: Default loopback binding is replaced by every-interface binding — anyone on the LAN or VPN of the workstation / bastion can reach the pod's port.
Fix suggestion: Drop `--address` (loopback default) or pick a trusted-network interface IP.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1744_test.go
+++ b/pkg/katas/katatests/zc1744_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #765, opened 2026-04-18
feat: ZC1746 — flag `sysctl kernel.randomize_va_space=0|1` (ASLR weakened)

afadesigns:
ZC1746 — `sysctl kernel.randomize_va_space=0|1`

What: Detect `sysctl` setting `kernel.randomize_va_space` to `0` or `1` (with or without `-w`).
Why: `0` disables ASLR entirely; `1` omits the heap. Exploits relying on absolute addresses (stack overflows, ROP chains) become deterministic instead of brute-forceable.
Fix suggestion: Keep `kernel.randomize_va_space=2` (default) outside a sandboxed kernel-debug context.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1746_test.go
+++ b/pkg/katas/katatests/zc1746_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #766, opened 2026-04-19
feat: ZC1747 — flag `npm/yarn/pnpm --registry http://...` (plaintext index, MITM)

afadesigns:
ZC1747 — `npm/yarn/pnpm --registry http://...`

What: Detect `npm install --registry`, `pnpm --registry`, or `yarn config set registry` with an `http://` URL (joined `=` form or space-separated).
Why: Plaintext HTTP tarball fetches let any MITM swap contents; npm install-time `postinstall` scripts then execute the replaced code on the build host.
Fix suggestion: Switch to `https://`; internal networks should terminate TLS at the mirror.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1747_test.go
+++ b/pkg/katas/katatests/zc1747_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #767, opened 2026-04-19
feat: ZC1748 — flag `helm repo add NAME http://...` (plaintext chart repo)

afadesigns:
ZC1748 — `helm repo add NAME http://URL`

What: Detect `helm repo add NAME URL` where URL begins with `http://`.
Why: Plaintext chart-index fetches let any MITM swap `index.yaml` or chart tarballs. Subsequent `helm install` pulls container images and Kubernetes manifests from the substituted content.
Fix suggestion: Use `https://`, pair with `helm install --verify` or OCI chart signatures to pin provenance.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1748_test.go
+++ b/pkg/katas/katatests/zc1748_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #769, opened 2026-04-19
feat: ZC1750 — flag `kubectl proxy --address 0.0.0.0` (cluster API on every interface)

afadesigns:
ZC1750 — `kubectl proxy --address 0.0.0.0`

What: Detect `kubectl` (or `oc`) `proxy` with `--address 0.0.0.0` / `::` / joined `=` form.
Why: Default loopback binding is replaced by every-interface binding — anyone on the LAN or VPN gets the cluster admin credentials the kubeconfig holds.
Fix suggestion: Keep the loopback default and tunnel over SSH, or scope `--address` to a firewalled interface.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1750_test.go
+++ b/pkg/katas/katatests/zc1750_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #770, opened 2026-04-19
feat: ZC1751 — flag `rpm/dnf/yum remove --nodeps` (bypasses dependency check)

afadesigns:
ZC1751 — `rpm -e --nodeps` / `dnf remove --nodeps` / `zypper remove --force`

What: Detect remove-verb invocations of rpm / dnf / yum / microdnf / zypper paired with `--nodeps` / `--no-deps`.
Why: Skips the dependency solver — dependents break (libc, openssl, systemd units, even dnf itself), leaving the host unbootable or unpackageable.
Fix suggestion: Resolve the dependency conflict explicitly (`dnf swap`, pin the conflicting package) instead of bypassing.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1751_test.go
+++ b/pkg/katas/katatests/zc1751_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #771, opened 2026-04-19
feat: ZC1752 — flag `pvcreate/vgcreate/lvcreate -ff|--yes` (force-init LVM)

afadesigns:
ZC1752 — `pvcreate/vgcreate/lvcreate -ff|--yes|-y|--force`

What: Detect LVM create commands paired with force / auto-yes flags.
Why: LVM's prompt is the last defense against a typo'd device destroying filesystem / RAID / LVM signatures. Skipping it turns one wrong arg into data loss.
Fix suggestion: Inspect the target with `wipefs -n` + `lsblk -f`, drop the flag, re-run deliberately.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1752_test.go
+++ b/pkg/katas/katatests/zc1752_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #773, opened 2026-04-19
feat: ZC1754 — flag `gh auth status -t` / `--show-token` (token in stdout)

afadesigns:
ZC1754 — `gh auth status -t` / `--show-token`

What: Detect `gh auth status` paired with `-t` or `--show-token`.
Why: Prints the OAuth token in the status output — CI logs, scrollback, and `script`-captured sessions become a repo-admin leak.
Fix suggestion: Use `gh auth token` in automation so the secret-handling path is explicit, or drop the flag for human-readable status.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1754_test.go
+++ b/pkg/katas/katatests/zc1754_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #774, opened 2026-04-19
feat: ZC1755 — flag `gcloud sql users ... --password PASS` (Cloud SQL pass in argv)

afadesigns:
ZC1755 — `gcloud sql users create|set-password --password PASS`

What: Detect `gcloud sql users {create,set-password}` paired with `--password` / `--password=`.
Why: Password lands in argv — visible in `ps`, `/proc/<pid>/cmdline`, history, CI logs, and Cloud Audit Logs' request payload.
Fix suggestion: Use `--prompt-for-password` or call the SQL Admin API with a body file via `gcloud auth print-access-token | curl`.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1755_test.go
+++ b/pkg/katas/katatests/zc1755_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #777, opened 2026-04-19
feat: ZC1758 — flag `gh codespace delete --force` (drops uncommitted work)

afadesigns:
ZC1758 — `gh codespace delete --force`

What: Detect `gh codespace delete` paired with `--force` or `-f`.
Why: Skips the confirmation prompt and drops uncommitted / unpushed work along with the codespace. Combined with `--all`, wipes every codespace on the account in one line.
Fix suggestion: Drop the flag, let the prompt enumerate the target, and verify `git status` / `git stash list` inside the codespace first.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1758_test.go
+++ b/pkg/katas/katatests/zc1758_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #779, opened 2026-04-19
feat: ZC1760 — flag `openssl rand -hex|-base64 N` with N<16 (weak entropy)

afadesigns:
ZC1760 — `openssl rand -hex N` / `-base64 N` with N < 16

What: Detect `openssl rand -hex N` or `openssl rand -base64 N` where N < 16.
Why: Sub-128-bit output is brute-forceable offline with modest GPU resources — too weak for passwords, tokens, reset URLs, or any persistent secret.
Fix suggestion: Use `-hex 32` (256-bit) for secrets; `-hex 16` is only acceptable for short-validity nonces paired with rate-limited consumers.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1760_test.go
+++ b/pkg/katas/katatests/zc1760_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #780, opened 2026-04-19
feat: ZC1761 — flag `gh gist create --public` (file becomes world-indexed)

afadesigns:
ZC1761 — `gh gist create --public`

What: Detect `gh gist create` paired with `--public` or `-p`.
Why: Public gists are listed on `gist.github.com/discover`, crawled by search engines, and archived — a leaked secret or private snippet is effectively permanent the moment it lands.
Fix suggestion: Drop `--public` (default is secret-by-URL) unless public exposure is the explicit goal.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1761_test.go
+++ b/pkg/katas/katatests/zc1761_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #781, opened 2026-04-19
feat: ZC1762 — flag `kubeadm join --discovery-token-unsafe-skip-ca-verification`

afadesigns:
ZC1762 — `kubeadm join --discovery-token-unsafe-skip-ca-verification`

What: Detect `kubeadm join` paired with `--discovery-token-unsafe-skip-ca-verification`.
Why: Skips CA verification of the control-plane API server; an on-path attacker impersonates the API, harvests the bootstrap token, and seeds malicious workloads on the joining node.
Fix suggestion: Pin the CA with `--discovery-token-ca-cert-hash sha256:<digest>` or supply a kubeconfig discovery file with the CA pinned.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1762_test.go
+++ b/pkg/katas/katatests/zc1762_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #782, opened 2026-04-19
feat: ZC1763 — flag `docker compose down -v` / `--volumes` (wipes named volumes)

afadesigns:
ZC1763 — `docker compose down -v` / `--volumes`

What: Detect `docker compose down` (or `docker-compose down`) paired with `-v` or `--volumes`.
Why: Tears the stack down AND deletes every named volume — database contents, cache state, uploaded assets. No soft-delete.
Fix suggestion: Drop the flag in CI / prod scripts; keep only for throwaway local testbeds where volume loss is intentional.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1763_test.go
+++ b/pkg/katas/katatests/zc1763_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #784, opened 2026-04-19
feat: ZC1765 — flag `snap remove --purge` (skips pre-remove data snapshot)

afadesigns:
ZC1765 — `snap remove --purge SNAP`

What: Detect `snap remove` paired with `--purge`.
Why: Default `snap remove` snapshots the snap's writable areas so data can be restored via `snap restore`. `--purge` skips that snapshot — snap and data gone, no rollback.
Fix suggestion: Drop `--purge` or capture a `snap save` set ID first.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1765_test.go
+++ b/pkg/katas/katatests/zc1765_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #785, opened 2026-04-19
feat: ZC1766 — flag `memcached -l 0.0.0.0` (unauthenticated cache exposed)

afadesigns:
ZC1766 — `memcached -l 0.0.0.0`

What: Detect `memcached` binding its listener to an unrestricted address (`0.0.0.0` / `::` / joined `-l0.0.0.0`).
Why: Memcached has no auth; the UDP handler used to be the largest DDoS-amplification vector on the internet. Binding to 0.0.0.0 exposes it to every reachable interface.
Fix suggestion: Bind to `127.0.0.1` or a private-network IP and firewall the port.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1766_test.go
+++ b/pkg/katas/katatests/zc1766_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #786, opened 2026-04-19
feat: ZC1767 — flag `mongod --bind_ip 0.0.0.0` (Mongo exposed on every interface)

afadesigns:
ZC1767 — `mongod --bind_ip 0.0.0.0`

What: Detect `mongod` binding to `0.0.0.0` / `::` (uses parser leading-flag mangling — SimpleCommand name becomes `bind_ip`).
Why: Combined with no-auth defaults or a wildcard user, this is the 2017 MongoDB-ransomware-wave recipe — tens of thousands of public instances wiped.
Fix suggestion: Bind to `127.0.0.1` or a private-network IP, enable `--auth`, firewall port 27017.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1767_test.go
+++ b/pkg/katas/katatests/zc1767_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #787, opened 2026-04-19
feat: ZC1768 — flag `sqlcmd/bcp/osql -P PASSWORD` (SQL Server pass in argv)

afadesigns:
ZC1768 — `sqlcmd/bcp/osql -P PASSWORD`

What: Detect Microsoft SQL Server CLI tools (`sqlcmd`, `bcp`, `osql`) paired with `-P` followed by a password value.
Why: Password lands in argv — visible in `ps`, `/proc/<pid>/cmdline`, history, CI logs, and SQL Server's audit trace.
Fix suggestion: Use `-P` with no value (prompts), or set `SQLCMDPASSWORD` env var from a secrets file; `-G` + Azure AD avoids the password entirely.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1768_test.go
+++ b/pkg/katas/katatests/zc1768_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #788, opened 2026-04-19
feat: ZC1769 — flag `vagrant destroy --force` (VM wiped without prompt)

afadesigns:
ZC1769 — `vagrant destroy --force`

What: Detect `vagrant destroy` paired with `--force` or `-f`.
Why: Skips the confirmation and tears down every VM in the Vagrantfile — any un-exported data inside (seeds, caches, fixtures) goes with the VM.
Fix suggestion: Drop the flag so the prompt gates the destroy; cycle CI boxes with `vagrant halt` + `vagrant up` instead of destroy.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1769_test.go
+++ b/pkg/katas/katatests/zc1769_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #789, opened 2026-04-19
feat: ZC1770 — warn on gpg --always-trust / --trust-model always

afadesigns:
ZC1770 — gpg --always-trust / --trust-model always

What: detect gpg / gpg2 called with --always-trust, --trust-model always, or --trust-model=always.
Why: these options mark every imported key as fully trusted; a signature from an attacker-supplied key verifies cleanly.
Fix suggestion: drop the flag and either pin the expected fingerprint and match after --status-fd 1, or set trust once via gpg --edit-key KEYID trust.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1770_test.go
+++ b/pkg/katas/katatests/zc1770_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #790, opened 2026-04-19
feat: ZC1771 — warn on alias -g / alias -s in scripts

afadesigns:
ZC1771 — alias -g / alias -s in scripts

What: detect `alias -g NAME=value` (global) or `alias -s ext=cmd` (suffix) definitions.
Why: both forms expand outside command position, producing surprising substitutions for anyone reading the script. Works as designed in interactive rc files but confuses script readers.
Fix suggestion: prefer a regular function; keep global/suffix aliases in `~/.zshrc` where the definition is discoverable.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1771_test.go
+++ b/pkg/katas/katatests/zc1771_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #791, opened 2026-04-19
feat: ZC1772 — error on hdparm --security-erase / --trim-sector-ranges

afadesigns:
ZC1772 — hdparm destructive ATA commands

What: detect hdparm called with --security-erase, --security-erase-enhanced, --security-set-pass, --security-unlock, --security-disable, --security-freeze, or --trim-sector-ranges.
Why: these flags talk to the drive firmware directly — filesystem and partition boundaries are ignored, the operation cannot be rolled back, and credential flags can lock the device out of future access.
Fix suggestion: keep the call behind a runbook, pin the disk via /dev/disk/by-id/…, and keep the password out of argv.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1772_test.go
+++ b/pkg/katas/katatests/zc1772_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #792, opened 2026-04-19
feat: ZC1773 — warn on xargs without -r / --no-run-if-empty

afadesigns:
ZC1773 — xargs without -r / --no-run-if-empty

What: detect xargs invocations that omit -r / --no-run-if-empty.
Why: GNU xargs runs the child command once with no arguments when stdin is empty. Paired with a destructive child (xargs rm, xargs kill, xargs docker stop) an empty producer silently issues a destructive call with no operand.
Fix suggestion: add -r (GNU) / --no-run-if-empty, or switch to find ... -exec cmd {} + which never runs the child on empty input.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1773_test.go
+++ b/pkg/katas/katatests/zc1773_test.go
@@ -1,0 +1,68 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #793, opened 2026-04-19
feat: ZC1774 — warn on setopt GLOB_SUBST / set -o globsubst

afadesigns:
ZC1774 — setopt GLOB_SUBST

What: detect setopt GLOB_SUBST, set -G, and set -o GLOB_SUBST (also case/underscore folded).
Why: with GLOB_SUBST on, every unquoted $var expansion is rescanned as a filename glob. User-controlled data becomes a filesystem query — an injection vector and a subtle source of no-matches-found failures on empty variables.
Fix suggestion: scope the option inside a narrow subshell / function, or use explicit expansion flags (~, (e), (P)) where you actually want the rescan.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1774_test.go
+++ b/pkg/katas/katatests/zc1774_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #794, opened 2026-04-19
feat: ZC1775 — warn on timeout without --kill-after / -k

afadesigns:
ZC1775 — timeout without --kill-after / -k

What: detect timeout DURATION cmd when no --kill-after or -k flag is present.
Why: timeout only sends SIGTERM after the duration. A child that blocks or ignores SIGTERM (long-running daemon, stuck in D state, trapped handler) keeps running, so the pipeline hangs past the intended bound.
Fix suggestion: add --kill-after=N so timeout escalates to SIGKILL after N seconds, guaranteeing exit within DURATION + N.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1775_test.go
+++ b/pkg/katas/katatests/zc1775_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #795, opened 2026-04-19
feat: ZC1776 — error on DB connection URI with password in argv

afadesigns:
ZC1776 — DB / broker URI with password in argv

What: detect scheme://user:password@host/... URIs on the command line (postgres, postgresql, mysql, mariadb, mongodb, mongodb+srv, redis, rediss, amqp, amqps, kafka, nats, clickhouse, cockroachdb, db2, mssql, oracle).
Why: the full URI shows up in ps, /proc/PID/cmdline, process accounting, audit logs, and shell history — everyone on the host who can list processes sees the password.
Fix suggestion: use password files (~/.pgpass, ~/.my.cnf), env vars (PGPASSWORD, REDISCLI_AUTH, etc.), or build the URI from a secret variable so the secret is not in argv.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1776_test.go
+++ b/pkg/katas/katatests/zc1776_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #796, opened 2026-04-19
feat: ZC1777 — error on tee/cp/mv/install/dd writing /etc/ld.so.preload

afadesigns:
ZC1777 — rootkit-persistence via /etc/ld.so.preload

What: detect writes to /etc/ld.so.preload via tee, cp, mv, install, dd, or shell redirection (>, >>).
Why: the linker force-loads every library listed in /etc/ld.so.preload into every dynamically-linked process — a textbook rootkit persistence primitive (libprocesshider, Azazel, Jynx). Package managers do not touch this file; almost every legitimate use case is solved by scoped LD_PRELOAD= on a single invocation.
Fix suggestion: remove the write, audit the file for unexpected entries (sha256sum, diff against backup). If preload is legitimately required, scope it to a single command.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1777_test.go
+++ b/pkg/katas/katatests/zc1777_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #797, opened 2026-04-19
feat: ZC1778 — warn on systemctl link from mutable source path

afadesigns:
ZC1778 — systemctl link from a mutable path

What: detect systemctl link <path> where the target path lives under /tmp, /var/tmp, /home, /root, /opt, /srv, /mnt, /media, /var/lib.
Why: systemctl link symlinks the unit from its original location. The definition remains at the source path, so a later tamper of that file silently changes what systemd runs on the next start — a persistence primitive the unit owner did not consent to.
Fix suggestion: copy the unit into /etc/systemd/system/ with root-only permissions, or install a package that ships it under /lib/systemd/system/.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1778_test.go
+++ b/pkg/katas/katatests/zc1778_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #799, opened 2026-04-19
feat: ZC1780 — warn on sysctl disabling fs.protected_* TOCTOU guards

afadesigns:
ZC1780 — fs.protected_* TOCTOU guards disabled

What: detect sysctl setting fs.protected_symlinks=0, fs.protected_hardlinks=0, fs.protected_fifos=0, fs.protected_regular=0.
Why: these four kernel guards block the classic TOCTOU race where a non-owner plants a symlink/hardlink/FIFO/regular file in a sticky directory (/tmp, /var/tmp, /dev/shm) so a privileged program that opens a predictable name is tricked into reading or writing a sensitive target. Disabling them re-opens the race host-wide.
Fix suggestion: leave the defaults (1 or 2) in place; if a specific application legitimately needs the old behavior, scope it to a mount namespace where /tmp is not sticky-shared.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1780_test.go
+++ b/pkg/katas/katatests/zc1780_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #800, opened 2026-04-19
feat: ZC1781 — error on git clone/fetch with credentials in URL

afadesigns:
ZC1781 — git URL with user:token@host

What: detect git clone / fetch / pull / push / ls-remote / archive invoked with https://user:secret@host/... (or http/git+https variants).
Why: the credentials show up in argv (ps, /proc/PID/cmdline, shell history, process accounting) and, for clone, in .git/config as the origin remote — every later fetch/pull re-exposes the same token to anyone who can read that file.
Fix suggestion: use a credential helper (git credential-store, credential-osxkeychain), GIT_ASKPASS reading from an env-sourced secret, HTTPS + an SSH deploy key, or set http.extraHeader with an Authorization: Bearer value from env.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1781_test.go
+++ b/pkg/katas/katatests/zc1781_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #801, opened 2026-04-19
feat: ZC1782 — error on flatpak remote-add --no-gpg-verify

afadesigns:
ZC1782 — flatpak remote without GPG verification

What: detect flatpak remote-add / remote-modify with --no-gpg-verify, --gpg-verify=false, or --no-gpg-verify=true.
Why: signatures tie flatpak install FOO to the operator that actually built FOO. Without them the install reduces to an HTTPS download with no identity, and any update that the remote server (or anyone on the path) sends is accepted.
Fix suggestion: sign the repo with ostree gpg-sign and import the key via --gpg-import=KEYFILE.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1782_test.go
+++ b/pkg/katas/katatests/zc1782_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #802, opened 2026-04-19
feat: ZC1783 — error on podman system reset / nerdctl prune -a --volumes

afadesigns:
ZC1783 — factory-reset the local container engine

What: detect podman system reset (any flags) and nerdctl system prune -a --volumes.
Why: both wipe every local container artifact — images, volumes, networks, pods — not just the ones related to the current project. On a dev workstation this drops caches for unrelated work; on a CI runner it invalidates every warm artifact the job depends on; on a prod host it loses the volumes the workload stores data in.
Fix suggestion: use narrower removals (podman rmi, podman volume rm, scoped prune) targeting the specific resource you intend to delete.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1783_test.go
+++ b/pkg/katas/katatests/zc1783_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #803, opened 2026-04-19
feat: ZC1784 — warn on git config core.hooksPath to mutable path

afadesigns:
ZC1784 — git config core.hooksPath from a mutable path

What: detect git config core.hooksPath <path> where path lives under /tmp, /var/tmp, /dev/shm, /home, /root, /opt, /srv, /mnt, /media.
Why: core.hooksPath turns any file named pre-commit, post-checkout, etc. under the directory into executable code run on routine git ops. Pointing it at a non-repo-owned path (notably /tmp or another user's home) hands the git CLI an execution primitive that a different local user can overwrite at will.
Fix suggestion: keep hooks inside the repo's .git/hooks/ (or a tracked .githooks/) and point core.hooksPath at repo-owned paths only.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1784_test.go
+++ b/pkg/katas/katatests/zc1784_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #804, opened 2026-04-19
feat: ZC1785 — error on ufw default allow flipping firewall baseline

afadesigns:
ZC1785 — ufw default allow

What: detect ufw default allow [incoming|outgoing|routed].
Why: flips the chain baseline from deny-by-default to allow-by-default — every port without an explicit deny is accepted. On an internet-facing host this is effectively \"firewall off\", and the change persists to /etc/default/ufw so it survives reboots.
Fix suggestion: restore ufw default deny incoming and add narrow ufw allow <port> rules only for services that must be reachable.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1785_test.go
+++ b/pkg/katas/katatests/zc1785_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #805, opened 2026-04-19
feat: ZC1786 — error on mount.cifs / mount -t cifs with password= in argv

afadesigns:
ZC1786 — SMB password in mount options

What: detect mount.cifs or mount -t cifs/smb3 called with -o ...password=SECRET or ...pass=SECRET.
Why: the password shows up in argv — readable via ps, /proc/PID/cmdline, process accounting, shell history, and any captured CI log.
Fix suggestion: use -o credentials=/path/to/cifs-creds (mode 0600 with username= and password= lines), let mount.cifs read $USER/$PASSWD from env, or use pam_mount for login-time mounts.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1786_test.go
+++ b/pkg/katas/katatests/zc1786_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #806, opened 2026-04-19
feat: ZC1787 — warn on setopt AUTO_CD / set -o autocd in scripts

afadesigns:
ZC1787 — AUTO_CD in a script silently cd's

What: detect setopt AUTO_CD, setopt autocd, set -o autocd (case/underscore folded).
Why: with AUTO_CD on, any bare word that happens to name an existing directory becomes cd <word>. A typo in a command, or a user-controlled variable that expands to a path, silently reshapes $PWD for every later relative path.
Fix suggestion: keep AUTO_CD in ~/.zshrc where it belongs, not in .zsh scripts. Never toggle it inside a function that external callers depend on.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1787_test.go
+++ b/pkg/katas/katatests/zc1787_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #807, opened 2026-04-19
feat: ZC1788 — warn on ssh/scp/sftp -F config from mutable path

afadesigns:
ZC1788 — ssh -F config from a mutable path

What: detect ssh -F / scp -F / sftp -F pointing at a path under /tmp, /var/tmp, /dev/shm, /home, /root, /opt, /srv, /mnt, /media.
Why: ssh -F loads a user-supplied config file. Anything in ssh_config can be overridden — notably ProxyCommand, LocalCommand, PermitLocalCommand, Include — so a mutable source file is an execution primitive for the ssh CLI. A local attacker flips ProxyCommand to /tmp/pwn and the next ssh run launches it with the caller's credentials and forwarded keys.
Fix suggestion: keep the config in ~/.ssh/config (or a repo-owned path with 0600 perms).
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1788_test.go
+++ b/pkg/katas/katatests/zc1788_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #808, opened 2026-04-19
feat: ZC1789 — warn on setopt CORRECT / CORRECT_ALL in scripts

afadesigns:
ZC1789 — Zsh spellcheck in a script silently rewrites tokens

What: detect setopt CORRECT, setopt CORRECT_ALL, set -o correct, set -o correctall (case/underscore folded).
Why: CORRECT prompts to rewrite mistyped command names; CORRECT_ALL extends that to every argument on the line. In a script a filename that is close enough to an existing file gets silently replaced, and the correction prompt reads from stdin — which may be the data the script was supposed to process.
Fix suggestion: keep CORRECT / CORRECT_ALL in ~/.zshrc only; never toggle inside a function a script calls.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1789_test.go
+++ b/pkg/katas/katatests/zc1789_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #809, opened 2026-04-19
feat: ZC1790 — warn on unsetopt PIPE_FAIL / setopt NOPIPEFAIL

afadesigns:
ZC1790 — globally disabling pipefail

What: detect unsetopt PIPE_FAIL and setopt NOPIPEFAIL (case/underscore folded).
Why: with PIPE_FAIL off, cmd1 | cmd2 | cmd3 exits with cmd3's status and failures in cmd1 / cmd2 are silently dropped. Disabling it mid-script typically happens because a flaky pipe stage was tripping set -e, and the global off-switch is the wrong fix.
Fix suggestion: scope the change to a subshell ( ( unsetopt pipefail; …; ) ) or a function with emulate -L zsh so the rest of the script keeps strict-pipe error propagation.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1790_test.go
+++ b/pkg/katas/katatests/zc1790_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #810, opened 2026-04-19
feat: ZC1791 — error on curl --unix-socket to docker/containerd/crio/podman daemon

afadesigns:
ZC1791 — curl talking directly to the container daemon

What: detect curl --unix-socket PATH where PATH is a known container-daemon socket (/var/run/docker.sock, /run/docker.sock, /run/containerd/containerd.sock, /run/crio/crio.sock, /run/podman/podman.sock and variants).
Why: a raw HTTP call to the daemon has no authentication beyond socket filesystem permissions. A POST to /containers/create with HostConfig.Privileged=true and a bind mount of / lands a root shell on the host — the primitive every docker-socket escape leans on.
Fix suggestion: use the real CLI (docker, podman, nerdctl) which enforces its own policy, or access the daemon over a TLS-protected TCP endpoint with mutual auth.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1791_test.go
+++ b/pkg/katas/katatests/zc1791_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #811, opened 2026-04-19
feat: ZC1792 — warn on btrfs subvolume delete / device remove

afadesigns:
ZC1792 — btrfs destructive ops without automatic rollback

What: detect btrfs subvolume delete PATH and btrfs device remove DEV POOL.
Why: subvolume delete drops all extents once cleanup completes — on Snapper / Timeshift systems that PATH is often the only remaining copy of pre-incident state. device remove tries to move chunks off DEV first; wrong device, mid-rebalance failure, or insufficient free space puts the filesystem into degraded mode with no automatic rollback.
Fix suggestion: keep fresh btrfs subvolume list / btrfs device usage snapshots and confirm the target explicitly before running either command in automation.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1792_test.go
+++ b/pkg/katas/katatests/zc1792_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #813, opened 2026-04-19
feat: ZC1794 — error on cosign --insecure-ignore-tlog / --allow-insecure-registry

afadesigns:
ZC1794 — cosign signature chain disabled

What: detect cosign with --insecure-ignore-tlog, --insecure-ignore-sct, --insecure-skip-verify, --allow-insecure-registry, --allow-http-registry, or --allow-insecure-bundle.
Why: each flag removes a distinct rung of the signature chain cosign was built to enforce. --insecure-ignore-tlog skips Rekor transparency-log verification, --insecure-ignore-sct skips Fulcio SCT checks, --insecure-skip-verify turns off TLS cert validation, and the --allow-*-registry flags push or pull signatures over plain HTTP. A malicious registry or on-path attacker passes verification without detection.
Fix suggestion: drop the flag, fix the underlying trust anchor (CA bundle, Rekor URL, Fulcio OIDC), and keep verification strict.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1794_test.go
+++ b/pkg/katas/katatests/zc1794_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #814, opened 2026-04-19
feat: ZC1795 — error on git remote add/set-url persisting credentials in .git/config

afadesigns:
ZC1795 — git remote URL with credentials

What: detect git remote add NAME URL and git remote set-url NAME URL where URL embeds https://user:token@host/path (or http / git+https / git+http variants).
Why: the URL is persisted verbatim to .git/config. Every reader of the repo — other local users, a backup, a CI cache, or anyone who runs git config --list — sees the secret. The command also leaks it via argv at creation time (ps, /proc/PID/cmdline).
Fix suggestion: use a credential helper (git credential-store, credential-osxkeychain), GIT_ASKPASS reading from an env-sourced secret, or HTTPS + an SSH deploy key.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1795_test.go
+++ b/pkg/katas/katatests/zc1795_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #815, opened 2026-04-19
feat: ZC1796 — warn on pg_restore -c / --clean dropping objects before restore

afadesigns:
ZC1796 — pg_restore --clean drops existing DB objects

What: detect pg_restore -c and pg_restore --clean (including the parser-mangled form where --clean becomes the SimpleCommand name and a pg_restore-shaped flag such as -d / --dbname / -U is present).
Why: -c issues DROP for every table, index, function, sequence in the target DB before recreating them from the archive. A stale / incomplete / wrong-target dump silently loses any object that isn't in the backup, including data added after the dump ran.
Fix suggestion: restore into a fresh empty database (createdb new && pg_restore -d new) or snapshot the target (pg_dump -Fc > pre.dump) before running --clean on a live DB; never pair with --if-exists in production without a tested rollback.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1796_test.go
+++ b/pkg/katas/katatests/zc1796_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #816, opened 2026-04-19
feat: ZC1797 — warn on ip link set <iface> down / ifdown locking out remote admin

afadesigns:
ZC1797 — disabling a network interface from a remote session

What: detect ip link set IFACE down and ifdown IFACE.
Why: running either from an SSH session that rides on the same interface cuts the script off mid-run — the TCP connection freezes, later steps silently fail, and recovery requires console / out-of-band access. Common bugs: typoed interface name, scripts that target the only uplink on a cloud VM, or running the command without confirming which interface carries the admin session.
Fix suggestion: wrap the down in a systemd-run --on-active=30s --unit=recover ip link set IFACE up rollback, or stage down and up through nmcli connection up/down with a pinned fallback profile.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1797_test.go
+++ b/pkg/katas/katatests/zc1797_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #817, opened 2026-04-19
feat: ZC1798 — warn on ufw reset wiping every firewall rule

afadesigns:
ZC1798 — ufw reset drops every rule

What: detect ufw reset (and the parser-mangled ufw --force reset form).
Why: ufw reset returns the firewall to distro default — every user-defined rule is removed. Mid-session on a production host, the change drops every allow-list entry in one step and breaks whatever traffic relied on them.
Fix suggestion: snapshot first (ufw status numbered > /tmp/ufw.bak), and prefer ufw delete <num> for targeted removals over a full reset.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1798_test.go
+++ b/pkg/katas/katatests/zc1798_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #818, opened 2026-04-19
feat: ZC1799 — warn on rclone sync without --dry-run

afadesigns:
ZC1799 — rclone sync without a dry-run guard

What: detect rclone sync SRC DST with none of --dry-run / -n / --interactive / -i.
Why: rclone sync is a one-way mirror — anything in DST that is not in SRC gets deleted, including object versions on providers that keep them. An accidentally empty SRC (wrong path, unmounted drive, wrong credentials pointing at an empty bucket) silently wipes every object under DST.
Fix suggestion: preview with rclone sync --dry-run first, and pin guards like --backup-dir, --max-delete, or --min-age so a bad SRC cannot cascade into unbounded deletion.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1799_test.go
+++ b/pkg/katas/katatests/zc1799_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #819, opened 2026-04-19
feat: ZC1800 — warn on pg_ctl stop -m immediate skipping shutdown checkpoint

afadesigns:
ZC1800 — pg_ctl immediate shutdown

What: detect pg_ctl stop or pg_ctl restart with -m immediate (short or --mode=immediate).
Why: immediate shutdown sends SIGQUIT to the postmaster. Server processes drop connections, no checkpoint is taken, and buffered changes stay in memory. Recovery on the next start replays WAL; if WAL is corrupt or on unhealthy storage, committed transactions can be lost.
Fix suggestion: use -m smart (default) or -m fast so the server issues a shutdown checkpoint and closes cleanly. Reserve -m immediate for emergencies paired with a tested PITR procedure.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1800_test.go
+++ b/pkg/katas/katatests/zc1800_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #820, opened 2026-04-19
feat: ZC1801 — warn on fwupdmgr update / install flashing firmware

afadesigns:
ZC1801 — firmware flash can brick the device on interruption

What: detect fwupdmgr update, upgrade, install, reinstall, downgrade.
Why: fwupdmgr pushes new firmware into BIOS / UEFI, SSDs, Thunderbolt controllers, NICs, dock microcontrollers. Most of those chips have no A/B rollback — an interrupted flash (power cut, unexpected reboot, PSU toggle) leaves the device unbootable and requires vendor recovery hardware.
Fix suggestion: run from a battery-backed session, mask reboot triggers with systemd-inhibit, pin the power supply, and verify the update history with fwupdmgr get-history once the device returns.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1801_test.go
+++ b/pkg/katas/katatests/zc1801_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #821, opened 2026-04-19
feat: ZC1802 — warn on dnf/yum history undo|rollback reverting transaction

afadesigns:
ZC1802 — reverting a package transaction without compat check

What: detect dnf / dnf5 / yum history undo N and history rollback N.
Why: undo reverts the exact package set of transaction N (installs → removes, removes → installs, updates → downgrades). rollback does the same for every transaction since N. Neither rechecks that older versions still resolve against the current package graph — dependencies can downgrade, security patches get reverted, and services whose configuration was migrated can fail to start on the older binary.
Fix suggestion: review with dnf history info N, pin scope with --exclude=, or restore a filesystem snapshot taken before the original transaction.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1802_test.go
+++ b/pkg/katas/katatests/zc1802_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #822, opened 2026-04-19
feat: ZC1803 — error on mysql --skip-ssl / psql sslmode=disable

afadesigns:
ZC1803 — database client with TLS disabled

What: detect MySQL / MariaDB clients (mysql, mysqldump, mysqladmin, mariadb, mariadb-dump, mariadb-admin) called with --skip-ssl, --ssl=0, --ssl=false, --ssl-mode=DISABLED/disable; and Postgres clients (psql, pg_dump, pg_restore, pgbench) with sslmode=disable in any argument (connection URI or keyword form).
Why: disabling TLS sends the login handshake (including the auth challenge or password) and every query / result over plaintext TCP. Anyone in the network path — cloud VPC, office LAN, compromised router — can sniff or modify the stream.
Fix suggestion: use --ssl-mode=VERIFY_IDENTITY (MySQL 8+) or sslmode=verify-full (psql) with a pinned CA bundle.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1803_test.go
+++ b/pkg/katas/katatests/zc1803_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #823, opened 2026-04-19
feat: ZC1804 — warn on aws ec2 destructive actions without --dry-run

afadesigns:
ZC1804 — AWS EC2 destructive actions

What: detect aws ec2 terminate-instances, delete-volume, delete-snapshot, delete-vpc, delete-internet-gateway, delete-network-interface, delete-security-group, delete-launch-template when the call does not include --dry-run.
Why: these drop cloud state with no automatic backup — instance-store volumes vanish on terminate, EBS volumes / snapshots cannot be restored AWS-side once deleted, and a mis-typed VPC / ENI / SG ID can take down unrelated workloads in the same account.
Fix suggestion: review the target list with aws ec2 describe-…, pair destructive commands with --dry-run, and pin IDs through --cli-input-json instead of passing them inline.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1804_test.go
+++ b/pkg/katas/katatests/zc1804_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #824, opened 2026-04-19
feat: ZC1805 — warn on destructive aws cloudformation/dynamodb/logs/kms/lambda/ecr actions

afadesigns:
ZC1805 — non-EC2 AWS destructive actions

What: detect aws cloudformation delete-stack / delete-stack-set, dynamodb delete-table / delete-backup, logs delete-log-group / delete-log-stream, kms schedule-key-deletion, lambda delete-function / delete-event-source-mapping, ecr delete-repository / batch-delete-image — when the call does not include --dry-run.
Why: each drops state AWS cannot restore: CloudFormation tears down every managed resource with no rollback, dynamodb delete-table loses items, logs delete-log-group erases the audit trail, and kms schedule-key-deletion makes every ciphertext unreadable after the grace window.
Fix suggestion: stage a typed confirmation, pin IDs via --cli-input-json, and export a backup first where the service supports one (dynamodb export-table-to-point-in-time, logs create-export-task).
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1805_test.go
+++ b/pkg/katas/katatests/zc1805_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1805(t *testing.T) {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #825, opened 2026-04-19
feat: ZC1806 — warn on zmv without -n dry-run or -i interactive guard

afadesigns:
ZC1806 — zmv silent bulk rename

What: detect zmv invocations (2+ args) with no -n / --dry-run or -i / --interactive, including combined short-flag forms (-Wn, -Mi, etc.).
Why: zmv rewrites every filename matching the pattern in one shot. A small typo — *.jpg vs *.JPG, a misplaced (..), forgetting ** recursion — collides names and can silently overwrite files. zmv's own conflict detector catches same-target collisions but not semantic errors.
Fix suggestion: preview with zmv -n 'PAT' 'REP' to see the rename list, or zmv -i to prompt per file. Drop the guard only after the preview matches expectations.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1806_test.go
+++ b/pkg/katas/katatests/zc1806_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #826, opened 2026-04-19
feat: ZC1807 — warn on gh api -X DELETE bypassing gh subcommand confirmations

afadesigns:
ZC1807 — raw GitHub DELETE via gh api

What: detect gh api -X DELETE <path>, gh api --method DELETE, and gh api --method=DELETE forms.
Why: raw API DELETE uses the caller's token and has no --yes guard, no dry-run, and no friendly prompt. A script that builds the path from a variable can wipe repos, releases, deploy keys, workflow runs, issue comments, or whole organisations in one call.
Fix suggestion: use the high-level gh subcommand for the target (gh repo delete, gh release delete, gh workflow disable) which at least requires --yes, or wrap the raw call with a preflight gh api -X GET + explicit confirmation.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1807_test.go
+++ b/pkg/katas/katatests/zc1807_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #827, opened 2026-04-19
feat: ZC1808 — warn on kubectl replace --force deleting and recreating resource

afadesigns:
ZC1808 — kubectl replace --force

What: detect kubectl replace --force (any args).
Why: --force makes replace perform delete + create instead of in-place update. Pods die, PodDisruptionBudgets are ignored, controllers watching the object see it disappear and reappear, and in-flight requests drop.
Fix suggestion: use kubectl apply -f FILE (server-side merge that preserves running pods) and reserve replace --force for schema changes apply cannot patch — with traffic drained first.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1808_test.go
+++ b/pkg/katas/katatests/zc1808_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #828, opened 2026-04-19
feat: ZC1809 — error on gsutil rm -r / rb -f bulk GCS deletion

afadesigns:
ZC1809 — bulk GCS deletion

What: detect gsutil rm with -r / -R / --recursive / -rf and gsutil rb with -f / --force.
Why: gsutil rm -r gs://bucket/prefix removes every object under the prefix (with -m in parallel, faster than any undo window). gsutil rb -f removes the bucket after force-deleting contents. Neither soft-deletes, Object Versioning helps only if enabled beforehand, and rb leaves no retention grace.
Fix suggestion: preview with gsutil ls, enable Object Versioning / retention locks ahead of time, and prefer narrower object-level gsutil rm calls.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1809_test.go
+++ b/pkg/katas/katatests/zc1809_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #829, opened 2026-04-19
feat: ZC1810 — warn on wget -r / --mirror without --level=N bound

afadesigns:
ZC1810 — unbounded recursive wget

What: detect wget called with -r / --recursive / -m / --mirror when no --level=N or -l N depth bound is present. Also catches parser-mangled forms where --mirror or --recursive becomes the SimpleCommand name.
Why: wget recursion follows links to arbitrary depth. Without --level the crawl keeps going until wget hits remote limits, fills local disk, or climbs into parent directories that the author did not intend to mirror.
Fix suggestion: pin a depth (--level=3), add --no-parent, and cap the byte budget (--quota=1G) before running a recursive wget in automation.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1810_test.go
+++ b/pkg/katas/katatests/zc1810_test.go
@@ -1,0 +1,63 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #830, opened 2026-04-19
feat: ZC1811 — error on chown/chmod/chgrp --no-preserve-root

afadesigns:
ZC1811 — --no-preserve-root on chown/chmod/chgrp

What: detect chown, chmod, chgrp invocations with --no-preserve-root (also the parser-mangled no-preserve-root-as-name form).
Why: GNU coreutils refuses to recurse into / by default (the --preserve-root safeguard). --no-preserve-root opts in to walking the entire filesystem, so a stray variable expansion combined with -R rewrites ownership or mode on every file on the host.
Fix suggestion: remove the flag; list explicit paths. ZC1625 already covers rm --no-preserve-root; this kata covers the ownership/mode siblings.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1811_test.go
+++ b/pkg/katas/katatests/zc1811_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #831, opened 2026-04-19
feat: ZC1812 — error on aws ssm put-parameter SecureString with plaintext --value

afadesigns:
ZC1812 — SecureString plaintext in argv

What: detect aws ssm put-parameter --type SecureString --value SECRET (space or =-form), where SECRET is an inline argument that is neither empty nor file://…
Why: SecureString is the SSM way to mark a parameter sensitive. Passing its plaintext inline puts the secret in argv — readable via ps, /proc/PID/cmdline, shell history, and AWS CLI --debug logs.
Fix suggestion: use --cli-input-json file://param.json (mode 0600) or --value file://secret; for richer secret storage use aws secretsmanager create-secret --secret-string file://secret.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1812_test.go
+++ b/pkg/katas/katatests/zc1812_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #832, opened 2026-04-19
feat: ZC1813 — warn on cryptsetup luksFormat / reencrypt destructive LUKS write

afadesigns:
ZC1813 — destructive LUKS header / reencrypt

What: detect cryptsetup with luksFormat, luks-format, or reencrypt as a subcommand arg.
Why: luksFormat writes a fresh LUKS2 header and treats remaining space as new ciphertext — any pre-existing filesystem or LUKS metadata is gone. reencrypt rewrites the whole device in place, and an interruption mid-write leaves the volume partially re-encrypted and dependent on --resume-only recovery.
Fix suggestion: verify DEV with lsblk -o NAME,MODEL,SERIAL, back up with cryptsetup luksHeaderBackup, and run on an unmounted volume with UPS-backed power.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1813_test.go
+++ b/pkg/katas/katatests/zc1813_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #833, opened 2026-04-19
feat: ZC1814 — error on dpkg --force-all enabling every force option

afadesigns:
ZC1814 — dpkg --force-all

What: detect dpkg --force-all, apt/apt-get -o Dpkg::Options::=--force-all, and the parser-mangled force-all-as-name form.
Why: --force-all enables ~18 distinct --force-<option> flags at once: overwrite, unsigned, downgrade, remove-essential, depends-broken, and more. The dpkg manual says it is \"almost always a bad idea\". In provisioning scripts it hides the specific constraint the author was trying to bypass; the same state re-surfaces on the next unattended upgrade.
Fix suggestion: drop --force-all and spell out only the specific --force-<option> you need, or fix the upstream dependency conflict.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1814_test.go
+++ b/pkg/katas/katatests/zc1814_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #834, opened 2026-04-19
feat: ZC1815 — warn on systemctl restart NetworkManager / systemd-networkd dropping SSH session

afadesigns:
ZC1815 — restarting the network manager over SSH

What: detect systemctl restart / stop / reload-or-restart of NetworkManager, systemd-networkd, networking, network, wpa_supplicant (including .service suffixes).
Why: these units supervise every active connection the host holds open. Restarting them from an SSH session tears down the one the script runs over — the script freezes and recovery usually needs console access.
Fix suggestion: use nmcli connection reload + nmcli connection up <name> for NetworkManager, networkctl reload for systemd-networkd, or schedule the restart behind systemd-run --on-active=30s with a rollback timer.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1815_test.go
+++ b/pkg/katas/katatests/zc1815_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #835, opened 2026-04-19
feat: ZC1816 — warn on docker/podman/nerdctl commit un-reproducible image

afadesigns:
ZC1816 — snapshot commit produces un-reproducible image

What: detect docker commit, podman commit, nerdctl commit.
Why: the subcommand snapshots a running container's filesystem into a new image. There is no Dockerfile, so the build is not reproducible; the snapshot inherits whatever /tmp scratch, shell history, environment variables, and often credentials the container held; and the image's layer metadata records only the container id, not what was actually installed.
Fix suggestion: build from a Dockerfile (or docker buildx build) so the image can be regenerated from source. Reserve docker commit for one-off rescue work on a local image about to be discarded.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1816_test.go
+++ b/pkg/katas/katatests/zc1816_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #836, opened 2026-04-19
feat: ZC1817 — warn on git push --delete / -d / :branch remote branch removal

afadesigns:
ZC1817 — scripted remote branch deletion

What: detect git push with --delete, -d, or the legacy colon form git push REMOTE :BRANCH.
Why: remote branch deletion is irreversible server-side. Local reflog does not cover it. Open pull requests become orphaned, CI runners that pin to the branch lose the target, and recovery needs the last commit SHA which may only live in somebody else's local clone.
Fix suggestion: confirm the remote name, check with git branch -r / gh pr list --head BRANCH first, and prefer the hosting platform's auto-delete-after-merge setting over a scripted push.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1817_test.go
+++ b/pkg/katas/katatests/zc1817_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #837, opened 2026-04-19
feat: ZC1818 — warn on rsync --delete without --dry-run or --max-delete

afadesigns:
ZC1818 — rsync --delete without a preview

What: detect rsync with --delete / --delete-before / --delete-during / --delete-delay / --delete-after / --delete-excluded / --delete-missing-args, where neither --dry-run / -n / --itemize-changes nor a combined short flag containing n is present. Also catches parser-mangled forms where --delete becomes the SimpleCommand name.
Why: the flag removes anything in DST that is not in SRC. An accidentally empty SRC (typo, unmounted mount point, wrong credentials pointing at an empty remote) wipes DST with no undo.
Fix suggestion: preview with rsync -av --delete --dry-run SRC DST, and cap the blast radius with --max-delete=N so a bad SRC cannot cascade.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1818_test.go
+++ b/pkg/katas/katatests/zc1818_test.go
@@ -1,0 +1,51 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #838, opened 2026-04-19
feat: ZC1819 — warn on xattr -d com.apple.quarantine / -cr stripping Gatekeeper

afadesigns:
ZC1819 — stripping macOS Gatekeeper quarantine

What: detect xattr -d com.apple.quarantine FILE and xattr -cr PATH (clear recursively).
Why: macOS sets com.apple.quarantine on every file downloaded from the internet, and Gatekeeper uses it to trigger the first-run notarization / signature check. Removing the xattr lets the binary run with no prompt — \"we vetted the binary\" becomes \"we trust whatever landed in the download folder\".
Fix suggestion: verify the signature (codesign --verify) and notarization (spctl --assess --type execute) before stripping, or install via xip/installer packages so Gatekeeper stays in the loop.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1819_test.go
+++ b/pkg/katas/katatests/zc1819_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #839, opened 2026-04-19
feat: ZC1820 — warn on netplan apply without prior netplan try rollback

afadesigns:
ZC1820 — netplan apply without rollback window

What: detect netplan apply (any additional flags).
Why: apply regenerates the rendered backend config (systemd-networkd / NetworkManager) and brings it live right away. A mistake in the YAML — wrong interface name, missing dhcp4, bad addresses, conflicting routes — drops the admin SSH session, and recovery needs console access.
Fix suggestion: run netplan try first (applies the config, waits for confirmation, auto-reverts if no keypress within the timeout), then fall through to netplan apply once the try window elapses successfully.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1820_test.go
+++ b/pkg/katas/katatests/zc1820_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #840, opened 2026-04-19
feat: ZC1821 — error on diskutil eraseDisk / secureErase / partitionDisk

afadesigns:
ZC1821 — macOS diskutil destructive subcommands

What: detect diskutil with eraseDisk, eraseVolume, secureErase, zeroDisk, randomDisk, reformat, eraseCD, erasePartitions, or partitionDisk.
Why: each rewrites disk or volume state with no Time Machine snapshot or APFS preservation. A wrong /dev/diskN (especially after a reboot that reordered BSD names) erases the wrong drive, and the only recovery is an offline backup.
Fix suggestion: resolve the target by diskutil info -plist or mount-point (not by index), run diskutil list immediately before, and require a typed confirmation.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1821_test.go
+++ b/pkg/katas/katatests/zc1821_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #841, opened 2026-04-19
feat: ZC1822 — error on csrutil disable / spctl --master-disable disabling macOS SIP

afadesigns:
ZC1822 — disabling macOS system integrity / Gatekeeper

What: detect csrutil disable, spctl --master-disable, spctl --global-disable, spctl kext-consent disable (plus parser-mangled master-disable / global-disable forms).
Why: csrutil disable turns off System Integrity Protection — the kernel stops blocking writes under /System, /bin, /sbin and unsigned kexts can load. spctl --master-disable / --global-disable removes Gatekeeper enforcement so any downloaded binary runs without a prompt. Neither has a legitimate provisioning use; both are high-value persistence steps for malware.
Fix suggestion: re-enable with csrutil enable (in recovery) and spctl --master-enable.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1822_test.go
+++ b/pkg/katas/katatests/zc1822_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #842, opened 2026-04-19
feat: ZC1823 — warn on keytool -import -noprompt trust store poisoning

afadesigns:
ZC1823 — Java trust store import without fingerprint check

What: detect keytool -import / -importcert / -importkeystore paired with -noprompt.
Why: keytool -import -noprompt adds the target certificate to the JVM trust store without showing its SHA-256 fingerprint or asking for confirmation. If the cert came from an HTTP download, a shared temp dir, or was fetched by a wrong provisioning step, the JVM pins an attacker's CA as trusted and happily verifies everything signed against it.
Fix suggestion: drop -noprompt; pre-verify with keytool -printcert -file CERT and store (alias, fingerprint) pairs in an audited inventory before touching the trust store.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1823_test.go
+++ b/pkg/katas/katatests/zc1823_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #843, opened 2026-04-19
feat: ZC1824 — warn on kubectl drain --disable-eviction bypassing PodDisruptionBudget

afadesigns:
ZC1824 — drain that ignores PodDisruptionBudget

What: detect kubectl drain <node> --disable-eviction / --disable-eviction=true.
Why: the Eviction API honours PodDisruptionBudget contracts (minAvailable, maxUnavailable). --disable-eviction falls back to raw DELETE, so pods are removed regardless of the PDB. On a multi-replica service a rolling drain turns into a hard outage.
Fix suggestion: fix the blocking PDB — raise minAvailable, wait for replicas to reschedule, or negotiate with the workload owner — instead of flipping the flag off.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1824_test.go
+++ b/pkg/katas/katatests/zc1824_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #844, opened 2026-04-19
feat: ZC1825 — warn on scp -O forcing legacy SCP wire protocol

afadesigns:
ZC1825 — scp -O downgrades to legacy protocol

What: detect scp -O anywhere in the argument list.
Why: OpenSSH 9.0 switched scp to SFTP by default to close the filename-injection class (CVE-2020-15778 and friends) that came from piping through a remote shell. -O forces the legacy SCP wire protocol, putting the connection back on the old code path.
Fix suggestion: drop -O; use sftp or upgrade the remote server if SFTP is the problem. Reserve -O for a named compatibility bug.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1825_test.go
+++ b/pkg/katas/katatests/zc1825_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #845, opened 2026-04-19
feat: ZC1826 — warn on install -m 4xxx/2xxx/6xxx dropping setuid binary

afadesigns:
ZC1826 — install with setuid / setgid mode

What: detect install with -m MODE (or -mMODE attached) where MODE is a four-digit octal whose leading digit is 4 (setuid), 2 (setgid), or 6 (both).
Why: install applies MODE atomically at copy time, so the setuid/setgid bit goes live in a single step with no intermediate chmod where a tripwire would fire. If DEST is on $PATH (e.g. /usr/local/bin) every local user can invoke the elevated binary.
Fix suggestion: only install setuid binaries from trusted builds you have reviewed, and prefer narrow capabilities (setcap cap_net_bind_service+ep) over broad setuid.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1826_test.go
+++ b/pkg/katas/katatests/zc1826_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #846, opened 2026-04-19
feat: ZC1827 — error on npm/pnpm/yarn unpublish breaking pinned downstreams

afadesigns:
ZC1827 — npm unpublish breaks the dependency graph

What: detect npm / pnpm / yarn with unpublish as the subcommand.
Why: npm unpublish PKG@VERSION removes a published version from the registry. Every downstream that pinned that version — directly or through a transitive lockfile entry — fails to install on the next npm ci / CI run. This is the 2016 left-pad pattern; npm now restricts unpublish to 72 hours and gates behind --force, but the blast radius within the window is the whole ecosystem that depended on the package.
Fix suggestion: use npm deprecate PKG@VERSION 'reason' — the version stays resolvable, installs print a warning, and downstreams can pin forward on their own schedule.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1827_test.go
+++ b/pkg/katas/katatests/zc1827_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #847, opened 2026-04-19
feat: ZC1828 — warn on gcore / strace -p dumping target process memory

afadesigns:
ZC1828 — ptrace attach dumps target memory

What: detect gcore PID and strace -p PID (excluding help / version flags).
Why: gcore writes a core dump of the running process to disk; strace -p streams every syscall. Both attach via ptrace and expose target memory, stack, environment variables, and argument buffers. Credentials, TLS session keys, and env-var secrets are all readable. A root-run script that attaches to another user's process extracts whatever the user has.
Fix suggestion: keep production scripts off ptrace; use coredumpctl on a captured core, or perf counters for syscall statistics without memory read. ZC1603 already covers gdb/ltrace; this kata adds the gcore / strace siblings.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1828_test.go
+++ b/pkg/katas/katatests/zc1828_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #848, opened 2026-04-19
feat: ZC1829 — warn on tailscale/wg-quick/nmcli down dropping VPN-carried SSH

afadesigns:
ZC1829 — tearing down the VPN from within an SSH session

What: detect tailscale down, wg-quick down <iface>, and nmcli connection / con / c down <name>.
Why: these commands tear the VPN link down in place with no grace or rollback. If the admin SSH session rides over the tunnel, the script cuts itself off mid-run — the TCP connection freezes and recovery needs console / out-of-band access.
Fix suggestion: schedule recovery via systemd-run --on-active=30s --unit=recover <bring-it-back-up> so the VPN returns before the timer expires, or run the command from the host's console / out-of-band path.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1829_test.go
+++ b/pkg/katas/katatests/zc1829_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #849, opened 2026-04-19
feat: ZC1830 — warn on unsetopt NOMATCH silencing unmatched-glob safety

afadesigns:
ZC1830 — silencing the unmatched-glob error

What: detect unsetopt NOMATCH and setopt NO_NOMATCH (case/underscore folded).
Why: NOMATCH is on by default in Zsh — an unmatched glob errors out instead of silently passing through. Disabling it reverts to POSIX-sh behaviour: the pattern is handed to the command verbatim. rm *.lgo fails loudly, but commands that accept arbitrary strings silently process the literal *.log pattern instead of files.
Fix suggestion: use the (N) null-glob qualifier per glob (*.log(N)) or scope the change inside a function with setopt LOCAL_OPTIONS; setopt NULL_GLOB so the rest of the script keeps the default safety.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1830_test.go
+++ b/pkg/katas/katatests/zc1830_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #850, opened 2026-04-19
feat: ZC1831 — error on systemctl stop/disable/mask ssh/sshd locking out remote login

afadesigns:
ZC1831 — shutting down sshd over SSH

What: detect systemctl stop / disable / mask applied to ssh, sshd, ssh.service, sshd.service, ssh.socket, sshd.socket, openssh-server, openssh.
Why: the SSH daemon owns the next remote login. stop closes it for this boot, disable survives reboot, mask does both and blocks manual start. Existing sessions survive for a while (spawned per-session sshd), but any reconnect / CI follow-up that needs to ssh back in gets Connection refused. Recovery needs console or out-of-band access.
Fix suggestion: use systemctl reload sshd for config changes; if the host is being retired, touch sshd last.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1831_test.go
+++ b/pkg/katas/katatests/zc1831_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

redteamx (approved):
LGTM.

PR #852, opened 2026-04-19
feat: ZC1832 — warn on Zsh limit coredumpsize unlimited enabling core dumps

afadesigns:
ZC1832 — Zsh limit builtin enabling unbounded core dumps

What: detect Zsh limit coredumpsize unlimited (with optional -h hard-limit flag) and unlimit coredumpsize.
Why: Zsh's limit is the csh-style sibling of ulimit. limit coredumpsize unlimited has the same effect as ulimit -c unlimited (covered by ZC1495): a crashing setuid or key-holding process leaves its address space on disk as a world-readable core file.
Fix suggestion: leave the coredump ceiling at the distro default (usually 0 for non-debug sessions), or use systemd-coredump with restricted permissions when post-mortem data is genuinely needed.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1832_test.go
+++ b/pkg/katas/katatests/zc1832_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #853, opened 2026-04-19
feat: ZC1742 — flag `mc alias set ACCESS_KEY SECRET_KEY` (S3 keys in argv)

afadesigns:
ZC1742 — flag `mc alias set NAME URL ACCESS_KEY SECRET_KEY`

What: MinIO `mc alias set` (and legacy `mc config host add`) with positional S3 keys.
Why: keys land in argv, visible in `ps`, `/proc/<pid>/cmdline`, shell history.
Fix suggestion: drop trailing keys (mc prompts) or use `MC_HOST_<alias>=URL` env-var form.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1742_test.go
+++ b/pkg/katas/katatests/zc1742_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

PR #854, opened 2026-04-19
feat: ZC1833 — warn on `unsetopt WARN_CREATE_GLOBAL` hiding accidental globals

afadesigns:
ZC1833 — warn on `unsetopt WARN_CREATE_GLOBAL`

What: `unsetopt WARN_CREATE_GLOBAL` (or `setopt NO_WARN_CREATE_GLOBAL`).
Why: disables Zsh's warning when a function assigns to a name without `local`/`typeset` — leaks state into caller scope.
Fix suggestion: leave the option on; declare variables with `local`/`typeset`; scope with `LOCAL_OPTIONS` if you must disable.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1833_test.go
+++ b/pkg/katas/katatests/zc1833_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #855, opened 2026-04-19
feat: ZC1834 — error on `tc qdisc … netem loss 100%` blackholing live interface

afadesigns:
ZC1834 — error on `tc qdisc add/replace dev X root netem loss 100%`

What: installs a netem qdisc that drops 100% of packets (or 100% corrupt/duplicate).
Why: canonical remote-SSH lockout — the kernel applies it, the next ACK never arrives.
Fix suggestion: stage on a secondary interface, or wrap with `at now + N minutes … tc qdisc del …` auto-recovery.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1834_test.go
+++ b/pkg/katas/katatests/zc1834_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #858, opened 2026-04-19
feat: ZC1837 — error on `chmod` opening /dev/kvm|mem|kmem|port to the world

afadesigns:
ZC1837 — `chmod` granting non-owner access to privileged kernel devices

What: flags `chmod <mode-with-other-bits> /dev/{kvm,mem,kmem,port}`.
Why: these nodes expose physical memory, kernel memory, raw I/O ports, and the full hypervisor API — opening to "other" is a classic local-privesc vector.
Fix suggestion: add the user to the managed group (e.g. `kvm`) or ship a proper udev rule — never blanket-chmod the device from a script.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1837_test.go
+++ b/pkg/katas/katatests/zc1837_test.go
@@ -1,0 +1,75 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #859, opened 2026-04-19
feat: ZC1838 — warn on `setopt GLOB_DOTS` making bare `*` match hidden files

afadesigns:
ZC1838 — warn on `setopt GLOB_DOTS` / `unsetopt NO_GLOB_DOTS`

What: flips the Zsh default where `*` skips dotfiles.
Why: subsequent `rm *`, `cp -r *`, `chmod *` silently touch `.git/`, `.env`, `.ssh/`.
Fix suggestion: leave the option alone; request dotfiles per-glob with `*(D)` or the explicit `.* *` pattern.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1838_test.go
+++ b/pkg/katas/katatests/zc1838_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #860, opened 2026-04-19
feat: ZC1839 — warn on disabling network time sync (NTP/chronyd/timesyncd)

afadesigns:
ZC1839 — disabling NTP / time-sync

What: `timedatectl set-ntp false|no|0|off`, or `systemctl disable|mask|stop` of `systemd-timesyncd` / `chronyd` / `ntpd`.
Why: clock drift breaks TLS cert validity windows, Kerberos tickets, TOTP, and log ordering — silently after days.
Fix suggestion: keep NTP enabled in production; isolate frozen clocks to namespaces or CI containers.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1839_test.go
+++ b/pkg/katas/katatests/zc1839_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #861, opened 2026-04-19
feat: ZC1840 — error on `openssl -k PASSWORD` legacy flag leaking secret in argv

afadesigns:
ZC1840 — `openssl -k PASSWORD` legacy flag

What: detects the pre-OpenSSL-3 short-form `-k PASSWORD` on `openssl enc`/`rsa`/`pkcs12` subcommands.
Why: password rides in argv — visible to `ps`, `/proc/<pid>/cmdline`, shell history, audit logs.
Fix suggestion: use `-pass env:VARNAME`, `-pass file:PATH`, or `-pass fd:N` instead.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1840_test.go
+++ b/pkg/katas/katatests/zc1840_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #862, opened 2026-04-19
feat: ZC1841 — error on `curl --proxy-insecure` disabling proxy-leg TLS

afadesigns:
ZC1841 — `curl --proxy-insecure`

What: detects `curl --proxy-insecure` in both flag-first and flag-trailing positions.
Why: disables TLS verification on the HTTPS-proxy hop — on-path attacker can decrypt the tunnel (including `Authorization:` headers).
Fix suggestion: install the proxy CA and use `--proxy-cacert PATH` / `CURL_CA_BUNDLE`.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1841_test.go
+++ b/pkg/katas/katatests/zc1841_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #863, opened 2026-04-19
feat: ZC1842 — warn on `setopt CDABLE_VARS` silently rewriting `cd` via variable

afadesigns:
ZC1842 — `setopt CDABLE_VARS`

What: flags `setopt CDABLE_VARS` / `unsetopt NO_CDABLE_VARS`.
Why: a failed `cd NAME` is retried as `cd $NAME` — a typo that shares a variable name silently jumps to the variable's path; every later relative path runs from the wrong root.
Fix suggestion: keep the option in `~/.zshrc` (interactive shortcut); in scripts use `cd "$dir" || exit`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1842_test.go
+++ b/pkg/katas/katatests/zc1842_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #864, opened 2026-04-19
feat: ZC1843 — warn on container `--cgroup-parent` pointing at host-managed slice

afadesigns:
ZC1843 — `docker/podman run --cgroup-parent=/system.slice|/init.scope|/`

What: flags `--cgroup-parent` where the parent is the host root slice or systemd-managed slice.
Why: container sits side-by-side with host services, engine-wide memory/CPU caps no longer apply — runaway container can starve sshd or kubelet.
Fix suggestion: drop the flag; use per-run `--memory`/`--cpus`/`--pids-limit` instead.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1843_test.go
+++ b/pkg/katas/katatests/zc1843_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #866, opened 2026-04-19
feat: ZC1845 — warn on `setopt PATH_DIRS` rerouting slash-bearing commands via `$PATH`

afadesigns:
ZC1845 — `setopt PATH_DIRS`

What: flags `setopt PATH_DIRS` / `unsetopt NO_PATH_DIRS`.
Why: a command containing `/` that is not directly executable is silently retried against every `$path` entry — `subdir/cmd` can run `/usr/local/bin/subdir/cmd` or a user-writable tree.
Fix suggestion: leave the option off; call locals as `./subdir/cmd` or use absolute paths.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1845_test.go
+++ b/pkg/katas/katatests/zc1845_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #867, opened 2026-04-19
feat: ZC1846 — warn on `certbot --force-renewal` burning ACME rate limits

afadesigns:
ZC1846 — `certbot renew/certonly --force-renewal`

What: flags the `--force-renewal` flag on certbot's `renew`, `certonly`, and `run` subcommands.
Why: reissues a certificate regardless of remaining validity; in a cron it eats Let's Encrypt's 50-certs-per-domain / 7-day rate limit and blocks fresh issuance during outages.
Fix suggestion: drop the flag; let the default 30-day-before-expiry gate do its job, or run a one-off manual reissue when truly needed.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1846_test.go
+++ b/pkg/katas/katatests/zc1846_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #868, opened 2026-04-19
feat: ZC1847 — warn on `setopt CHASE_LINKS` silently resolving `cd` symlinks

afadesigns:
ZC1847 — `setopt CHASE_LINKS`

What: flags `setopt CHASE_LINKS` / `unsetopt NO_CHASE_LINKS`.
Why: every subsequent `cd` resolves symlinks to the physical inode, breaking blue/green-style `current` symlink navigation and `..` traversal.
Fix suggestion: keep the option off; request per-call physical resolution with `cd -P target` when a specific line needs it.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1847_test.go
+++ b/pkg/katas/katatests/zc1847_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #869, opened 2026-04-19
feat: ZC1848 — warn on `ssh -o CheckHostIP=no` silencing IP-mismatch warning

afadesigns:
ZC1848 — `ssh -o CheckHostIP=no`

What: flags `ssh`/`scp`/`sftp` invocations that set `CheckHostIP=no|false|0|off` (both split `-o KEY=VALUE` and attached `-oKEY=VALUE` forms).
Why: disables the IP-mismatch warning for known hosts — DNS spoofing paired with a leaked/stolen host key becomes invisible.
Fix suggestion: leave the default on; for load-balanced pools use `HostKeyAlias` instead.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1848_test.go
+++ b/pkg/katas/katatests/zc1848_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #870, opened 2026-04-19
feat: ZC1849 — warn on `setopt ALL_EXPORT` auto-exporting every assignment

afadesigns:
ZC1849 — `setopt ALL_EXPORT`

What: flags `setopt ALL_EXPORT` / `unsetopt NO_ALL_EXPORT` (POSIX `set -a` equivalent).
Why: every later `var=value` becomes `export var=value` — secrets like `password=$(cat secret)` leak into every child's env, visible via `/proc/<pid>/environ` and any `ps e` later in the pipeline.
Fix suggestion: drop the `setopt`; use explicit `export VAR=value`, or scope inside `LOCAL_OPTIONS` function.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1849_test.go
+++ b/pkg/katas/katatests/zc1849_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #871, opened 2026-04-19
feat: ZC1850 — warn on `ssh -o LogLevel=QUIET` silencing security diagnostics

afadesigns:
ZC1850 — `ssh -o LogLevel=QUIET|fatal|error`

What: flags `ssh`/`scp`/`sftp` with `LogLevel` set to `QUIET`, `FATAL`, or `ERROR` in split or attached `-o` form.
Why: suppresses every informational/warning message from ssh — host-key changes, key-exchange downgrades, agent-forwarding denials, canonical-hostname rewrites — so MITM events leave no trace on stderr.
Fix suggestion: keep the default `INFO` level; capture stderr to a log if the noise is too much.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1850_test.go
+++ b/pkg/katas/katatests/zc1850_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #872, opened 2026-04-19
feat: ZC1851 — warn on `unsetopt FUNCTION_ARGZERO` hiding function $0

afadesigns:
ZC1851 — `unsetopt FUNCTION_ARGZERO`

What: flags `unsetopt FUNCTION_ARGZERO` / `setopt NO_FUNCTION_ARGZERO`.
Why: reverts Zsh to POSIX-sh `$0` behaviour — `$0` inside a function points at the outer script, so `log "$0: bad input"` helpers and `case $0` dispatchers silently break.
Fix suggestion: keep the option on; if one helper needs the outer script name, use `$ZSH_ARGZERO` / `$ZSH_SCRIPT` explicitly.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1851_test.go
+++ b/pkg/katas/katatests/zc1851_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #873, opened 2026-04-19
feat: ZC1852 — error on `firewall-cmd --panic-on` dropping every packet

afadesigns:
ZC1852 — `firewall-cmd --panic-on`

What: flags `firewall-cmd --panic-on` in mangled-name and trailing-flag shapes.
Why: firewalld panic mode drops every inbound/outbound packet regardless of zone — running over remote SSH loses the session instantly.
Fix suggestion: use targeted zone rules; if you truly need panic, gate behind `at now + N minutes … firewall-cmd --panic-off` auto-recovery.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1852_test.go
+++ b/pkg/katas/katatests/zc1852_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #874, opened 2026-04-19
feat: ZC1853 — warn on `setopt MARK_DIRS` appending trailing `/` to dir globs

afadesigns:
ZC1853 — `setopt MARK_DIRS`

What: flags `setopt MARK_DIRS` / `unsetopt NO_MARK_DIRS`.
Why: every glob-matched directory gains a trailing `/` — `[[ -f "$f" ]]` rejects it, `rm -f *` silently skips, basename-indexed hash maps double-count.
Fix suggestion: keep the option off; request dirs-only per-glob with the `*(/)` qualifier.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1853_test.go
+++ b/pkg/katas/katatests/zc1853_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #875, opened 2026-04-19
feat: ZC1854 — error on `yum-config-manager`/`dnf`/`zypper` adding plaintext HTTP repo

afadesigns:
ZC1854 — plaintext-HTTP package repos

What: flags `yum-config-manager --add-repo http://…`, `dnf config-manager --add-repo http://…`, and `zypper addrepo http://…`.
Why: metadata and RPMs fetched without TLS — on-path attacker substitutes packages and can strip `repo_gpgcheck=1` from the unsigned `.repo` before it lands.
Fix suggestion: use the `https://` mirror (every major distro publishes one); pin `gpgkey=file:///etc/pki/...` in the `.repo`.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1854_test.go
+++ b/pkg/katas/katatests/zc1854_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #876, opened 2026-04-19
feat: ZC1855 — avoid `$GROUPS`, use Zsh `$groups` associative array

afadesigns:
ZC1855 — `$GROUPS` Bash-ism

What: flags any command arg referencing `$GROUPS` or `${GROUPS[...]}`.
Why: `$GROUPS` is a Bash magic array; Zsh does not populate it. `${GROUPS[@]}` silently expands to empty, skipping group-membership checks.
Fix suggestion: use Zsh's `$groups` associative array (`${(k)groups}` for names, `${(v)groups}` for GIDs) or fall back to `id -Gn`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1855_test.go
+++ b/pkg/katas/katatests/zc1855_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #877, opened 2026-04-19
feat: ZC1856 — warn on `unset arr[N]` Bash-ism that leaves Zsh array intact

afadesigns:
ZC1856 — `unset arr[N]` Bash-ism

What: flags `unset <identifier>[<subscript>]` on any command named `unset`.
Why: Bash removes the N-th element; Zsh tries to unset a parameter literally named `arr[N]`, finds none, returns success — the array is unchanged. Silent bug.
Fix suggestion: use `arr[N]=()` for index deletion, or rebuild with `arr=("${(@)arr:#pattern}")`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1856_test.go
+++ b/pkg/katas/katatests/zc1856_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #880, opened 2026-04-19
feat: ZC1859 — warn on `unsetopt MULTIOS` dropping Zsh multi-redirection

afadesigns:
ZC1859 — `unsetopt MULTIOS`

What: flags `unsetopt MULTIOS` / `setopt NO_MULTIOS`.
Why: disables Zsh's default multi-output redirection; `cmd >a >b` silently keeps only the last file, log collectors tailing the earlier file stop receiving new lines.
Fix suggestion: keep the option on script-wide; if one line genuinely needs POSIX single-output semantics, wrap it in a function with `setopt LOCAL_OPTIONS; unsetopt MULTIOS`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1859_test.go
+++ b/pkg/katas/katatests/zc1859_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #881, opened 2026-04-19
feat: ZC1860 — warn on `hostnamectl set-hostname` leaving running services on old name

afadesigns:
ZC1860 — live hostname change

What: flags `hostnamectl set-hostname NEW`, `hostnamectl hostname NEW`, and bare `hostname NEW`.
Why: kernel hostname updates live, but every service that already called `gethostname()` keeps the old value until restart — syslog tags, Prometheus labels, and TLS `subjectAltName` stay stale.
Fix suggestion: ship the new hostname via cloud-init / Ignition at provisioning, or plan a reboot window.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1860_test.go
+++ b/pkg/katas/katatests/zc1860_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #882, opened 2026-04-19
feat: ZC1861 — warn on `setopt OCTAL_ZEROES` reinterpreting leading-zero integers

afadesigns:
ZC1861 — `setopt OCTAL_ZEROES`

What: flags `setopt OCTAL_ZEROES` / `unsetopt NO_OCTAL_ZEROES`.
Why: reverts arithmetic to POSIX-sh octal semantics — `(( n = 0100 ))` assigns 64 instead of 100, so timestamp, phone-prefix, and file-mode parsing silently lies.
Fix suggestion: keep the option off; use explicit `8#100` or `$(( 8#$val ))` when you really want octal.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1861_test.go
+++ b/pkg/katas/katatests/zc1861_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #883, opened 2026-04-19
feat: ZC1862 — warn on `ssh-keygen -R HOST` scrubbing known-hosts silently

afadesigns:
ZC1862 — `ssh-keygen -R HOST`

What: flags `ssh-keygen -R HOST` in automation.
Why: scrubs the known-hosts entry so the next `ssh` silently re-trusts whatever key the network returns — a MITM waiting for the rebuild window slips in without a prompt, especially paired with `StrictHostKeyChecking=no`.
Fix suggestion: fetch the new fingerprint out-of-band (`ssh-keyscan | ssh-keygen -lf -`), verify, then pin it in a managed `known_hosts`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1862_test.go
+++ b/pkg/katas/katatests/zc1862_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #884, opened 2026-04-19
feat: ZC1863 — warn on `unsetopt CASE_GLOB` flipping globs to case-insensitive

afadesigns:
ZC1863 — `unsetopt CASE_GLOB`

What: flags `unsetopt CASE_GLOB` / `setopt NO_CASE_GLOB`.
Why: every later glob and `[[ == pat ]]` silently goes case-insensitive — `rm *.log` sweeps `APP.LOG`, case-distinguishing dispatchers (e.g. `README` vs `readme`) collide.
Fix suggestion: keep the option on script-wide; request case-folding per-pattern with Zsh's `(#i)pattern` qualifier.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1863_test.go
+++ b/pkg/katas/katatests/zc1863_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #886, opened 2026-04-19
feat: ZC1865 — warn on `unsetopt CASE_MATCH` flipping regex to case-insensitive

afadesigns:
ZC1865 — `unsetopt CASE_MATCH`

What: flags `unsetopt CASE_MATCH` / `setopt NO_CASE_MATCH`.
Why: every later `[[ =~ ]]` and `[[ == pat ]]` goes case-insensitive — `Admin` matches `ADMIN`, regex dispatchers that distinguished case silently collapse.
Fix suggestion: keep the option on; scope case-folding per-pattern with `(#i)pattern`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1865_test.go
+++ b/pkg/katas/katatests/zc1865_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #887, opened 2026-04-19
feat: ZC1866 — warn on `docker exec -u 0` bypassing image's non-root USER

afadesigns:
ZC1866 — `docker/podman exec -u 0|root`

What: flags `docker exec`/`podman exec` with `-u 0`, `-u root`, `--user=0`, `--user=root`, or UID-group forms like `0:...`.
Why: hardened images set a non-root `USER`; `-u 0` drops a root shell that bypasses that and equals host root without userns remap.
Fix suggestion: keep execs as the container's configured user; bake required privileges into the image so `-u 0` is never needed.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1866_test.go
+++ b/pkg/katas/katatests/zc1866_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #888, opened 2026-04-19
feat: ZC1867 — warn on `unsetopt GLOB` turning pattern expansion off script-wide

afadesigns:
ZC1867 — `unsetopt GLOB`

What: flags `unsetopt GLOB` / `setopt NO_GLOB` (POSIX `set -f` equivalent).
Why: every subsequent pattern goes through verbatim — `rm *.log` tries to remove a file literally named `*.log`, `for f in *.txt` loops once over the pattern string, array-length checks always return 1.
Fix suggestion: keep the option on script-wide; quote specific args (`'*.log'`) or scope inside a function with `setopt LOCAL_OPTIONS; setopt NO_GLOB`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1867_test.go
+++ b/pkg/katas/katatests/zc1867_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #889, opened 2026-04-19
feat: ZC1868 — error on `gcloud config set auth/disable_ssl_validation true`

afadesigns:
ZC1868 — gcloud TLS off

What: flags `gcloud config set auth/disable_ssl_validation true|1|on`.
Why: persists the flag in the active gcloud config; every subsequent `gcloud` call skips Google-API cert verification — service-account tokens and deploy credentials become interceptable.
Fix suggestion: `gcloud config unset auth/disable_ssl_validation`; pin a custom CA via `core/custom_ca_certs_file` if a corporate proxy requires it.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1868_test.go
+++ b/pkg/katas/katatests/zc1868_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #891, opened 2026-04-19
feat: ZC1870 — warn on `setopt GLOB_ASSIGN` glob-expanding RHS of `var=`

afadesigns:
ZC1870 — `setopt GLOB_ASSIGN`

What: flags `setopt GLOB_ASSIGN` / `unsetopt NO_GLOB_ASSIGN`.
Why: turns unquoted `var=*.log` and `cert=~/secrets/*` into filesystem globs — the script captures whatever matches first (or an empty string), and attacker-dropped files can slip into sensitive assignments.
Fix suggestion: keep the option off; use explicit `arr=( *.log )` when you really want glob expansion.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1870_test.go
+++ b/pkg/katas/katatests/zc1870_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #893, opened 2026-04-19
feat: ZC1872 — error on `badblocks -w` overwriting every disk sector

afadesigns:
ZC1872 — `badblocks -w`

What: flags `badblocks -w` / `--write-mode` including combined short-option clusters like `-wsv`.
Why: write-mode bad-block test overwrites every sector with a pattern — a silent data wipe on a populated disk that still returns success.
Fix suggestion: use `badblocks -n` for non-destructive checking; gate destructive runs behind a confirmation prompt and a fresh partition.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1872_test.go
+++ b/pkg/katas/katatests/zc1872_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #894, opened 2026-04-19
feat: ZC1873 — warn on `setopt ERR_RETURN` bailing functions on first non-zero exit

afadesigns:
ZC1873 — `setopt ERR_RETURN`

What: flags `setopt ERR_RETURN` / `unsetopt NO_ERR_RETURN`.
Why: enabling it script-wide makes every function `return` on the first non-zero exit — probing helpers like `test -f`, `grep -q`, `id -u` bail before the fallback branch runs, with no stderr.
Fix suggestion: keep the option off script-wide; scope inside one function that truly wants fail-fast with `setopt LOCAL_OPTIONS; setopt ERR_RETURN`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1873_test.go
+++ b/pkg/katas/katatests/zc1873_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #895, opened 2026-04-19
feat: ZC1874 — warn on `sshuttle -r HOST 0/0` tunneling all traffic via jump host

afadesigns:
ZC1874 — `sshuttle -r HOST 0/0`

What: flags `sshuttle -r user@host` with a catch-all CIDR (`0/0`, `0.0.0.0/0`, `::/0`).
Why: installs a VPN-wide route — every TCP session and DNS lookup egresses through the jump host; a compromise of `user@host` sees the whole workstation's traffic.
Fix suggestion: scope the subnet list to what you actually need (`10.0.0.0/8 172.16.0.0/12 192.168.0.0/16`); prefer `ssh -D` + browser-level rules for single-app tunneling.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1874_test.go
+++ b/pkg/katas/katatests/zc1874_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #896, opened 2026-04-19
feat: ZC1875 — warn on `setopt RC_QUOTES` flipping single-quote escape semantics

afadesigns:
ZC1875 — `setopt RC_QUOTES`

What: flags `setopt RC_QUOTES` / `unsetopt NO_RC_QUOTES`.
Why: reinterprets `''` inside single-quoted strings as a literal apostrophe — `'it''s'` flips from `its` (empty-concat) to `it's`, breaking tokens, SQL fragments, and log strings across the file.
Fix suggestion: keep the option off; use double quotes or `\'` when you need a literal apostrophe.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1875_test.go
+++ b/pkg/katas/katatests/zc1875_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #897, opened 2026-04-19
feat: ZC1876 — warn on `cargo publish --allow-dirty` shipping uncommitted tarball

afadesigns:
ZC1876 — `cargo publish --allow-dirty`

What: flags `cargo publish --allow-dirty`.
Why: skips the clean-tree check and uploads the current on-disk state — local debug prints, uncommitted `Cargo.toml` bumps, and `patch.crates-io` overrides ride into the published version, which cannot be replaced after upload.
Fix suggestion: commit first and publish from a clean checkout; rehearse with `--dry-run`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1876_test.go
+++ b/pkg/katas/katatests/zc1876_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #898, opened 2026-04-19
feat: ZC1877 — warn on `unsetopt SHORT_LOOPS` breaking compact loop syntax

afadesigns:
ZC1877 — `unsetopt SHORT_LOOPS`

What: flags `unsetopt SHORT_LOOPS` / `setopt NO_SHORT_LOOPS`.
Why: disables Zsh's default single-command `for`/`while`/`repeat` bodies — `for f in *.log; print $f` suddenly raises a parse error, and the flip is global so sourced helpers also break.
Fix suggestion: keep the option on script-wide; scope inside a function with `setopt LOCAL_OPTIONS; unsetopt SHORT_LOOPS` if POSIX-strict parsing is genuinely needed.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1877_test.go
+++ b/pkg/katas/katatests/zc1877_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #899, opened 2026-04-19
feat: ZC1878 — warn on `kubectl apply --force-conflicts` stealing field ownership

afadesigns:
ZC1878 — `kubectl apply --force-conflicts`

What: flags `kubectl apply --force-conflicts` (typically paired with `--server-side`).
Why: grabs ownership of every conflicting field from other controllers — HPA replicas, cert-manager certs, sidecar-injector labels — which silently lose reconcile pressure until their next write.
Fix suggestion: resolve the conflict deliberately — drop the disputed fields from your manifest, or hand off via explicit `--field-manager` edit.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1878_test.go
+++ b/pkg/katas/katatests/zc1878_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #901, opened 2026-04-19
feat: ZC1880 — warn on `kubectl annotate|label --overwrite` rewriting controller signals

afadesigns:
ZC1880 — `kubectl annotate|label --overwrite`

What: flags `kubectl annotate --overwrite` and `kubectl label --overwrite`.
Why: annotations and labels are controller protocol — overwriting silently retriggers cert-manager reissue, external-dns record churn, HPA reconcile, or breaks Helm's `last-applied-configuration` tracking.
Fix suggestion: inspect the existing value (`kubectl get -o jsonpath='{.metadata.annotations}'`), drop `--overwrite`, and let conflicts surface as errors so you reconcile deliberately.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1880_test.go
+++ b/pkg/katas/katatests/zc1880_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #902, opened 2026-04-19
feat: ZC1881 — warn on `unsetopt MULTIBYTE` turning string ops into per-byte math

afadesigns:
ZC1881 — `unsetopt MULTIBYTE`

What: flags `unsetopt MULTIBYTE` / `setopt NO_MULTIBYTE`.
Why: reverts every string operation to per-byte math — `${#emoji}` returns 4 instead of 1, substrings slice mid-codepoint, `[[ =~ ]]` Unicode ranges stop matching.
Fix suggestion: keep the option on script-wide; if you truly want byte counts, use `wc -c <<< $str` or `${#${(%)str}}`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1881_test.go
+++ b/pkg/katas/katatests/zc1881_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #903, opened 2026-04-19
feat: ZC1882 — warn on bare `sudo -s` / `sudo su` / `sudo bash` in scripts

afadesigns:
ZC1882 — interactive root shell via sudo

What: flags `sudo -s`, `sudo -i`, `sudo su[ -]`, and `sudo bash|zsh|sh|ksh|dash|ash` when no command/`-c` follows.
Why: in a non-interactive script the spawned root shell hangs on stdin or, worse, drains the script's remaining lines into the new shell — audit trails record "root shell opened" instead of the intended work.
Fix suggestion: pass the actual command to sudo — `sudo /usr/local/bin/setup.sh` — or scope with `sudo su -c "commands"`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1882_test.go
+++ b/pkg/katas/katatests/zc1882_test.go
@@ -1,0 +1,75 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #904, opened 2026-04-19
feat: ZC1883 — warn on `setopt PATH_SCRIPT` redirecting `source` via `$PATH`

afadesigns:
ZC1883 — `setopt PATH_SCRIPT`

What: flags `setopt PATH_SCRIPT` / `unsetopt NO_PATH_SCRIPT`.
Why: lets `.` / `source` fall back to `$PATH` lookup when a literal path misses — a `helper.sh` dropped into `~/bin` or any user-writable `$PATH` entry runs inside the current shell with every exported secret.
Fix suggestion: keep the option off; always source with an explicit path (`./helper.sh`, absolute path).
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1883_test.go
+++ b/pkg/katas/katatests/zc1883_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #905, opened 2026-04-19
feat: ZC1884 — error on credentials in `curl/wget` URL query string

afadesigns:
ZC1884 — credential in URL query

What: flags `curl`/`wget`/`http(ie)` when a URL contains common secret-bearing query keys (`apikey=`, `token=`, `access_token=`, `password=`, `secret=`, `signature=`, `client_secret=` and variants).
Why: query strings are logged by every proxy, CDN, and server access log along the path — credentials tattooed there live forever and leak to downstream teams.
Fix suggestion: move secrets to an HTTP header (`-H "Authorization: Bearer $TOKEN"`), POST body with `--data-urlencode`, or `-u user:` basic auth.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1884_test.go
+++ b/pkg/katas/katatests/zc1884_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #906, opened 2026-04-19
feat: ZC1885 — warn on `setopt CSH_NULL_GLOB` hiding partial glob mismatches

afadesigns:
ZC1885 — `setopt CSH_NULL_GLOB`

What: flags `setopt CSH_NULL_GLOB` / `unsetopt NO_CSH_NULL_GLOB`.
Why: applies csh-style null-glob to patterns in a list — `rm *.lg *.bak` deletes the `.bak` files and silently drops the `.lg` mismatch instead of erroring. Partial failures become invisible.
Fix suggestion: keep the option off; request null-glob per-pattern with the `*(N)` qualifier.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1885_test.go
+++ b/pkg/katas/katatests/zc1885_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #907, opened 2026-04-19
feat: ZC1886 — error on `tee/cp/mv/install/dd` writing system shell-init files

afadesigns:
ZC1886 — persistent foothold via system shell-init write

What: flags `tee`/`cp`/`mv`/`install`/`dd` targeting `/etc/profile`, `/etc/bash.bashrc`, `/etc/zshrc`, `/etc/zsh/*`, `/etc/environment`, `/etc/profile.d/*`.
Why: these files are sourced by every interactive shell (and `/etc/zshenv` by every Zsh invocation) — arbitrary writes become persistent footholds that run on the next root login.
Fix suggestion: stage a temp file, validate in a dry-run login, and drop into place atomically with `install -m 644`.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1886_test.go
+++ b/pkg/katas/katatests/zc1886_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #908, opened 2026-04-19
feat: ZC1887 — warn on `setopt POSIX_TRAPS` flipping EXIT trap scope to shell-exit

afadesigns:
ZC1887 — `setopt POSIX_TRAPS`

What: flags `setopt POSIX_TRAPS` / `unsetopt NO_POSIX_TRAPS`.
Why: flips `trap ... EXIT` inside functions from function-return scope to shell-exit scope — per-invocation cleanup leaks across the whole shell, and Zsh's TRAPEXIT/TRAPZERR function-scoped hooks stop firing.
Fix suggestion: keep the option off script-wide; install top-level `trap ... EXIT` explicitly when you really want shell-scope POSIX behaviour.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1887_test.go
+++ b/pkg/katas/katatests/zc1887_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #909, opened 2026-04-19
feat: ZC1888 — warn on `aws iam create-access-key` minting long-lived static creds

afadesigns:
ZC1888 — `aws iam create-access-key`

What: flags `aws iam create-access-key` invocations.
Why: hands out a static `AKIA.../secret` pair valid forever until someone rotates it — modern AWS deploys should use short-lived credentials (instance profiles, IRSA, Lambda roles, OIDC federation, IAM Identity Center).
Fix suggestion: switch the caller to a role-based short-lived credential flow; if a static key is truly required, store in Secrets Manager and rotate with `aws iam update-access-key`/`delete-access-key`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1888_test.go
+++ b/pkg/katas/katatests/zc1888_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #910, opened 2026-04-19
feat: ZC1889 — error on `skopeo --src-tls-verify=false`/`--dest-tls-verify=false`

afadesigns:
ZC1889 — skopeo TLS bypass

What: flags `skopeo copy/inspect/sync` invocations with `--src-tls-verify=false|0|no|off`, `--dest-tls-verify=…`, or bare `--tls-verify=…`.
Why: dropping cert verification on either end of an image copy/mirror lets an on-path attacker swap manifest or layer contents without a warning — classic supply-chain bypass.
Fix suggestion: pin a private CA via `--src-cert-dir` / `--dest-cert-dir`, or fix the upstream cert instead.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1889_test.go
+++ b/pkg/katas/katatests/zc1889_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #911, opened 2026-04-19
feat: ZC1890 — error on `kadmin -w PASS` embedding Kerberos admin password in argv

afadesigns:
ZC1890 — Kerberos password in argv

What: flags `kadmin`, `kadmin.local`, and `kpasswd` with `-w PASSWORD`.
Why: the admin principal's password lands in argv — visible to `ps`, `/proc/<pid>/cmdline`, shell history, and CI logs. An attacker who sees it can edit the realm.
Fix suggestion: use `-k -t /etc/krb5.keytab` (keytab root-only) or pipe the password through stdin with batch mode; never in argv.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1890_test.go
+++ b/pkg/katas/katatests/zc1890_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #912, opened 2026-04-19
feat: ZC1891 — error on `kubectl config view --raw` printing full kubeconfig with creds

afadesigns:
ZC1891 — `kubectl config view --raw`

What: flags `kubectl config view --raw` and the `-R` synonym.
Why: the default output redacts `client-certificate-data`, `client-key-data`, `token`, and `password` fields; `--raw` undoes that and prints the full kubeconfig — script-captured stdout ends up in CI logs, journalctl, or Slack pastes.
Fix suggestion: emit specific fields with `-o jsonpath='{…}'`, or pipe into `shred`-deletable temp files.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1891_test.go
+++ b/pkg/katas/katatests/zc1891_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #913, opened 2026-04-19
feat: ZC1892 — error on `install -m 4755|2755|6755` setting setuid/setgid bits

afadesigns:
ZC1892 — setuid install mode

What: flags `install -m <mode>` / `mkdir -m <mode>` where `<mode>` has the setuid (0o4000) or setgid (0o2000) bit set.
Why: installs a setuid binary at build/install time — every future execution becomes a privilege-elevation vector.
Fix suggestion: use `0755` and grant the narrow capability with `setcap cap_xxx+ep`; audit with `find / -perm -4000`.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1892_test.go
+++ b/pkg/katas/katatests/zc1892_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #914, opened 2026-04-19
feat: ZC1893 — warn on `unsetopt BARE_GLOB_QUAL` breaking null-glob qualifier syntax

afadesigns:
ZC1893 — `unsetopt BARE_GLOB_QUAL`

What: flags `unsetopt BARE_GLOB_QUAL` / `setopt NO_BARE_GLOB_QUAL`.
Why: disables Zsh's default per-glob qualifier syntax — `*(N)`, `*(.x)`, `*(Om)` stop being qualifiers and revert to ksh-style alternation, so null-glob idioms silently break.
Fix suggestion: keep the option on; scope inside a function with `setopt LOCAL_OPTIONS; unsetopt BARE_GLOB_QUAL` only if ksh-style parsing is genuinely needed.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1893_test.go
+++ b/pkg/katas/katatests/zc1893_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #915, opened 2026-04-19
feat: ZC1894 — error on `conntrack -F` wiping netfilter connection tracking

afadesigns:
ZC1894 — `conntrack -F`

What: flags `conntrack -F` / `conntrack --flush` (and mangled-name form from the parser).
Why: wipes the netfilter connection-tracking table — every stateful-NAT session, every `ctstate ESTABLISHED` allowance, and every running SSH session loses its entry and is dropped as a new packet.
Fix suggestion: gate behind `at now + N min` auto-recovery, or narrow with `conntrack -D -s <ip>` for a single stuck flow.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1894_test.go
+++ b/pkg/katas/katatests/zc1894_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #916, opened 2026-04-19
feat: ZC1895 — warn on `setopt NUMERIC_GLOB_SORT` reshuffling glob output order

afadesigns:
ZC1895 — `setopt NUMERIC_GLOB_SORT`

What: flags `setopt NUMERIC_GLOB_SORT` / `unsetopt NO_NUMERIC_GLOB_SORT`.
Why: every subsequent glob and array expansion sorts numeric runs numerically instead of lexicographically — `app-1.log app-10.log app-2.log` becomes `app-1.log app-2.log app-10.log`, reshuffling log rotations and "latest-file" pipelines.
Fix suggestion: keep the option off script-wide; request numeric sort per-glob with the `*(n)` qualifier.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1895_test.go
+++ b/pkg/katas/katatests/zc1895_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #917, opened 2026-04-19
feat: ZC1896 — error on `docker run -v /proc|/sys` bind-mounting host kernel ifaces

afadesigns:
ZC1896 — host /proc, /sys bind-mount

What: flags `docker run`/`podman run` / `create` with `-v`, `--volume`, or `--mount` sourcing `/proc`, `/sys`, or any subpath of either.
Why: the container sees every host process's `environ`/`cmdline`, can open `/proc/1/ns/*` for a namespace breakout, and can rewrite cgroup limits — the `:ro` flag does not block ns-fd reuse.
Fix suggestion: use `--cap-add=SYS_PTRACE` or run the monitoring agent on the host rather than inside an untrusted image.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1896_test.go
+++ b/pkg/katas/katatests/zc1896_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #918, opened 2026-04-19
feat: ZC1897 — warn on `setopt SH_GLOB` breaking Zsh-extended glob patterns

afadesigns:
ZC1897 — `setopt SH_GLOB`

What: flags `setopt SH_GLOB` / `unsetopt NO_SH_GLOB`.
Why: forces strict POSIX glob parsing — `*(N)` null-glob qualifier, `<1-10>` numeric range globs, `(alt1|alt2)` alternation, and the `(#i)` flag family all stop parsing and raise errors.
Fix suggestion: keep the option off at script level; scope inside a function with `setopt LOCAL_OPTIONS; setopt SH_GLOB` if POSIX patterns are really needed.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1897_test.go
+++ b/pkg/katas/katatests/zc1897_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #919, opened 2026-04-19
feat: ZC1898 — detect `gpg --export-secret-keys` private-key leak

afadesigns:
ZC1898 — Error on `gpg --export-secret-keys` — private-key material leaks to stdout

What: `gpg --export-secret-keys KEYID` / `--export-secret-subkeys` streams the ASCII-armoured private key to stdout.
Why: One CI log capture, wrong-tty redirect, or world-readable tee permanently exfiltrates the secret key.
Fix suggestion: Back up keys interactively on an air-gapped host, or write to a `umask 077` path and immediately re-encrypt.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1898_test.go
+++ b/pkg/katas/katatests/zc1898_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #920, opened 2026-04-19
feat: ZC1899 — detect `mokutil --disable-validation` Secure Boot bypass

afadesigns:
ZC1899 — Error on `mokutil --disable-validation`

What: `mokutil --disable-validation` queues a shim request to stop validating kernel/modules against enrolled MOK/PK certificates.
Why: At next boot Secure Boot becomes advisory — unsigned kernels or rootkit modules load without prompt.
Fix suggestion: Leave validation on. If you need a custom module, enrol its key with `mokutil --import` and approve via `MokManager`.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1899_test.go
+++ b/pkg/katas/katatests/zc1899_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #921, opened 2026-04-19
feat: ZC1900 — detect `curl --location-trusted` credential-forward on redirect

afadesigns:
ZC1900 — Warn on `curl --location-trusted`

What: `curl --location-trusted` replays `Authorization`, cookies, and `-u user:pass` on every redirect hop.
Why: A 302 to an attacker-controlled origin (or compromised CDN edge) receives the bearer token verbatim.
Fix suggestion: Drop the flag; scope a short-lived token per destination and verify the final hostname before sending secrets.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1900_test.go
+++ b/pkg/katas/katatests/zc1900_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #922, opened 2026-04-19
feat: ZC1901 — detect `setopt POSIX_BUILTINS` Zsh-vs-POSIX scope flip

afadesigns:
ZC1901 — Warn on `setopt POSIX_BUILTINS`

What: `setopt POSIX_BUILTINS` switches Zsh to the POSIX rules for special builtins (`export`, `readonly`, `eval`, `.`, `trap`, `set`, etc.).
Why: Assignments before those builtins stop being local and silently leak into the caller's scope; `command builtin` can also now resolve shell builtins.
Fix suggestion: Leave the option off. Scope any POSIX-specific block with `emulate -LR sh` rather than toggling globally.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1901_test.go
+++ b/pkg/katas/katatests/zc1901_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #923, opened 2026-04-19
feat: ZC1902 — detect `ln -s /dev/null <logfile>` audit/history silencing

afadesigns:
ZC1902 — Error on `ln -s /dev/null <logfile>`

What: A symlink from an audit or shell-history path to `/dev/null` turns every subsequent append into a no-op.
Why: `/var/log/auth.log`, `wtmp`, `~/.bash_history`, `~/.zsh_history` all stop recording — textbook way to cover tracks on a compromised host.
Fix suggestion: Disable the writer (rsyslog rule, `set +o history`) or rotate with `logrotate`. Never redirect an audit/history path into `/dev/null`.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1902_test.go
+++ b/pkg/katas/katatests/zc1902_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #924, opened 2026-04-19
feat: ZC1903 — detect `tee /etc/sudoers*` unvalidated sudoers write

afadesigns:
ZC1903 — Error on `tee /etc/sudoers*`

What: `tee /etc/sudoers` / `tee -a /etc/sudoers.d/<name>` writes a sudoers rule without running `visudo -c`.
Why: A malformed line makes sudo refuse every invocation — you lock yourself out of root recovery.
Fix suggestion: Write the rule to a temp file, run `visudo -cf /tmp/rule`, then `install -m 0440 /tmp/rule /etc/sudoers.d/<name>`.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1903_test.go
+++ b/pkg/katas/katatests/zc1903_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #926, opened 2026-04-19
feat: ZC1905 — detect `ssh -g -L/-D` forwarding on all interfaces

afadesigns:
ZC1905 — Warn on `ssh -g -L …`

What: `ssh -g` flips the default for `-L` and `-D` from `127.0.0.1:port` to `0.0.0.0:port`.
Why: Any host on the same LAN/VPN/WiFi segment can use the tunnel without authenticating to the SSH session.
Fix suggestion: Drop `-g`, pin the bind explicitly with `-L bind_address:port:target:port`, or use a firewall rule.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1905_test.go
+++ b/pkg/katas/katatests/zc1905_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #927, opened 2026-04-19
feat: ZC1906 — detect `setopt POSIX_CD` altering CDPATH semantics

afadesigns:
ZC1906 — Warn on `setopt POSIX_CD`

What: `setopt POSIX_CD` makes `cd`/`chdir`/`pushd` skip `CDPATH` when the argument starts with `/`, `.`, or `..`.
Why: Scripts that relied on Zsh's default (consult `CDPATH` unless the argument is absolute) silently enter different directories.
Fix suggestion: Keep the option off. Scope POSIX-specific code with `emulate -LR sh` instead of toggling globally.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1906_test.go
+++ b/pkg/katas/katatests/zc1906_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #929, opened 2026-04-19
feat: ZC1908 — detect `setopt MAGIC_EQUAL_SUBST` altering arg expansion

afadesigns:
ZC1908 — Warn on `setopt MAGIC_EQUAL_SUBST`

What: `MAGIC_EQUAL_SUBST` tells Zsh that every unquoted argument of the form `identifier=value` gets file expansion on the RHS.
Why: Literal CLI args like `rsync host:dst=~/backup` suddenly tilde-expand — a whole class of arguments silently changes meaning.
Fix suggestion: Keep the option off. If a specific assignment needs expansion, wrap it in quotes or use a temporary variable.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1908_test.go
+++ b/pkg/katas/katatests/zc1908_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #930, opened 2026-04-19
feat: ZC1909 — detect `kexec -l/-e` bypassing bootloader Secure Boot

afadesigns:
ZC1909 — Warn on `kexec -l` / `-e`

What: `kexec -l /path/vmlinuz` stages an alternate kernel; `kexec -e` transfers control to it.
Why: The staged kernel is never verified against enrolled MOK/PK — an attacker who lands root can boot a hostile kernel while leaving /boot untouched.
Fix suggestion: Reserve `kexec` for live-patching / crash-dump flows. Gate behind `sudo` + audit, prefer `systemctl kexec` or a real reboot.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1909_test.go
+++ b/pkg/katas/katatests/zc1909_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #931, opened 2026-04-19
feat: ZC1910 — detect `setopt GLOB_STAR_SHORT` bare-`**` recursion

afadesigns:
ZC1910 — Warn on `setopt GLOB_STAR_SHORT`

What: `GLOB_STAR_SHORT` expands bare `**` (not followed by `/`) as if it were `**/*`.
Why: `rm **` / `grep -l '**'` suddenly recurse into the whole tree. Literal `**` arguments silently become deep globs.
Fix suggestion: Keep the option off. Spell `**/*` explicitly when recursion is actually wanted.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1910_test.go
+++ b/pkg/katas/katatests/zc1910_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #932, opened 2026-04-19
feat: ZC1911 — detect `umount -l` lazy-unmount ghost fds

afadesigns:
ZC1911 — Warn on `umount -l` / `--lazy`

What: Lazy unmount detaches the filesystem from the tree now, defers real cleanup until every open fd closes.
Why: Writers keep writing into a phantom mount, re-mounts stack invisibly, and `lsof`/`fuser` often miss the stale handles.
Fix suggestion: Stop the fd holder first (`lsof`/`fuser`/`systemd-cgls`), then do a normal `umount`. Reserve `-l` for break-glass.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1911_test.go
+++ b/pkg/katas/katatests/zc1911_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #933, opened 2026-04-19
feat: ZC1912 — detect `dhclient -r`/`dhcpcd -k` dropping DHCP lease

afadesigns:
ZC1912 — Warn on `dhclient -r`/`-x` / `dhcpcd -k`

What: Releases or terminates the DHCP lease from inside a script.
Why: On a remote host the next thing that happens is the SSH session drops; VPC automation waiting on a reply never sees one.
Fix suggestion: Pair with a re-acquire (`dhclient -1` / `nmcli device reapply`), or schedule via `systemd-run --on-active=` so the operator is not cut off.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1912_test.go
+++ b/pkg/katas/katatests/zc1912_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #934, opened 2026-04-19
feat: ZC1913 — detect `setopt ALIAS_FUNC_DEF` function/alias shadowing

afadesigns:
ZC1913 — Warn on `setopt ALIAS_FUNC_DEF`

What: `setopt ALIAS_FUNC_DEF` disables the default guardrail that refuses `name () { … }` when `name` is aliased.
Why: The alias is suppressed at definition time, and the function silently shadows the alias afterwards — one sourced rc file replaces the function with the alias.
Fix suggestion: Keep the option off. If the override is intentional, quote the name.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1913_test.go
+++ b/pkg/katas/katatests/zc1913_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #935, opened 2026-04-19
feat: ZC1914 — detect `curl --doh-url`/`--dns-servers` resolver bypass

afadesigns:
ZC1914 — Warn on `curl --doh-url`/`--dns-servers`

What: `--doh-url` and `--dns-servers` make curl route lookups through a caller-specified DNS endpoint.
Why: The detour skips `/etc/hosts`, `systemd-resolved`, `nsswitch`, and split-horizon DNS — the request lands at an IP the operator did not vet.
Fix suggestion: Drop the flag, or pair it with explicit `--resolve` pinning + `--doh-insecure` audit.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1914_test.go
+++ b/pkg/katas/katatests/zc1914_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #937, opened 2026-04-19
feat: ZC1916 — detect `setopt NULL_GLOB` silencing every unmatched glob

afadesigns:
ZC1916 — Warn on `setopt NULL_GLOB`

What: `setopt NULL_GLOB` removes Zsh's default behaviour of erroring on unmatched globs; every later pattern silently expands to nothing.
Why: `cp *.log /dest` becomes `cp /dest` (wrong target), `rm *.tmp` becomes `rm` (argv too short), `for f in *.json` silently no-ops.
Fix suggestion: Use the per-glob `*(N)` qualifier when a single pattern can be empty. Or scope with `setopt LOCAL_OPTIONS NULL_GLOB` inside one function.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1916_test.go
+++ b/pkg/katas/katatests/zc1916_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #938, opened 2026-04-19
feat: ZC1917 — detect `iw dev $IF scan` active WiFi probe from scripts

afadesigns:
ZC1917 — Info on `iw dev scan` / `iwlist scanning`

What: Active WiFi probe-request sweep from a script, requires `CAP_NET_ADMIN`.
Why: Interrupts the current association, broadcasts the host's MAC to every nearby AP, and pollutes logs on the other side.
Fix suggestion: Use cached `iw dev $IF link` / `iwctl station $IF show` for passive lookups. Reserve `scan` for console-approved diagnostics.
Severity: Info

### DIFF
--- a/pkg/katas/katatests/zc1917_test.go
+++ b/pkg/katas/katatests/zc1917_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #939, opened 2026-04-19
feat: ZC1918 — detect `setopt HIST_SUBST_PATTERN` `:s` semantics flip

afadesigns:
ZC1918 — Warn on `setopt HIST_SUBST_PATTERN`

What: Makes the `:s` and `:&` history modifiers (and `${foo:s/pat/rep/}`) match on patterns instead of literal strings.
Why: Text safe as a literal constant (`#`, `^`, `?`, `*`) suddenly gets interpreted as glob metacharacters — replacements that always returned the original string now edit it in surprising ways.
Fix suggestion: Keep the option off. Use `${var//pat/rep}` explicitly when glob substitution is actually wanted.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1918_test.go
+++ b/pkg/katas/katatests/zc1918_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #940, opened 2026-04-19
feat: ZC1919 — detect `ss -K`/`--kill` socket termination

afadesigns:
ZC1919 — Warn on `ss -K` / `--kill`

What: Issues `SOCK_DESTROY` to every socket matching the filter.
Why: Broad filters (`state established`, `dport 22`) happily terminate the SSH session that is running the command, along with every backend keep-alive that matches.
Fix suggestion: Preview with the same filter minus `-K` first; pin to a specific `dst`/`dport`/`state` tuple; gate behind review, not automation.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1919_test.go
+++ b/pkg/katas/katatests/zc1919_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #941, opened 2026-04-19
feat: ZC1920 — detect `setopt VERBOSE` command-echo leak to stderr

afadesigns:
ZC1920 — Warn on `setopt VERBOSE`

What: Zsh name for the POSIX `set -v` flag — shell prints each command line to stderr as it reads.
Why: Scripts that touch secrets (`mysql -pSECRET`, `curl -u user:pass`, `export DB_PASS=…`) leak them to the trace stream.
Fix suggestion: Remove the call and use a logger. If a debug trace is required, scope via `setopt LOCAL_OPTIONS VERBOSE` in a helper.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1920_test.go
+++ b/pkg/katas/katatests/zc1920_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #942, opened 2026-04-19
feat: ZC1921 — detect `systemctl kill -s KILL` skipping ExecStop cleanup

afadesigns:
ZC1921 — Warn on `systemctl kill -s KILL` / `--signal=SIGKILL`

What: Sends SIGKILL to a unit via systemctl, bypassing its `ExecStop=` sequence and `TimeoutStopSec=` budget.
Why: Lockfiles, sockets, and shm segments the service was supposed to unlink survive — next restart often fails with 'address already in use' or a corrupt journal.
Fix suggestion: Default to `systemctl stop` (or `restart`) and let ExecStop run. Reserve `-s KILL` for last-resort recovery with a runbook.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1921_test.go
+++ b/pkg/katas/katatests/zc1921_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #943, opened 2026-04-19
feat: ZC1922 — detect `rpm --import http://` plaintext GPG key fetch

afadesigns:
ZC1922 — Error on `rpm --import http://` / `rpmkeys --import ftp://`

What: Adds an ASCII-armoured GPG key to the system RPM trust store from a plaintext URL.
Why: An on-path attacker swaps the key, and every subsequent package they sign installs cleanly.
Fix suggestion: Serve keys over HTTPS from a TLS-authenticated origin, pin the SHA-256, or stage an offline copy verified with `gpg --verify` against a known fingerprint.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1922_test.go
+++ b/pkg/katas/katatests/zc1922_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #944, opened 2026-04-19
feat: ZC1923 — detect `setopt PRINT_EXIT_VALUE` stderr leak

afadesigns:
ZC1923 — Warn on `setopt PRINT_EXIT_VALUE`

What: Zsh emits `zsh: exit N` to stderr after every foreground command that returns non-zero.
Why: Silent grep/test/curl probes suddenly leak status to supervisors and log aggregators, and tools parsing stderr (`git`, `ssh`, `rsync`) see interleaved shell chatter.
Fix suggestion: Remove the call. Use `|| printf …` on specific commands, or a `precmd` hook for scoped post-mortems.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1923_test.go
+++ b/pkg/katas/katatests/zc1923_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #945, opened 2026-04-19
feat: ZC1924 — detect libguestfs host-side VM disk access

afadesigns:
ZC1924 — Warn on `virt-cat` / `virt-copy-out` / `virt-edit` / `guestfish` / `guestmount` / `virt-customize`

What: libguestfs tools open a VM's disk image from the hypervisor and read or mutate it without going through the guest OS.
Why: Bypasses in-guest permissions, audit, and LUKS keyslots. On a live VM, two writers mounted on the same image risk filesystem corruption.
Fix suggestion: Snapshot the disk first, work on the clone, and prefer in-guest `ssh`/`scp`/`ansible` for anything that does not need out-of-band recovery.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1924_test.go
+++ b/pkg/katas/katatests/zc1924_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #946, opened 2026-04-19
feat: ZC1925 — detect `unsetopt EQUALS` disabling Zsh path/tilde idioms

afadesigns:
ZC1925 — Warn on `unsetopt EQUALS`

What: Zsh `EQUALS` option (on by default) drives `=cmd` → absolute path lookup and tilde-after-colon (`PATH=~/bin:$PATH`).
Why: Turning it off makes `=python`/`=ls` literals, and `PATH=~/bin:$PATH` stops expanding the tilde inside the colon-separated list.
Fix suggestion: Keep the option on. Scope via `setopt LOCAL_OPTIONS; unsetopt EQUALS` inside any function that needs literal `=` arguments.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1925_test.go
+++ b/pkg/katas/katatests/zc1925_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #949, opened 2026-04-19
feat: ZC1928 — detect `setopt SHARE_HISTORY` cross-session history leak

afadesigns:
ZC1928 — Warn on `setopt SHARE_HISTORY`

What: Flushes each command to `$HISTFILE` immediately and tells every other zsh session to re-read it.
Why: A secret typed in a one-off 'private' terminal (`ssh user@host "$PASS"`, `git push https://user:token@…`) shows up in every other terminal's `fc -l` seconds later.
Fix suggestion: Prefer `setopt INC_APPEND_HISTORY` (per-session isolation) and `setopt HIST_IGNORE_SPACE` so a leading space keeps the line out of history.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1928_test.go
+++ b/pkg/katas/katatests/zc1928_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #950, opened 2026-04-19
feat: ZC1929 — detect `cpio -i` extraction without path safety

afadesigns:
ZC1929 — Warn on `cpio -i` / `--extract` without `--no-absolute-filenames`

What: Copy-in mode materialises every path verbatim — absolute paths land where the archive says, relative `..` paths escape the extraction dir.
Why: A rogue initramfs/firmware bundle drops files into `/etc/cron.d/`, `/usr/lib/systemd/system/`, or `~/.ssh/authorized_keys`.
Fix suggestion: Pass `--no-absolute-filenames` and extract into a fresh scratch dir reviewed before `mv` into place.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1929_test.go
+++ b/pkg/katas/katatests/zc1929_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #951, opened 2026-04-19
feat: ZC1930 — detect `unsetopt HASH_CMDS` per-call PATH walks

afadesigns:
ZC1930 — Warn on `unsetopt HASH_CMDS`

What: Disables Zsh's default command-path cache. Every invocation re-walks `$PATH` and re-runs `stat()` on every candidate.
Why: On slow filesystems (NFS home, encrypted volume, long `$PATH`) adds tens to hundreds of ms per command; doubles long-pipeline runtime.
Fix suggestion: Keep the option on. Use `rehash` (one-shot) or `hash -r` to invalidate after a targeted binary swap.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1930_test.go
+++ b/pkg/katas/katatests/zc1930_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #952, opened 2026-04-19
feat: ZC1931 — detect `ip netns delete` namespace teardown

afadesigns:
ZC1931 — Warn on `ip netns delete $NS` / `ip netns del $NS`

What: Unmounts the namespace and tears down every interface, veth pair, VXLAN, and WireGuard peer living inside.
Why: Processes still attached lose network abruptly — container health checks fail, BGP sessions drop, `ip netns exec $NS` errors with 'No such file or directory'.
Fix suggestion: Stop the workloads first, verify `ip -n $NS link` is empty, then delete. Leave Docker/containerd/nspawn-managed namespaces alone.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1931_test.go
+++ b/pkg/katas/katatests/zc1931_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #953, opened 2026-04-19
feat: ZC1932 — detect `unsetopt GLOBAL_EXPORT` function-local typeset -x

afadesigns:
ZC1932 — Warn on `unsetopt GLOBAL_EXPORT`

What: Flips `typeset -x VAR=val` inside a function from "export AND promote to outer scope" to function-local.
Why: Helper functions that set `PATH`, `VIRTUAL_ENV`, `AWS_*` no longer propagate — subsequent commands run under the old environment.
Fix suggestion: Keep the option on. Scope temporary exports with a subshell rather than a shell-wide flip.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1932_test.go
+++ b/pkg/katas/katatests/zc1932_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #954, opened 2026-04-19
feat: ZC1933 — detect `ipvsadm -C` wiping IPVS load-balancer table

afadesigns:
ZC1933 — Error on `ipvsadm -C` / `--clear`

What: Removes every virtual service, real server, and connection entry from the in-kernel IPVS table.
Why: Traffic that was being load-balanced to a backend farm falls through to local listeners (or drops); keepalived / ldirectord states invert; clients see 5xx until config is replayed.
Fix suggestion: Save first (`ipvsadm --save -n > /run/ipvs.bak`), drain specific services with `ipvsadm -D`, keep `--clear` in break-glass runbooks.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1933_test.go
+++ b/pkg/katas/katatests/zc1933_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #955, opened 2026-04-19
feat: ZC1934 — detect `setopt AUTO_NAME_DIRS` implicit named dir

afadesigns:
ZC1934 — Warn on `setopt AUTO_NAME_DIRS`

What: Auto-registers any parameter whose value is an absolute directory path as a named directory (`~name`).
Why: `foo=/srv/data` silently makes `~foo` expand to `/srv/data` in later commands and in `%~` prompt sequences — names the user never opted into.
Fix suggestion: Keep the option off. Call `hash -d name=/path` explicitly when a named directory is actually wanted.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1934_test.go
+++ b/pkg/katas/katatests/zc1934_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #956, opened 2026-04-19
feat: ZC1935 — detect `apt autoremove --purge` stripping deps + config

afadesigns:
ZC1935 — Warn on `apt autoremove --purge` / `zypper rm --clean-deps`

What: Removes every package the resolver thinks is no longer required, plus (with `--purge`) their `/etc` config and data dirs.
Why: In CI, quietly uproots packages someone installed manually but never `apt-mark manual`-ed. `--purge` makes the removal irreversible.
Fix suggestion: Run `apt autoremove --dry-run` in review, mark the keepers with `apt-mark manual`, drop `--purge` from unattended jobs.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1935_test.go
+++ b/pkg/katas/katatests/zc1935_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #957, opened 2026-04-19
feat: ZC1936 — detect `setopt POSIX_ALIASES` narrowing alias scope

afadesigns:
ZC1936 — Warn on `setopt POSIX_ALIASES`

What: Narrows alias expansion to plain identifiers — aliases on reserved words (`if`, `for`, `function`, `while`, `do`, `[`) stop firing.
Why: Libraries that hooked reserved words (oh-my-zsh-style plugin preexec wrappers) silently stop being picked up.
Fix suggestion: Keep the option off. Scope POSIX-specific blocks with `emulate -LR sh` instead of flipping globally.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1936_test.go
+++ b/pkg/katas/katatests/zc1936_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #958, opened 2026-04-19
feat: ZC1937 — detect `tmux kill-server`/`screen -X quit` session teardown

afadesigns:
ZC1937 — Warn on `tmux kill-server` / `tmux kill-session` / `screen -X quit`

What: Terminates the whole multiplexer daemon or a named session.
Why: Builds, `tail -F` on production logs, held `sudo` tokens, port-forwards parked inside die with no cleanup — detached processes get `SIGHUP`'d.
Fix suggestion: Use `tmux kill-window -t …` for surgical removal, send `SIGTERM` to the specific backend, or use `systemd-run --scope` for workloads that should survive terminal churn.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1937_test.go
+++ b/pkg/katas/katatests/zc1937_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #959, opened 2026-04-19
feat: ZC1938 — detect `setopt POSIX_JOBS` job-table scope flip

afadesigns:
ZC1938 — Warn on `setopt POSIX_JOBS`

What: Makes job control POSIX-compliant — `%1` / `%n` refer only to jobs of the current shell, forked subshells get their own job table.
Why: Scripts that launched a background job in the parent and then `wait %1` from inside a subshell fail with 'no such job'. `fg`/`bg`/`disown` scope-differently too.
Fix suggestion: Leave off in Zsh. If POSIX job semantics are required, scope with `emulate -LR sh` inside one function.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1938_test.go
+++ b/pkg/katas/katatests/zc1938_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #960, opened 2026-04-19
feat: ZC1939 — detect `reboot -f`/`halt -f`/`poweroff -f` force-shutdown

afadesigns:
ZC1939 — Error on `reboot -f` / `halt -f` / `poweroff -f`

What: Short-circuits the systemd shutdown graph — no `ExecStop=`, no filesystem sync, no clean unmount.
Why: `reboot(2)` fires immediately; dirty buffers lost, journal writes stop mid-line, databases replay from last checkpoint, LUKS/NFS/cephfs log a dirty state.
Fix suggestion: Use `systemctl reboot` / `shutdown -r +N`. Reserve `-f` for recovery when the normal path is already wedged.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1939_test.go
+++ b/pkg/katas/katatests/zc1939_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #961, opened 2026-04-19
feat: ZC1940 — detect `setopt POSIX_ARGZERO` freezing $0 in functions

afadesigns:
ZC1940 — Warn on `setopt POSIX_ARGZERO`

What: `$0` no longer changes to the current function name; retains the outer script name (or interpreter when sourced).
Why: Loggers like `log() { printf '%s: %s\n' "$0" "$*"; }` print the script path for every call — call-site context lost. `case $0` dispatch inside auto-loaded functions breaks.
Fix suggestion: Leave off. If POSIX `$0` is required, scope with `emulate -LR sh` inside the function.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1940_test.go
+++ b/pkg/katas/katatests/zc1940_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #962, opened 2026-04-19
feat: ZC1941 — detect `restic --insecure-no-password` unencrypted repo

afadesigns:
ZC1941 — Error on `restic init --insecure-no-password`

What: Creates a restic repo whose data chunks are reachable without a key.
Why: Every `backup` / `restore` round-trips plaintext blocks — any operator with read access to the bucket / NFS share / SFTP directory can reassemble the backed-up filesystem (shell history, SSH keys, database dumps).
Fix suggestion: Pass a passphrase via `--password-file` (mode 0400, readable only by the backup user) or `--password-command`. Never use `--insecure-*` outside a local test repo.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1941_test.go
+++ b/pkg/katas/katatests/zc1941_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #963, opened 2026-04-19
feat: ZC1942 — detect `setopt CLOBBER_EMPTY` weakening NO_CLOBBER

afadesigns:
ZC1942 — Warn on `setopt CLOBBER_EMPTY`

What: Relaxes `NO_CLOBBER` so `>file` still overwrites zero-length targets.
Why: Scripts using `setopt NO_CLOBBER` as a safety net lose it for every freshly-`touch`ed lock file, sentinel, or `install -D`-created placeholder — next stray `>sentinel` silently overwrites.
Fix suggestion: Keep the option off. Use explicit `>|file` to bypass `NO_CLOBBER` for a specific write.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1942_test.go
+++ b/pkg/katas/katatests/zc1942_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #964, opened 2026-04-19
feat: ZC1943 — detect `systemd-nspawn -b` booting untrusted rootfs

afadesigns:
ZC1943 — Warn on `systemd-nspawn -b` / `--boot`

What: Launches the rootfs's `/sbin/init` inside a minimally-isolated namespace.
Why: Default caps inherit `CAP_AUDIT_CONTROL`, `CAP_NET_ADMIN`, r/w access to host `/dev` matching the cgroup. If `$ROOT` is operator-supplied, its init scripts run first and can probe the host.
Fix suggestion: Use `-U` for user-namespace isolation, drop caps with `--capability=`, pair with `--private-network`. Prefer `machinectl start` on a reviewed image.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1943_test.go
+++ b/pkg/katas/katatests/zc1943_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #965, opened 2026-04-19
feat: ZC1944 — detect `setopt IGNORE_EOF` masking runaway subshells

afadesigns:
ZC1944 — Warn on `setopt IGNORE_EOF`

What: Interactive shell treats EOF on stdin as nothing — `Ctrl-D` stops terminating a login.
Why: In unattended `zsh -i -c` launches or sourced rc files, subshells, sudo holds, SSH tunnels, port-forwards, and build supervisors linger long after the parent left.
Fix suggestion: Keep off. If a stale-tty guard is wanted, set `TMOUT=NN` for a timed exit.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1944_test.go
+++ b/pkg/katas/katatests/zc1944_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #966, opened 2026-04-19
feat: ZC1945 — detect `bpftrace -e`/`bpftool prog load` in-kernel eBPF

afadesigns:
ZC1945 — Warn on `bpftrace -e` / `bpftool prog load`

What: Compiles or installs an eBPF program that attaches to kprobes, tracepoints, or uprobes.
Why: Needs `CAP_BPF`/`CAP_SYS_ADMIN` and can read arbitrary kernel/userland memory — every syscall arg, every TCP payload, every command a sibling process runs.
Fix suggestion: Pin loaded programs to a directory the operator owns, gate invocation behind a runbook, prefer a short-lived `bpftrace -c CMD` window.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1945_test.go
+++ b/pkg/katas/katatests/zc1945_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #967, opened 2026-04-19
feat: ZC1946 — detect `unsetopt HUP` leaving background jobs past exit

afadesigns:
ZC1946 — Warn on `unsetopt HUP`

What: Stops the shell from sending `SIGHUP` to running child jobs on exit (default ON in Zsh).
Why: Long pipelines, `sleep` loops, and spawned daemons outlive the session — `ps aux` accumulates orphans across logouts and resource use creeps up.
Fix suggestion: Keep `HUP` on globally. Use `disown` or `systemd-run --scope` on specific commands that need to outlive the shell.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1946_test.go
+++ b/pkg/katas/katatests/zc1946_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #968, opened 2026-04-19
feat: ZC1947 — detect `ip xfrm state/policy flush` IPsec teardown

afadesigns:
ZC1947 — Error on `ip xfrm state flush` / `ip xfrm policy flush`

What: Removes every IPsec Security Association and every policy driving them.
Why: Strongswan/libreswan/FRR/WireGuard-over-xfrm tunnels drop instantly; kernel packet paths stop encrypting; peers renegotiate from scratch (plaintext may leak during the gap on misconfigured hosts).
Fix suggestion: Use `ip xfrm state deleteall src $A dst $B` to scope per-tunnel, pair flushes with a maintenance window.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1947_test.go
+++ b/pkg/katas/katatests/zc1947_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #969, opened 2026-04-19
feat: ZC1948 — detect `ipmitool -P PASS` BMC password in argv

afadesigns:
ZC1948 — Error on `ipmitool -P PASS`

What: BMC credential passed as a command-line argument.
Why: Lands in `ps`, `/proc/PID/cmdline`, crash dumps. The BMC is a root-equivalent out-of-band controller (power, console, firmware update) — one of the most sensitive tokens on the host.
Fix suggestion: Use `-f <password_file>` (mode `0400`, owned by the automation user) or `IPMI_PASSWORD=… ipmitool -E`.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1948_test.go
+++ b/pkg/katas/katatests/zc1948_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #970, opened 2026-04-19
feat: ZC1949 — detect `rmmod -f` bypassing module refcount

afadesigns:
ZC1949 — Error on `rmmod -f` / `rmmod --force`

What: Tears down a kernel module even when its reference count is non-zero.
Why: Live `open("/dev/…")`, mounted filesystems, or in-flight network drivers become dangling pointers — kernel oopses or panics on the next callback. Compiled out on most distros; strictly break-glass when present.
Fix suggestion: Stop holders first (`lsof /dev/FOO`, `umount`, `ip link set dev … down`), then plain `rmmod` or `modprobe -r`.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1949_test.go
+++ b/pkg/katas/katatests/zc1949_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #971, opened 2026-04-19
feat: ZC1950 — detect `tune2fs -O ^has_journal`/`-m 0` resilience strip

afadesigns:
ZC1950 — Error on `tune2fs -O ^has_journal` / `tune2fs -m 0`

What: Strips the ext3/4 journal or zeroes the root-reserved space.
Why: Crash recovery drops from journal-replay to full `fsck -y` (truncates partial files). Zero reserve leaves no headroom for `journald`/`apt`/root shells when the fs fills up — recovery via rescue media.
Fix suggestion: Keep the journal on. Leave `-m` at the distro default (5% is overkill on huge disks; `-m 1` is still safe).
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1950_test.go
+++ b/pkg/katas/katatests/zc1950_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #972, opened 2026-04-20
feat: ZC1951 — detect `ceph --yes-i-really-really-mean-it` safety bypass

afadesigns:
ZC1951 — Error on `ceph … --yes-i-really-really-mean-it` / `--yes-i-really-mean-it`

What: Bakes Ceph's intentional double-confirm phrase into a script.
Why: Ceph requires the phrase precisely so an operator-typo during a live op cannot drop production data. Automating it defeats the friction — stale `for pool in $(…)` loops, rebase mishaps, or wrong-variable expansions silently delete real pools.
Fix suggestion: Remove the flag from scripts. Do deletions interactively or wrap in a runbook that spells the pool name in a commit the operator acknowledges.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1951_test.go
+++ b/pkg/katas/katatests/zc1951_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #973, opened 2026-04-20
feat: ZC1952 — detect `zfs set sync=disabled` fsync no-op

afadesigns:
ZC1952 — Error on `zfs set sync=disabled POOL/DATASET`

What: Turns `fsync()`, `O_SYNC`, and `O_DSYNC` into no-ops on the dataset.
Why: PostgreSQL, MariaDB, etcd, and every fsync-relying app report success for writes still in ARC — panic or power cut loses minutes of committed transactions.
Fix suggestion: Leave sync at `standard`. If latency is the concern, add a `log` vdev (SLOG) or tune `zfs_txg_timeout`.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1952_test.go
+++ b/pkg/katas/katatests/zc1952_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #974, opened 2026-04-20
feat: ZC1953 — detect `mount --make-shared` propagation escape vector

afadesigns:
ZC1953 — Warn on `mount --make-shared` / `mount --make-rshared`

What: Turns the mount point into a peer in a shared-subtree group.
Why: Later bind-mounts propagate to every peer, including containers and other namespaces. Combined with `CAP_SYS_ADMIN` in a pod, this is a classic container-escape stepping stone.
Fix suggestion: Use `--make-private` on sensitive paths; mount containers with `--mount-propagation=private`/`slave` unless the app needs `shared`.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1953_test.go
+++ b/pkg/katas/katatests/zc1953_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #976, opened 2026-04-20
feat: ZC1955 — detect `rfkill block all` radio mass-down

afadesigns:
ZC1955 — Warn on `rfkill block all` / `block wifi|bluetooth|wwan`

What: Soft-kills every radio the kernel registered — WiFi, Bluetooth, WWAN, NFC, GPS, UWB — in one call.
Why: Host drops off the network. `unblock all` takes seconds-to-minutes on some drivers and requires the operator to be physically present or have cellular fallback.
Fix suggestion: Scope to a specific radio type. Schedule via `at now + N minutes … rfkill unblock all` so the host recovers on its own.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1955_test.go
+++ b/pkg/katas/katatests/zc1955_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #977, opened 2026-04-20
feat: ZC1956 — detect `tailscale --auth-key=SECRET` in argv

afadesigns:
ZC1956 — Error on `tailscale up --auth-key=tskey-…` / `--authkey tskey-…`

What: Tailscale pre-auth key passed as a command-line argument.
Why: Grants full tailnet membership. Value ends up in `ps`, `/proc/PID/cmdline`, shell history, and process dumps taken before the join completes.
Fix suggestion: Use `--auth-key=file:/etc/ts.key` (mode `0400`) or `--authkey-env=TS_AUTHKEY` with an env-var source.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1956_test.go
+++ b/pkg/katas/katatests/zc1956_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #978, opened 2026-04-20
feat: ZC1957 — detect `lvchange -an`/`vgchange -an` live-volume deactivation

afadesigns:
ZC1957 — Warn on `lvchange -an VG/LV` / `vgchange -an VG`

What: Deactivates a logical volume (or entire VG) by removing its device-mapper entry.
Why: If mounted, kernel-buffered writes may be lost and processes with open fds get EIO on next syscall.
Fix suggestion: `umount` and stop holders first, verify with `lsof`/`fuser`, then `lvchange -an`. For teardown, prefer `umount` + `lvremove` with a recovery snapshot.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1957_test.go
+++ b/pkg/katas/katatests/zc1957_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #979, opened 2026-04-20
feat: ZC1958 — detect `helm upgrade --force` delete+create strategy

afadesigns:
ZC1958 — Warn on `helm upgrade --force`

What: Flips the upgrade strategy from three-way merge to delete+create for every resource Helm owns.
Why: Deployments become new objects, Services lose `clusterIP` for a beat, `PodDisruptionBudget` is bypassed because the resource is deleted not rolled out.
Fix suggestion: Use plain `helm upgrade` (three-way merge) or `--atomic`/`--wait` for supervised rolls. Reserve `--force` for recovery after a failed upgrade with a stuck resource.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1958_test.go
+++ b/pkg/katas/katatests/zc1958_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #980, opened 2026-04-20
feat: ZC1959 — detect `trivy --skip-db-update` stale CVE DB

afadesigns:
ZC1959 — Warn on `trivy --skip-db-update` / `--skip-update`

What: Disables the per-scan refresh of trivy's vulnerability database.
Why: Each scan then misses every CVE disclosed since the cached DB was last refreshed — stale CI gate lets known vulns through.
Fix suggestion: Keep the default download. Pre-populate via scheduled `trivy --download-db-only`, and only skip inside that same job so every subsequent scan sees fresh data.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1959_test.go
+++ b/pkg/katas/katatests/zc1959_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #981, opened 2026-04-20
feat: ZC1960 — detect cloud-plane RCE: `az vm run-command` / `aws ssm send-command`

afadesigns:
ZC1960 — Warn on `az vm run-command invoke` / `aws ssm send-command` / `gcloud compute ssh --command`

What: Runs arbitrary shell on a remote VM via the cloud control plane's IAM credentials.
Why: The identity making the call is whatever role the script carries; if the command string is composed from operator/attacker input, the result is RCE on the target instance through IAM.
Fix suggestion: Pin the document/script to a reviewed asset in blob/S3, template-escape any injected parameters, and require MFA on the invoking role.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1960_test.go
+++ b/pkg/katas/katatests/zc1960_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #982, opened 2026-04-20
feat: ZC1961 — detect `gcloud iam service-accounts keys create` static key

afadesigns:
ZC1961 — Warn on `gcloud iam service-accounts keys create key.json --iam-account=…`

What: Exports an RSA key pair as a JSON bearer credential.
Why: Forever-valid — no auto-rotation, no refresh. A single `cat key.json` leaks a full service-account identity with no revocation path short of manual key delete.
Fix suggestion: Use Workload Identity Federation, `--impersonate-service-account` for short-lived tokens, or the key-less GCE/GKE attached SA. Reserve JSON keys for provably off-platform callers.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1961_test.go
+++ b/pkg/katas/katatests/zc1961_test.go
@@ -1,0 +1,51 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #983, opened 2026-04-20
feat: ZC1962 — detect `kustomize build --load-restrictor=None`

afadesigns:
ZC1962 — Warn on `kustomize build … --load-restrictor=LoadRestrictionsNone` / `--load_restrictor none`

What: Drops kustomize's default path restriction so overlays can reference files outside the kustomization root.
Why: Untrusted remote bases or overlays can then load `../../secrets/prod.env`, absolute paths, or arbitrary host files — secrets leak into the render.
Fix suggestion: Keep the default `LoadRestrictionsRootOnly`. If an overlay legitimately needs a sibling file, vendor it into the overlay directory.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1962_test.go
+++ b/pkg/katas/katatests/zc1962_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #984, opened 2026-04-20
feat: ZC1963 — detect unpinned `npx`/`pnpm dlx`/`bunx` runners

afadesigns:
ZC1963 — Warn on `npx PKG` / `pnpm dlx PKG` / `bunx PKG` / `bun x PKG` without version pin

What: Fetches the named package from the npm registry and runs its `bin` entry using the `latest` tag.
Why: A compromised maintainer, squatted name, or mistyped package is enough to land attacker code in the build.
Fix suggestion: Pin the version (`pkg@1.2.3`), or `npm install` the package and invoke `./node_modules/.bin/…` with a lockfile. Verify tarball signature for high-value packages.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1963_test.go
+++ b/pkg/katas/katatests/zc1963_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #985, opened 2026-04-20
feat: ZC1964 — detect unpinned `uvx`/`uv tool run`/`pipx run` runners

afadesigns:
ZC1964 — Warn on `uvx PKG` / `uv tool run PKG` / `pipx run PKG` without version pin

What: Resolves the package against PyPI and executes its entry point; without `==X.Y.Z` or `@X.Y.Z` each run takes whatever the registry currently serves.
Why: Squatted typo names (`reqeusts`, `djanggo`), compromised maintainer releases, or sudden major bumps land untested code in the pipeline.
Fix suggestion: Pin `pkg==X.Y.Z` / `pkg@X.Y.Z` at the call site, or use `uv tool install pkg==X.Y.Z` once so the lockfile is authoritative.
Severity: Warning

### DIFF
--- a/pkg/katas/katatests/zc1964_test.go
+++ b/pkg/katas/katatests/zc1964_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #986, opened 2026-04-20
feat: ZC1965 — detect `systemd-cryptenroll --wipe-slot=all` LUKS keyslot wipe

afadesigns:
ZC1965 — Error on `systemd-cryptenroll --wipe-slot=all $DEV`

What: Removes every LUKS key slot — passphrase, recovery key, TPM2, FIDO2, PKCS#11 — in one call.
Why: No confirmation; one-shot brick. `--wipe-slot=recovery` / `--wipe-slot=empty` are scoped; `all` is not.
Fix suggestion: Enrol the new slot first, wipe the specific retired index (`--wipe-slot=<n>`). Back up the header with `cryptsetup luksHeaderBackup` before the call.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1965_test.go
+++ b/pkg/katas/katatests/zc1965_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #987, opened 2026-04-20
feat: ZC1966 — detect `zpool import -f`/`export -f` forced pool op

afadesigns:
ZC1966 — Error on `zpool import -f` / `zpool export -f` — forced ZFS pool op bypasses hostid/txg checks

What: `zpool import -f $POOL` or `zpool export -f $POOL` forcing past safety checks.
Why: `-f` on import bypasses the hostid guard — on SAN/multipath/HA a pool already imported elsewhere gets a second writer and silently corrupts. `-f` on export skips graceful txg flush and can drop the tail of the ZIL.
Fix suggestion: `zfs unmount -a` then plain `zpool export`/`import`; verify with `zpool import` (no target) first.
Severity: Error

### DIFF
--- a/pkg/katas/katatests/zc1966_test.go
+++ b/pkg/katas/katatests/zc1966_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #988, opened 2026-04-20
feat: ZC1967 — detect `setopt PROMPT_SUBST` prompt-substitution footgun

afadesigns:
ZC1967 — Warn on `setopt PROMPT_SUBST` — expansions inside `$PROMPT` evaluate command substitution every redraw

What: Script flips `setopt PROMPT_SUBST` (or `unsetopt NO_PROMPT_SUBST`) globally.
Why: With the option on, parameter/command/arith expansion runs on every prompt redraw against values from untrusted sources — git branch, directory name, hostname, env var set by a spawned tool. A branch called `$(id>/tmp/p)` executes on each render.
Fix suggestion: Use Zsh prompt escapes (`%n`/`%d`/`%~`/`%m`/`vcs_info`) which sanitise their inputs, or scope with `setopt LOCAL_OPTIONS PROMPT_SUBST` inside the prompt-building function.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1967` passes
- [x] `scripts/update-version.sh` → 0.9.64

### DIFF
--- a/pkg/katas/katatests/zc1967_test.go
+++ b/pkg/katas/katatests/zc1967_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1967(t *testing.T) {
+	tests := []struct {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #989, opened 2026-04-20
fix: ZC1964 description drops spell-check tripwire words

afadesigns:
Spelling check on main flagged `reqeusts` in the ZC1964 description (intentional typosquat example). Rephrase without the literal misspelled tokens so the check passes cleanly; behavior unchanged.


redteamx (approved):
LGTM.

PR #990, opened 2026-04-20
feat: ZC1968 — detect `dnf/yum versionlock add` RPM pin blocking CVE updates

afadesigns:
ZC1968 — Warn on `dnf versionlock add` / `yum versionlock add` — pins RPM, blocks CVE updates

What: Script calls `dnf versionlock add pkg` / `yum versionlock add pkg` / `microdnf versionlock add pkg` (or the `exclude` subform).
Why: Writes a persistent entry to `/etc/dnf/plugins/versionlock.list` that hides the package from future `dnf update`. Unattended-upgrades never sees the new rpm — kernel/openssl/glibc CVEs accumulate silently. RPM-side parallel of ZC1550 (`apt-mark hold`).
Fix suggestion: Prefer a one-shot `--exclude` on the single transaction (`dnf update --exclude=pkg`), document the reason, schedule a `dnf versionlock delete` review.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1968` passes
- [x] `scripts/update-version.sh` → 0.9.65

### DIFF
--- a/pkg/katas/katatests/zc1968_test.go
+++ b/pkg/katas/katatests/zc1968_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1968(t *testing.T) {
+	tests := []struct {
+		name     string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #991, opened 2026-04-20
feat: ZC1969 — detect `zsh -f`/`zsh -d` skipping startup-file hardening

afadesigns:
ZC1969 — Warn on `zsh -f` / `zsh -d` — skips `/etc/zsh*` and `~/.zsh*` startup files

What: Script invokes `zsh -f` (`--no-rcs`) or `zsh -d` (`--no-globalrcs`).
Why: Either form strips corp-mandated settings that live in zsh startup files — proxy/hosts overrides, audit hooks, umask, `HISTFILE` redirection, `PATH` hardening — and the launched shell runs without them.
Fix suggestion: Use `-f`/`-d` only for a pristine test harness or a minimal repro. When isolation is required, prefer `env -i zsh` with an explicit allow-list of variables.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1969` passes
- [x] `scripts/update-version.sh` → 0.9.66

### DIFF
--- a/pkg/katas/katatests/zc1969_test.go
+++ b/pkg/katas/katatests/zc1969_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #992, opened 2026-04-20
feat: ZC1970 — detect kernel partscan on untrusted disk image (`losetup -P`/`kpartx -a`/`partprobe`)

afadesigns:
ZC1970 — Warn on `losetup -P` / `kpartx -a` / `partprobe` on untrusted image — runs kernel partition parser

What: Script calls `losetup -P`, `kpartx -a[v]`, or `partprobe` against a disk image.
Why: Each asks the kernel to rescan the partition table and publish `/dev/loopNpX`/`dm-N` nodes. The MBR/GPT/LVM parsers run over attacker-controlled bytes. Historically that path has tripped kernel CVEs (fsconfig heap overflow, ext4 mount bugs).
Fix suggestion: Inspect layouts offline with `fdisk -l $IMG`/`sfdisk --dump $IMG`, which never ask the kernel to scan. Only attach partitions with `losetup -P` once the image is known-good; isolate the attach in a throwaway VM for customer-supplied media.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1970` passes
- [x] `scripts/update-version.sh` → 0.9.67

### DIFF
--- a/pkg/katas/katatests/zc1970_test.go
+++ b/pkg/katas/katatests/zc1970_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1970(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #993, opened 2026-04-20
feat: ZC1971 — detect `unsetopt GLOBAL_RCS` bypassing system Zsh rc files

afadesigns:
ZC1971 — Warn on `unsetopt GLOBAL_RCS` / `setopt NO_GLOBAL_RCS` — skips `/etc/zprofile`, `/etc/zshrc`, `/etc/zlogin`, `/etc/zlogout`

What: Script flips `GLOBAL_RCS` off (either `unsetopt GLOBAL_RCS` or `setopt NO_GLOBAL_RCS`).
Why: `GLOBAL_RCS` is on by default and the system `/etc/zsh*` files are where admins park corp `PATH` hardening, audit hooks, umask, `HISTFILE` redirection, proxy vars. Disabling the option — typically from `/etc/zshenv`, the one file that runs before it can be toggled — neutralises every downstream system rc silently.
Fix suggestion: Keep the option on. For helpers that need a pristine shell, scope with `emulate -LR zsh` inside a function, or spawn `env -i zsh -f` for just that child.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1971` passes
- [x] `scripts/update-version.sh` → 0.9.68

### DIFF
--- a/pkg/katas/katatests/zc1971_test.go
+++ b/pkg/katas/katatests/zc1971_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1971(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected []katas.Violation
+	}{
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #994, opened 2026-04-20
feat: ZC1972 — detect `dmsetup remove_all`/`remove -f` tearing down live DM mappings

afadesigns:
ZC1972 — Error on `dmsetup remove_all` / `dmsetup remove -f` — tears down live LVM/LUKS/multipath mappings

What: Script calls `dmsetup remove_all` or `dmsetup remove -f $NAME`.
Why: `remove_all` iterates every DM node on the host — LVM LVs, LUKS containers, multipath aggregates, `cryptsetup` mappings — and forces each out. `remove --force` does the same for a single mapping. When any of those devices is mounted or backing a VM, in-flight I/O returns `ENXIO`, `fsck` becomes impossible, and LVM metadata needs a cold reboot to reappear.
Fix suggestion: `umount` → `vgchange -an` → `cryptsetup close` in order, then `dmsetup remove $NAME` without force. Never `remove_all` on production.
Severity: Error

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1972` passes
- [x] `scripts/update-version.sh` → 0.9.69

### DIFF
--- a/pkg/katas/katatests/zc1972_test.go
+++ b/pkg/katas/katatests/zc1972_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1972(t *testing.T) {
+	tests := []struct {
+		name     string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #995, opened 2026-04-20
feat: ZC1973 — detect `setopt POSIX_IDENTIFIERS` breaking Unicode parameter names

afadesigns:
ZC1973 — Warn on `setopt POSIX_IDENTIFIERS` — restricts parameter names to ASCII, breaks Unicode `\$var`

What: Script flips `POSIX_IDENTIFIERS` on (either `setopt POSIX_IDENTIFIERS` or `unsetopt NO_POSIX_IDENTIFIERS`).
Why: Default Zsh accepts Unicode parameter names (`\$café`, `\$π`, `\$данные`). With the option on, the parser rejects anything outside POSIX (ASCII alnum + `_`, not digit-leading), so i18n-named vars in libraries stop loading and later assignments fail.
Fix suggestion: Leave the option off globally. When POSIX identifiers are genuinely needed for a helper, scope with `emulate -LR sh` inside that function rather than flipping the global setting.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1973` passes
- [x] `scripts/update-version.sh` → 0.9.70

### DIFF
--- a/pkg/katas/katatests/zc1973_test.go
+++ b/pkg/katas/katatests/zc1973_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1973(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #996, opened 2026-04-20
feat: ZC1974 — detect `ipset flush`/`ipset destroy` wiping firewall-referenced sets

afadesigns:
ZC1974 — Error on `ipset flush` / `ipset destroy` — nukes named sets referenced by iptables/nft rules

What: Script calls `ipset flush` (empties every entry from a named set) or `ipset destroy` with no args (removes every set).
Why: iptables/nft rules of the form `-m set --match-set $NAME src` then point at an empty or missing set. Block-lists stop blocking; allow-lists stop whitelisting. The policy silently falls through to the default (ACCEPT/DROP) and the firewall's intent is gone.
Fix suggestion: Target by name (`ipset destroy $NAME` after confirming no rule references it), or reload atomically with `ipset restore -! < snapshot` when a full replace is genuinely needed.
Severity: Error

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1974` passes
- [x] `scripts/update-version.sh` → 0.9.71

### DIFF
--- a/pkg/katas/katatests/zc1974_test.go
+++ b/pkg/katas/katatests/zc1974_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1974(t *testing.T) {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #997, opened 2026-04-20
feat: ZC1975 — detect `unsetopt EXEC` turning the script into a silent no-op

afadesigns:
ZC1975 — Warn on `unsetopt EXEC` / `setopt NO_EXEC` — parser keeps scanning, commands stop running

What: Script flips `EXEC` off (either `unsetopt EXEC` or `setopt NO_EXEC`).
Why: `EXEC` on = parse + run; off = parse only. Once disabled in-line, every later command is silently skipped, parameter assignments on the same line don't take effect, and `\$?` stays frozen. Same semantics as `zsh -n script.zsh`, except the script thinks it is running.
Fix suggestion: Run syntax checks from outside the script (`zsh -n script.zsh`). Never flip `EXEC` in the script body.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1975` passes
- [x] `scripts/update-version.sh` → 0.9.72

### DIFF
--- a/pkg/katas/katatests/zc1975_test.go
+++ b/pkg/katas/katatests/zc1975_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #998, opened 2026-04-20
feat: ZC1976 — detect `exportfs -au`/`-u` unexport that ESTALEs live NFS mounts

afadesigns:
ZC1976 — Error on `exportfs -au` / `exportfs -u` — unexports live NFS shares, clients get `ESTALE`

What: Script calls `exportfs -au` (unexport all) or `exportfs -u HOST:/PATH` (unexport one).
Why: Clients that already have the export mounted are not notified. The next read/write returns `ESTALE`, every open fd fails, and recovery requires a client-side `umount -l` + remount.
Fix suggestion: After editing `/etc/exports`, use `exportfs -f` (flush cache) — that's almost always what's actually wanted. Reserve `-u`/`-au` for a planned shutdown with a coordinated client `umount` first.
Severity: Error

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1976` passes
- [x] `golangci-lint run ./...` clean (added `exportfs` to misspell ignore-words)
- [x] `scripts/update-version.sh` → 0.9.73

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,6 +20,9 @@ linters:
     - misspell
 
 linters-settings:
+  misspell:
+    ignore-words:
+      - exportfs
   lll:
     line-length: 250
   funlen:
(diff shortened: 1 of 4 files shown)


redteamx (approved):
LGTM.

PR #999, opened 2026-04-20
feat: ZC1977 — detect `setopt CHASE_DOTS` breaking blue/green `current` symlinks

afadesigns:
ZC1977 — Warn on `setopt CHASE_DOTS` — `cd ..` physically resolves before walking up, breaking logical paths

What: Script flips `CHASE_DOTS` on (`setopt CHASE_DOTS` or `unsetopt NO_CHASE_DOTS`).
Why: Default behaviour keeps `..` logical: from `/app/current/lib` (where `/app/current` is a symlink), `cd ..` returns to `/app/current`. With the option on, `..` first resolves the current directory to its physical inode, so `cd ..` lands in the real release dir and `cd ../foo` looks for `foo` under that release — blue/green deployment symlinks break silently.
Fix suggestion: Leave the option off. Request one-shot physical resolution with `cd -P` when a specific call really needs it.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1977` passes
- [x] `scripts/update-version.sh` → 0.9.74

### DIFF
--- a/pkg/katas/katatests/zc1977_test.go
+++ b/pkg/katas/katatests/zc1977_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1977(t *testing.T) {
+	tests := []struct {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1000, opened 2026-04-20
feat: ZC1978 — detect `ftp`/`tftp` plaintext file transfer

afadesigns:
ZC1978 — Warn on `ftp` / `tftp` — cleartext transfer, credentials and payload exposed on the wire

What: Script calls `ftp HOST…` or `tftp HOST…`.
Why: `ftp` negotiates `USER`/`PASS` and moves payloads in plaintext TCP/21; anything between caller and server sees both. `tftp` has no auth at all and runs in the clear over UDP/69. Dual-channel design also trips modern NAT/firewall gear.
Fix suggestion: Use `sftp` / `scp` / `rsync -e ssh` for authenticated transfers. For PXE-style provisioning that historically used `tftp`, fetch a signed payload over HTTPS with `curl`.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1978` passes
- [x] `scripts/update-version.sh` → 0.9.75

### DIFF
--- a/pkg/katas/katatests/zc1978_test.go
+++ b/pkg/katas/katatests/zc1978_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1001, opened 2026-04-20
feat: ZC1979 — detect `setopt HIST_FCNTL_LOCK` NFS lockd deadlock

afadesigns:
ZC1979 — Warn on `setopt HIST_FCNTL_LOCK` — `fcntl()` lock on NFS `\$HISTFILE` stalls or deadlocks

What: Script flips `HIST_FCNTL_LOCK` on (`setopt HIST_FCNTL_LOCK` or `unsetopt NO_HIST_FCNTL_LOCK`).
Why: Off by default, Zsh uses its own lock-file dance next to `\$HISTFILE`. With the option on, it switches to POSIX `fcntl()` advisory locking — safer locally, but on NFS homes the lock is proxied through `rpc.lockd`. A single hung client or rebooted NFS server leaves every other shell blocked the next time it writes history; interactive shells appear frozen, and scripts that source user rc files hang in `zshaddhistory`.
Fix suggestion: Keep off on NFS homes. Enable only when `\$HISTFILE` lives on a local fs (ext4/xfs/btrfs/zfs-local) that implements `fcntl()` without network round-trips.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1979` passes
- [x] `scripts/update-version.sh` → 0.9.76

### DIFF
--- a/pkg/katas/katatests/zc1979_test.go
+++ b/pkg/katas/katatests/zc1979_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1979(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected []katas.Violation
+	}{
+		{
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1002, opened 2026-04-20
feat: ZC1980 — detect `udevadm trigger --action=remove` live-device detach

afadesigns:
ZC1980 — Error on `udevadm trigger --action=remove` — replays `remove` uevents, detaches live devices

What: Script calls `udevadm trigger --action=remove` (or `-c remove`).
Why: `udevadm trigger` walks `/sys` and synthesises the requested uevent for every matching device. With `--action=remove` the kernel reacts as if every matched disk, NIC, GPU, or USB node was physically yanked — SATA controllers detach drives backing mounted filesystems, netdevs disappear mid-session, and `systemd-udevd` fires per-device cleanup rules never meant for a live host.
Fix suggestion: To replay rules after a config edit, use `udevadm control --reload` then `udevadm trigger` with the default `change` action. Scope any genuine `--action=remove` with `--subsystem-match=` / `--attr-match=` and test on a non-production box first.
Severity: Error

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1980` passes
- [x] `scripts/update-version.sh` → 0.9.77

### DIFF
--- a/pkg/katas/katatests/zc1980_test.go
+++ b/pkg/katas/katatests/zc1980_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1980(t *testing.T) {
+	tests := []struct {
+		name     string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1003, opened 2026-04-20
feat: ZC1981 — detect `exec -a NAME` masking real binary from `ps`

afadesigns:
ZC1981 — Warn on `exec -a NAME cmd` — replaces `argv[0]`, hides the real binary from `ps`

What: Script calls `exec -a NAME $BIN`.
Why: `-a` sets `argv[0]` on the `exec`'d process to `NAME` instead of the real program. `ps`, `top`, `/proc/*/cmdline`-based audit, and systemd unit accounting show `NAME` — the actual binary only surfaces via `/proc/PID/exe`, which most monitoring does not read. Legitimate use cases exist (login shells spelling themselves `-zsh`), but the feature is also a cheap disguise for reverse shells or cron-triggered helpers.
Fix suggestion: Run the binary at its real path so operators can match process name to the on-disk file. Only use `-a` where the alias is documented and expected.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1981` passes
- [x] `scripts/update-version.sh` → 0.9.78

### DIFF
--- a/pkg/katas/katatests/zc1981_test.go
+++ b/pkg/katas/katatests/zc1981_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1981(t *testing.T) {
+	tests := []struct {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1004, opened 2026-04-20
feat: ZC1982 — detect `ipcrm -a` wiping every SysV IPC object

afadesigns:
ZC1982 — Error on `ipcrm -a` — removes every SysV IPC object, breaks Postgres/Oracle/shm apps

What: Script calls `ipcrm -a`.
Why: Deletes every SysV shared-memory segment, semaphore set, and message queue owned by the caller (or, as root, every object on the host). Services that rely on SysV IPC — PostgreSQL shared buffers, Oracle SGA, shm-based mutexes — lose their backing store mid-transaction and either SIGSEGV or return `EINVAL` on the next access.
Fix suggestion: Scope the removal with `ipcrm -m ID`/`-s ID`/`-q ID` against the specific identifier reported by `ipcs -a`, after confirming no running process has it attached.
Severity: Error

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1982` passes
- [x] `scripts/update-version.sh` → 0.9.79

### DIFF
--- a/pkg/katas/katatests/zc1982_test.go
+++ b/pkg/katas/katatests/zc1982_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1982(t *testing.T) {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1005, opened 2026-04-20
feat: ZC1983 — detect `setopt CSH_JUNKIE_QUOTES` breaking multi-line quoted strings

afadesigns:
ZC1983 — Warn on `setopt CSH_JUNKIE_QUOTES` — single/double-quoted strings that span lines become errors

What: Script flips `CSH_JUNKIE_QUOTES` on (`setopt CSH_JUNKIE_QUOTES` or `unsetopt NO_CSH_JUNKIE_QUOTES`).
Why: With the option off (default) Zsh happily lets `"foo\nbar"` and `'line1\nline2'` span physical lines. Turning it on makes the parser emit an error on the first newline inside a quote, breaking inlined SQL/JSON/heredoc-style payloads and anything autoloaded afterward. The diagnostic points at the closing quote, not at the option toggle.
Fix suggestion: Leave the option off. If csh-style strictness is genuinely required for one helper, scope with `emulate -LR csh` inside that function.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1983` passes
- [x] `scripts/update-version.sh` → 0.9.80

### DIFF
--- a/pkg/katas/katatests/zc1983_test.go
+++ b/pkg/katas/katatests/zc1983_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1983(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1007, opened 2026-04-20
feat: ZC1985 — detect `setopt SH_FILE_EXPANSION` flipping Zsh expansion order

afadesigns:
ZC1985 — Warn on `setopt SH_FILE_EXPANSION` — expansion order flips from Zsh-native to sh/bash, `~` leaks

What: Script flips `SH_FILE_EXPANSION` on (`setopt SH_FILE_EXPANSION` or `unsetopt NO_SH_FILE_EXPANSION`).
Why: Default Zsh runs parameter expansion first, then filename/`~` expansion — so a `VAR='~/cache'` keeps the tilde literal when later referenced. With the option on, the order flips to POSIX (filename first, parameter second): the same value suddenly resolves `~` to `\$HOME`, `~other` into another user's home, and `=cmd` looks up `\$PATH`.
Fix suggestion: Keep the option off. If POSIX ordering is genuinely needed for one helper, scope with `emulate -LR sh` inside the function.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1985` passes
- [x] `scripts/update-version.sh` → 0.9.82

### DIFF
--- a/pkg/katas/katatests/zc1985_test.go
+++ b/pkg/katas/katatests/zc1985_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1985(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1008, opened 2026-04-20
feat: ZC1986 — detect `touch -d`/`-t`/`-r` timestamp rewrite

afadesigns:
ZC1986 — Warn on `touch -d` / `-t` / `-r` — explicit timestamp write is a common antiforensics pattern

What: Script calls `touch -d`, `touch -t`, or `touch -r $REF` (the `--date`/`--time`/`--reference` forms).
Why: All three write a specific atime/mtime rather than the current clock. Legitimate cases exist (mirror re-stamp, reproducible build output), but the pattern also matches the classic "age the dropped file" antiforensics trick — normalise a new binary to match its neighbours so `find -mtime` triage skips it.
Fix suggestion: In reproducible-build contexts, derive from `\$SOURCE_DATE_EPOCH` (`touch -d @\$SOURCE_DATE_EPOCH`) so operators can recognise the intent. Audit rules should flag the forms in production scripts.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1986` passes
- [x] `scripts/update-version.sh` → 0.9.83

### DIFF
--- a/pkg/katas/katatests/zc1986_test.go
+++ b/pkg/katas/katatests/zc1986_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1986(t *testing.T) {
+	tests := []struct {
+		name     string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1009, opened 2026-04-20
feat: ZC1987 — detect `setopt BRACE_CCL` csh-style brace character class

afadesigns:
ZC1987 — Warn on `setopt BRACE_CCL` — `{a-z}` expands to each character instead of staying literal

What: Script flips `BRACE_CCL` on (`setopt BRACE_CCL` or `unsetopt NO_BRACE_CCL`).
Why: Default Zsh leaves `{a-z}` literal; only the numeric range `{1..10}` expands. Turning the option on promotes single-character ranges and enumerations to csh-style classes, so `{a-z}` becomes every letter and `{ABC}` becomes `A B C`. Regex fragments, hex masks, and CI job names that embed single-character ranges expand unexpectedly.
Fix suggestion: Leave the option off. When a real range is wanted, use the unambiguous `{a..z}` form and quote literals that contain `{…}`.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1987` passes
- [x] `scripts/update-version.sh` → 0.9.84

### DIFF
--- a/pkg/katas/katatests/zc1987_test.go
+++ b/pkg/katas/katatests/zc1987_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1987(t *testing.T) {
+	tests := []struct {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1010, opened 2026-04-20
feat: ZC1988 — detect `nsupdate -y` TSIG key in argv

afadesigns:
ZC1988 — Error on `nsupdate -y HMAC:NAME:SECRET` — TSIG key visible in argv and shell history

What: Script calls `nsupdate -y [alg:]name:base64secret`.
Why: The shared secret is on the command line, so `ps auxf`, `/proc/*/cmdline`, and `\$HISTFILE` capture it. Whoever owns the key can rewrite any zone that trusts it — DNS hijack, MX hijack, ACME domain-validation bypass.
Fix suggestion: Use `nsupdate -k \$KEYFILE` with a `0600` keyfile. When the secret must come from a secret store, pipe via `nsupdate -k /dev/stdin <<<"\$KEYFILE_CONTENTS"` so the raw material never lands in argv.
Severity: Error

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1988` passes
- [x] `scripts/update-version.sh` → 0.9.85

### DIFF
--- a/pkg/katas/katatests/zc1988_test.go
+++ b/pkg/katas/katatests/zc1988_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1988(t *testing.T) {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1011, opened 2026-04-20
feat: ZC1989 — detect `setopt REMATCH_PCRE` flipping regex engine mid-script

afadesigns:
ZC1989 — Warn on `setopt REMATCH_PCRE` — `[[ =~ ]]` regex flips from POSIX ERE to PCRE, changes semantics

What: Script flips `REMATCH_PCRE` on (`setopt REMATCH_PCRE` or `unsetopt NO_REMATCH_PCRE`).
Why: Default `[[ \$str =~ pat ]]` uses POSIX extended regex. With the option on (after `zmodload zsh/pcre`), every later match runs under PCRE. Patterns that pass through both engines change meaning: `\b` is a word boundary in PCRE but literal `b` in ERE; `\d`/`\s`/`\w` exist only in PCRE; lookahead/lookbehind parse in PCRE but error in ERE; inline flags like `(?i)` only exist in PCRE.
Fix suggestion: Use the explicit `pcre_compile` + `pcre_match` API when PCRE syntax is needed, or scope with `setopt LOCAL_OPTIONS REMATCH_PCRE` inside the single function that uses PCRE.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1989` passes
- [x] `scripts/update-version.sh` → 0.9.86

### DIFF
--- a/pkg/katas/katatests/zc1989_test.go
+++ b/pkg/katas/katatests/zc1989_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1989(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected []katas.Violation
+	}{
+		{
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1013, opened 2026-04-20
feat: ZC1991 — detect `setopt CSH_NULLCMD` breaking bare-redirect idioms

afadesigns:
ZC1991 — Warn on `setopt CSH_NULLCMD` — bare `> file` raises an error instead of running `\$NULLCMD`

What: Script flips `CSH_NULLCMD` on (`setopt CSH_NULLCMD` or `unsetopt NO_CSH_NULLCMD`).
Why: Default Zsh executes `\$NULLCMD` (initially `cat`) when a line has redirections but no command, so `> file < input` is a valid copy and `< file` pages through it. With the option on, Zsh follows csh — any command line without an explicit command is a parse error regardless of redirections. Scripts relying on bare-redirect idioms (log truncation, drop-in includes) stop with a confusing `parse error near '<'`.
Fix suggestion: Keep the option off. Write `: > file` (or `true > file`) explicitly when the intent is to truncate a file.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1991` passes
- [x] `scripts/update-version.sh` → 0.9.88

### DIFF
--- a/pkg/katas/katatests/zc1991_test.go
+++ b/pkg/katas/katatests/zc1991_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1991(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1014, opened 2026-04-20
feat: ZC1992 — detect `pkexec` PolicyKit elevation from scripts

afadesigns:
ZC1992 — Warn on `pkexec cmd` — PolicyKit privilege elevation is historically bug-prone and hard to audit from scripts

What: Script calls `pkexec CMD`.
Why: `pkexec` lifts a command to the UID configured in a PolicyKit `.policy` (typically root) after consulting an auth agent. A non-interactive script has no agent to prompt, so the call either depends on a pre-authorised `.policy` override or fails in a confusing way. The binary also has a poor CVE record (CVE-2021-4034 pwnkit, CVE-2017-16089) and its audit trail is split across journald and `/var/log/auth.log`.
Fix suggestion: Prefer `sudo` with a targeted `sudoers.d` drop-in for scripted elevation, or run the script under a systemd unit with `User=` / `AmbientCapabilities=` when specific capabilities are enough.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1992` passes
- [x] `scripts/update-version.sh` → 0.9.89

### DIFF
--- a/pkg/katas/katatests/zc1992_test.go
+++ b/pkg/katas/katatests/zc1992_test.go
@@ -1,0 +1,41 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1992(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1015, opened 2026-04-20
feat: ZC1993 — detect `setopt KSH_TYPESET` re-splitting typeset RHS

afadesigns:
ZC1993 — Warn on `setopt KSH_TYPESET` — `typeset var=\$val` starts word-splitting the RHS

What: Script flips `KSH_TYPESET` on (`setopt KSH_TYPESET` or `unsetopt NO_KSH_TYPESET`).
Why: Off by default, Zsh treats each `typeset`/`local`/`declare` assignment as one token after `=`, so `typeset msg="a b c"` produces a single-element string. With the option on, Zsh follows ksh — words on the line are independent assignments, and the RHS re-splits on whitespace. Helpers that used to accept `typeset path=\$HOME/My Files` suddenly treat `Files` as a second variable.
Fix suggestion: Keep the option off. When ksh compatibility is genuinely required, scope with `emulate -LR ksh` inside the helper that needs it.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1993` passes
- [x] `scripts/update-version.sh` → 0.9.90

### DIFF
--- a/pkg/katas/katatests/zc1993_test.go
+++ b/pkg/katas/katatests/zc1993_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1993(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1016, opened 2026-04-20
feat: ZC1994 — detect `lvreduce -f`/`-y` skipping shrink-confirm

afadesigns:
ZC1994 — Error on `lvreduce -f` / `lvreduce -y` — shrinks the LV without checking the filesystem above

What: Script calls `lvreduce` with `-f`/`-y`/`--force`/`--yes`.
Why: `lvreduce -L SIZE \$LV` cuts the block device under an existing filesystem. The interactive prompt exists precisely because ext4/xfs/btrfs do not shrink themselves — LVM lops off the tail even though the filesystem still believes those blocks are allocated, and the next mount returns corruption or missing files.
Fix suggestion: Shrink the filesystem first (`resize2fs \$LV \$NEWSIZE`; xfs can't shrink — offline backup + recreate), verify with `fsck`/`df`, then either confirm the prompt or use `lvreduce --resizefs` which sequences both steps atomically.
Severity: Error

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1994` passes
- [x] `scripts/update-version.sh` → 0.9.91

### DIFF
--- a/pkg/katas/katatests/zc1994_test.go
+++ b/pkg/katas/katatests/zc1994_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1994(t *testing.T) {
+	tests := []struct {
+		name     string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1017, opened 2026-04-20
feat: ZC1995 — detect `unsetopt BG_NICE` bg jobs stealing interactive priority

afadesigns:
ZC1995 — Warn on `unsetopt BG_NICE` — background jobs run at full interactive priority, starve the foreground

What: Script flips `BG_NICE` off (`unsetopt BG_NICE` or `setopt NO_BG_NICE`).
Why: Default Zsh gives every `cmd &` bg job `nice +5`, so long-running work does not starve interactive input. With the option off, bg jobs compete at the same priority as the shell — SSH keystroke handling, editor redraws, and batch fan-out all feel laggy; a single CPU-bound bg job can peg every core of a container shared with a human operator.
Fix suggestion: Keep the option on. When a specific bg job legitimately needs full priority (audio pipeline, real-time simulator) wrap that one with `nice -n 0 -- cmd &` or a systemd unit with `Nice=` instead of flipping globally.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1995` passes
- [x] `scripts/update-version.sh` → 0.9.92

### DIFF
--- a/pkg/katas/katatests/zc1995_test.go
+++ b/pkg/katas/katatests/zc1995_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1995(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1018, opened 2026-04-20
feat: ZC1996 — detect `unshare -U`/`-r` unprivileged root-in-userns primitive

afadesigns:
ZC1996 — Warn on `unshare -U` / `-r` — unprivileged user namespace maps caller to root inside the NS

What: Script calls `unshare` with `-U` / `-r` / `-Ur` / `--user` / `--map-root-user` (also short-flag bundles like `-Urm`).
Why: The pair opens a user namespace and maps the caller's UID to 0 inside it. Legitimate in rootless-container runtimes, but also the opening move for a long list of kernel-LPE chains — once uid 0 inside a userns, the caller can create additional mount/net/cgroup namespaces, overlay-mount attacker paths, and poke kernel attack surface that is normally gated on `CAP_SYS_ADMIN`.
Fix suggestion: Route legitimate rootless runtimes through `bwrap`, `podman --rootless`, `flatpak`, etc. so the invocation is recognisable to audit rules. Keep raw `unshare -U` out of production scripts unless explicitly documented.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1996` passes
- [x] `scripts/update-version.sh` → 0.9.93

### DIFF
--- a/pkg/katas/katatests/zc1996_test.go
+++ b/pkg/katas/katatests/zc1996_test.go
@@ -1,0 +1,70 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1996(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected []katas.Violation
+	}{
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1019, opened 2026-04-20
feat: ZC1997 — detect `setopt HIST_NO_FUNCTIONS` hiding function defs from history

afadesigns:
ZC1997 — Warn on `setopt HIST_NO_FUNCTIONS` — function definitions skipped from `\$HISTFILE`, breaks forensic trail

What: Script flips `HIST_NO_FUNCTIONS` on (`setopt HIST_NO_FUNCTIONS` or `unsetopt NO_HIST_NO_FUNCTIONS`).
Why: Default Zsh records every typed command in `\$HISTFILE`, including function definitions. With the option on, function-definition commands are suppressed from history. On a multi-admin or shared-root box the forensic trail breaks — the definition vanishes while the invocation that used it still shows, leaving responders with a name that no longer resolves to anything on disk or in any log.
Fix suggestion: Keep the option off. Scope any hiding needs with a `zshaddhistory { return 1 }` hook inside the specific function that touches the secret.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1997` passes
- [x] `scripts/update-version.sh` → 0.9.94

### DIFF
--- a/pkg/katas/katatests/zc1997_test.go
+++ b/pkg/katas/katatests/zc1997_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1997(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1020, opened 2026-04-20
feat: ZC1998 — detect `tpm2_clear`/`tpm2 clear` wiping TPM storage hierarchy

afadesigns:
ZC1998 — Error on `tpm2_clear` / `tpm2 clear` — wipes TPM storage hierarchy, kills every sealed key

What: Script calls `tpm2_clear` or `tpm2 clear` (the tpm2-tools `TPM2_Clear` wrapper).
Why: The command invalidates every object sealed against the storage hierarchy — LUKS-TPM2 keyslots, `systemd-cryptenroll --tpm2-device` slots, sshd TPM-backed host keys, SecureBoot measured-boot state. The machine can still boot, but any disk that unlocked through the TPM now needs a recovery passphrase, and every TLS cert issued from a TPM-sealed CA loses its anchor. No undo.
Fix suggestion: Run only under a documented recovery runbook with recovery material in hand. Never put it in an automated scheduled script.
Severity: Error

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1998` passes
- [x] `scripts/update-version.sh` → 0.9.95

### DIFF
--- a/pkg/katas/katatests/zc1998_test.go
+++ b/pkg/katas/katatests/zc1998_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1998(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1021, opened 2026-04-20
feat: ZC1999 — detect `setopt AUTO_NAMED_DIRS` auto-registering dir-valued scalars as `~name`

afadesigns:
ZC1999 — Warn on `setopt AUTO_NAMED_DIRS` — every scalar holding a directory path becomes `~name`

What: Script flips `AUTO_NAMED_DIRS` on (`setopt AUTO_NAMED_DIRS` or `unsetopt NO_AUTO_NAMED_DIRS`).
Why: Off by default, Zsh only treats `~USER`/explicit `hash -d` entries as named directories. With the option on, any scalar whose value is an existing directory is auto-registered, so `release=/srv/app/releases` makes `~release/config` a valid path. That silently collides with real usernames (`alice` in `/etc/passwd` vs. an `alice=\$HOME/stage` scalar) and turns every unquoted `~\$var` inside a heredoc or `cd` arg into a prompt-expander replacement with the wrong path.
Fix suggestion: Keep the option off. When a script legitimately wants a named dir, register it explicitly with `hash -d NAME=PATH`.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC1999` passes
- [x] `scripts/update-version.sh` → 0.9.96

### DIFF
--- a/pkg/katas/katatests/zc1999_test.go
+++ b/pkg/katas/katatests/zc1999_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC1999(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected []katas.Violation
+	}{
+		{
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1022, opened 2026-04-20
feat: ZC2000 — detect `kubectl taint nodes ...:NoExecute` mass pod eviction

afadesigns:
ZC2000 — Error on `kubectl taint nodes \$NODE key=value:NoExecute` — evicts every non-tolerating pod off the node

What: Script calls `kubectl taint nodes ...` with an effect of `:NoExecute`.
Why: `NoExecute` kicks every existing pod off the node unless the pod spec explicitly tolerates it. A single mistyped command on `--all` nodes empties the cluster in seconds and triggers cascade reschedules that overwhelm the scheduler.
Fix suggestion: Prefer `kubectl drain \$NODE` (which respects PodDisruptionBudget and runs PreStop hooks) or a `:NoSchedule` taint for gentle drain. Reserve `:NoExecute` for genuine incident response with a runbook and a safety countdown.
Severity: Error

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC2000` passes
- [x] `scripts/update-version.sh` → 0.9.97

### DIFF
--- a/pkg/katas/katatests/zc2000_test.go
+++ b/pkg/katas/katatests/zc2000_test.go
@@ -1,0 +1,46 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC2000(t *testing.T) {
+	tests := []struct {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1023, opened 2026-04-20
feat: ZC2001 — detect `unsetopt EVAL_LINENO` collapsing eval'd line numbers

afadesigns:
ZC2001 — Warn on `unsetopt EVAL_LINENO` — `\$LINENO` inside `eval` stops tracking source, stack traces go blank

What: Script flips `EVAL_LINENO` off (`unsetopt EVAL_LINENO` or `setopt NO_EVAL_LINENO`).
Why: On by default, `EVAL_LINENO` keeps `\$LINENO`, `\$funcfiletrace`, and `\$funcstack` pointing at the line inside the `eval`ed string where the error actually happened. Off, `\$LINENO` collapses to the line that launched the `eval`, so every runtime error in a generated config, lazy-loaded function, or `compile`d string reports the same line and the stack trace drops every frame past the eval.
Fix suggestion: Keep the option on. If strict POSIX-matching line numbers are needed for one helper, scope with `emulate -LR sh` inside that function.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC2001` passes
- [x] `scripts/update-version.sh` → 0.9.98

### DIFF
--- a/pkg/katas/katatests/zc2001_test.go
+++ b/pkg/katas/katatests/zc2001_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC2001(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1024, opened 2026-04-20
feat: ZC2002 — detect `crictl rmi -a`/`rm -af` node-level CRI wipe

afadesigns:
ZC2002 — Error on `crictl rmi -a` / `crictl rm -af` — wipes every image/container on the Kubernetes node

What: Script calls `crictl rmi -a` or `crictl rm -af` (full `-a --all` + `-f --force`).
Why: `crictl` talks directly to the node CRI runtime (containerd, CRI-O), below the kubelet and the cluster API. `rmi -a` removes every cached image, including the ones currently backing running pods — the kubelet must re-pull and a registry blip turns the node Unready. `rm -af` force-removes every container on the node, skipping PreStop hooks and PodDisruptionBudget.
Fix suggestion: Route maintenance through `kubectl drain \$NODE` + `kubectl delete pod --grace-period=30`. Reserve `crictl` for a cordoned, drained node with a documented recovery plan.
Severity: Error

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC2002` passes
- [x] `scripts/update-version.sh` → 0.9.99

### DIFF
--- a/pkg/katas/katatests/zc2002_test.go
+++ b/pkg/katas/katatests/zc2002_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC2002(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1025, opened 2026-04-20
feat: ZC2003 — detect `setopt KSH_ZERO_SUBSCRIPT` breaking `$arr[0]` alias

afadesigns:
ZC2003 — Warn on `setopt KSH_ZERO_SUBSCRIPT` — `\$arr[0]` stops aliasing the first element

What: Script flips `KSH_ZERO_SUBSCRIPT` on (`setopt KSH_ZERO_SUBSCRIPT` or `unsetopt NO_KSH_ZERO_SUBSCRIPT`).
Why: Default Zsh treats `\$arr[0]` as a quirk-compatibility alias for `\$arr[1]` — `arr=(a b c); echo \$arr[0]` prints `a`, and `arr[0]=new` rewrites the first element. With the option on, Zsh switches to ksh semantics: `\$arr[0]` becomes a distinct slot, reads silently return empty, and `arr[0]=new` stops touching `\$arr[1]`. Any Zsh code that intentionally used `\$arr[0]` as a shortcut breaks.
Fix suggestion: Leave the option off. Use `\$arr[1]` explicitly for the first element. When ksh semantics are genuinely required, scope with `emulate -LR ksh` inside the specific function.
Severity: Warning

## Test plan
- [x] `go test ./pkg/katas/katatests/ -run TestZC2003` passes
- [x] `scripts/update-version.sh` → 1.0.0 (1000 katas)

### DIFF
--- a/pkg/katas/katatests/zc2003_test.go
+++ b/pkg/katas/katatests/zc2003_test.go
@@ -1,0 +1,58 @@
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+func TestZC2003(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected []katas.Violation
+	}{
+		{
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1028, opened 2026-04-20
fix: narrow 5 duplicate kata pairs (ZC1327/1441/1826/1978/1999)

afadesigns:
Post-1000 duplicate-hunter sweep surfaced 5 pairs firing on the same input with overlapping advice. Per project rule "once committed, fix — don't remove", each pair is narrowed so one kata owns the case.

## Pairs addressed

| Kept / owner | Narrowed sibling | Change |
| --- | --- | --- |
| **ZC1934** (AUTO_NAME_DIRS) | **ZC1999** | ZC1999 was detecting `AUTO_NAMED_DIRS` (with a trailing `D`) — not a real Zsh option (`setopt` rejects it). Rewritten as a typo detector that points at ZC1934 and bumped to **Error**. |
| **ZC1892** (numeric setuid install) | **ZC1826** | ZC1826's numeric `-m 4xxx/2xxx/6xxx` handling removed; replaced with symbolic-mode detection (`u+s`, `g+s`, `ug+s`, `u=rwxs`). ZC1892 keeps numeric and covers `mkdir` too. |
| **ZC1545** (`docker system prune -af --volumes`) | **ZC1441** | ZC1441 now skips when `--volumes` is present so the stricter ZC1545 owns that input. |
| **ZC1487** (`history -c/-d` anti-forensics) | **ZC1327** | ZC1327 drops `-c` from its flag list and narrows to `-w/-r/-a` (Bash-only portability flags). |
| **ZC1200** (`ftp`) | **ZC1978** | ZC1978 narrows to `tftp` only. `ftp` remains owned by ZC1200. Title + description updated; numeric fixture tests tightened. |

## Test plan
- [x] `go test ./...` green across all 11 packages
- [x] Tests exercise the new narrow scope per kata AND confirm the sibling no longer double-fires
- [x] `KATAS.md` regenerated to reflect updated titles/descriptions

### DIFF
--- a/KATAS.md
+++ b/KATAS.md
@@ -6,8 +6,8 @@ Auto-generated list of all 1000 implemented checks. Do not edit by hand — rege
 
 | Severity | Count |
 | :--- | ---: |
-| `error` | 219 |
-| `warning` | 450 |
+| `error` | 220 |
+| `warning` | 449 |
 | `info` | 66 |
 | `style` | 265 |
 | **total** | **1000** |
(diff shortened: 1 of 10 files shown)


redteamx (approved):
LGTM.

PR #1029, opened 2026-04-20
ci: unblock Spell Check — allowlist `exportfs`, rephrase `mis-keying` in ZC2000

afadesigns:
The `typos` workflow has been red on every push since ~ZC1976 (NFS `exportfs` tool) and ZC2000 (`mis-keying`).

- **`_typos.toml`** — adds `exportfs` to `[default.extend-identifiers]` alongside `mke2fs` / `chage`. That silences 45+ false positives across KATAS.md and `pkg/katas/zc1976.go`.
- **`pkg/katas/zc2000.go`** — rephrases `mis-keying` to `typoed` so `typos` does not trip on the word fragment.
- **`KATAS.md`** regenerated via `go run ./internal/tools/gen-katas-md` so the updated description flows through.

Needed for the v1.0.0 release tag — release-auditor flagged the red workflow as a blocker.

### Test plan
- [x] `go test ./...` green
- [x] Manual grep confirms no other `mis-` prose in tracked `*.go`

### DIFF
--- a/KATAS.md
+++ b/KATAS.md
@@ -11978,7 +11978,7 @@ Disable by adding `ZC1999` to `disabled_katas` in `.zshellcheckrc`.
 
 **Severity:** `error`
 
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1031, opened 2026-04-20
docs: correct author name (Andreas Fahl) + contact email (github@afadesign.co)

afadesigns:
Fix author metadata and contact channels so all public-facing docs point at the correct identity and inbox.

## Changes

- **`CITATION.cff`** — `family-names: Fahl`, `given-names: Andreas` (was incorrectly filled as `Designs, Afa`); adds `email: github@afadesign.co`.
- **`SECURITY.md`** — reporting email now `github@afadesign.co` (was the legacy `afadesign.official@gmail.com`).
- **`CODE_OF_CONDUCT.md`** — enforcement email now `github@afadesign.co` (same migration).

Historical `CHANGELOG.md` entries that reference the older `afadesign.official@gmail.com` are left untouched as accurate records of what was true at the time of each release.

PR #1032, opened 2026-04-20
build(deps): action group bump + golangci-lint v2 config migration

afadesigns:
Supersedes #350. The dependabot group bump (19 action updates) was blocked by the \`golangci/golangci-lint-action@v9\` upgrade: v9 ships golangci-lint v2, which rejected the v1 config shape.

This PR keeps every bump from #350 and adds the required \`.golangci.yml\` migration in a second commit so CI goes green.

## Config migration

- \`version: \"2\"\` top-level key added.
- \`linters.disable-all: true\` → \`linters.default: none\`.
- \`linters-settings\` → \`linters.settings\`.
- \`gosimple\` and \`unused\` dropped (both subsumed into \`staticcheck\` in v2; still ran transitively).
- \`gofumpt\` promoted to the new \`formatters\` block.
- \`misspell.ignore-words\` → \`misspell.ignore-rules\` (rename only).
- Revive rule exclusions kept verbatim.

Behaviour delta: none. Same enabled linter coverage, same rule exclusions, same ignore-rules.

## Test plan
- [x] Rebased on main after #1031 (contacts fix) landed.
- [x] \`go test ./...\` green locally.
- [x] CI \`lint\` job will execute under v9 runner using the v2 config.

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -24,9 +24,9 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
-      - uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594 # v2.16.0
+      - uses: step-security/harden-runner@8d3c67de8e2fe68ef647c8db1e6a09f647780f40 # v2.19.0
         with:
           egress-policy: audit
-      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.1.7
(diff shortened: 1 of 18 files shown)


redteamx (approved):
LGTM.

afadesigns:
Closing and re-opening — the CI run showed golangci-lint v2 also surfaces ~20 pre-existing staticcheck QF1001–QF1003 findings across pkg/katas that are not in the scope of a dependency bump. Will submit a fresh branch that holds golangci-lint-action at v6.5.2; v9 migration + lint-finding cleanup will ship in a follow-up.

PR #1034, opened 2026-04-20
feat: XDG Base Directory support — ~/.config/zshellcheck/config.{yml,yaml} merged with ~/.zshellcheckrc and ./.zshellcheckrc

### LINKED ISSUE #309: [FEAT] Support XDG Base Directory Specification

afadesigns:
Rebased from #310 (original author: @norwd, credited in commits on this branch). Adds the review follow-ups so the feature can land: \`go mod tidy\`, empty-path guard, and 3 test cases covering merge-order, empty-path skip, and XDG-only behaviour.

Closes #309. Supersedes #310.

## Behaviour

\`zshellcheck\` now looks for configuration in three layers, merged in this order (later wins):

1. \`\$XDG_CONFIG_HOME/zshellcheck/config.yml\` (or fallback \`\$HOME/.config/zshellcheck/config.yml\`), with \`.yaml\` extension as a secondary search
2. \`\$HOME/.zshellcheckrc\` (the existing single-user config)
3. \`./.zshellcheckrc\` (the existing per-project config)

Every existing \`.zshellcheckrc\`-only setup keeps working unchanged — the XDG layer is additive and only takes effect when a file is found there.

## Test plan
- [x] \`go test ./...\` green across all 11 packages
- [x] \`golangci-lint run ./...\` clean
- [x] \`go mod tidy\` idempotent (go.mod now has \`github.com/adrg/xdg\` as a direct dep)
- [x] New tests: \`TestLoadConfig_EmptyPathSkipped\`, \`TestLoadConfig_MergeOrderLocalWins\`, \`TestLoadConfig_XDGPathOnly\`

### DIFF
--- a/cmd/zshellcheck/main.go
+++ b/cmd/zshellcheck/main.go
@@ -77,9 +77,6 @@ func run() int {
 		fmt.Fprint(os.Stderr, config.Banner)
 	}
 
-	// Start with default config
-	config := config.DefaultConfig()
-
 	configFileXdgConfigPath, _ := xdg.ConfigFile("zshellcheck/config.yaml")
 	configFileHomePath := filepath.Join(xdg.Home, ".zshellcheckrc")
 	config, err := loadConfig(configFileXdgConfigPath, configFileHomePath, ".zshellcheckrc")
--- a/cmd/zshellcheck/main.go
+++ b/cmd/zshellcheck/main.go
@@ -77,7 +77,7 @@ func run() int {
 		fmt.Fprint(os.Stderr, config.Banner)
 	}
 
-	configFileXdgConfigPath, _ := xdg.ConfigFile("zshellcheck/config.yaml")
+	configFileXdgConfigPath, _ := xdg.SearchConfigFile("zshellcheck/config.yaml")
 	configFileHomePath := filepath.Join(xdg.Home, ".zshellcheckrc")
 	config, err := loadConfig(configFileXdgConfigPath, configFileHomePath, ".zshellcheckrc")
 	if err != nil {
--- a/cmd/zshellcheck/main.go
+++ b/cmd/zshellcheck/main.go
@@ -77,7 +77,12 @@ func run() int {
 		fmt.Fprint(os.Stderr, config.Banner)
 	}
 
-	configFileXdgConfigPath, _ := xdg.SearchConfigFile("zshellcheck/config.yaml")
+	// Pick either config.yml or config.yaml, but not both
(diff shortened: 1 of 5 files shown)


redteamx (approved):
LGTM.

redteamx (approved):
LGTM.

redteamx (approved):
LGTM.

Issue #345, opened 2026-04-16
Critical: 10 katas (ZC1022-ZC1029, ZC1033, ZC1035) are identical copies detecting let statement


afadesigns:
## Critical duplicate cluster

Ten katas detect exactly the same pattern (`let` statement) with identical Check function bodies, identical titles, identical descriptions, and identical messages:

- ZC1022
- ZC1023
- ZC1024
- ZC1025
- ZC1026
- ZC1027
- ZC1028
- ZC1029
- ZC1033
- ZC1035

All of them:

```go
if let, ok := node.(*ast.LetStatement); ok {
    violations = append(violations, Violation{
        KataID:  "ZC####",
        Message: "Use `$((...))` for arithmetic expansion instead of `let`.",
        ...
        Level:   SeverityStyle,
    })
}
```

Only the `KataID` field differs.

## Impact

Any script with `let x=1` emits **ten** identical violations. This inflates kata count without providing coverage, and would be caught immediately by any external reviewer looking at the kata catalogue.

## Decision Needed

Per project policy ("once committed, fix, do not remove"), the options are:

1. **Keep ZC1022 as canonical**, convert ZC1023-ZC1029, ZC1033, ZC1035 into distinct scope (e.g. different `let` sub-forms: `let x=1+2`, `let "x+=1"`, `let x=$y`, `let x` with no RHS, `let` with multiple vars on one line, `let` inside a pipe, etc.). Requires rewriting 9 Check functions with narrow, non-overlapping detection.
2. **Keep ZC1022 active**, mark the other 9 as deprecated aliases that emit the same violation but refer users to ZC1022.
3. **Community discussion** — accept that the 9 are pure duplicates and document them as historical noise. Future detections for `let` sub-patterns get new IDs.

## References

- pkg/katas/zc1022.go
- pkg/katas/zc1023.go
- pkg/katas/zc1024.go
- pkg/katas/zc1025.go
- pkg/katas/zc1026.go
- pkg/katas/zc1027.go
- pkg/katas/zc1028.go
- pkg/katas/zc1029.go
- pkg/katas/zc1033.go
- pkg/katas/zc1035.go

## Priority

**High.** Any automated duplicate-detection run from a potential reviewer or competitor will find this cluster in seconds. Addressing it before public beta launch (v0.3.35) is recommended.

afadesigns:
Resolved by #1035 — 10 duplicate let-statement katas retired as no-op stubs; canonical detection in the remaining ZC.

Issue #344, opened 2026-04-16
Duplicate detections: ZC1108 and ZC1277 both flag tr for case conversion


afadesigns:
## Overlap

- `ZC1108` [Style] — "Use Zsh case conversion instead of `tr`"
- `ZC1277` [Style] — "Use Zsh `${var:l}` / `${var:u}` instead of `tr` for case conversion"

Same detection target (`tr a-z A-Z` or `tr [:lower:] [:upper:]`). Difference: `ZC1277` is specific about the Zsh replacement syntax.

Related (not overlapping):
- `ZC1145` [Style] — "Avoid `tr -d` for character deletion" — scoped to deletion, keep
- `ZC1292` [Style] — "Use Zsh `${var//old/new}` instead of `tr` for character translation" — scoped to substitution, keep

## Decision Needed

Pick one canonical kata for case conversion. The other could be narrowed to a sub-pattern (e.g. `tr` with explicit character ranges vs POSIX classes).

## References

- `pkg/katas/zc1108.go`
- `pkg/katas/zc1277.go`

afadesigns:
Resolved by #1035 — ZC1277 retired as no-op stub; canonical tr→${(U)x}/${(L)x} detection stays in ZC1108.

Issue #343, opened 2026-04-16
Duplicate detections: ZC1009, ZC1018, ZC1278 overlap on expr arithmetic


afadesigns:
## Overlap

Three-kata cluster around arithmetic — pairwise overlap on `expr`:

- `ZC1009` [Style] — "Use `((...))` for C-style arithmetic"
- `ZC1018` [Info] — "Use `((...))` for C-style arithmetic instead of `expr`"
- `ZC1278` [Style] — "Use `$(( ))` instead of `expr` for arithmetic"

`ZC1009` and `ZC1018` have nearly identical titles, with `ZC1018` being more explicit about targeting `expr`. `ZC1278` recommends the expression-form syntax variant.

## Decision Needed

- Pick one canonical kata for the expr→arithmetic migration recommendation.
- Scope the others to distinct sub-cases (e.g. one for command-substitution-wrapping `$(expr ...)`, one for bare `expr ...`).
- Reconcile severity: `ZC1009` is Style, `ZC1018` is Info — should be the same if they detect the same pattern.

## References

- `pkg/katas/zc1009.go`
- `pkg/katas/zc1018.go`
- `pkg/katas/zc1278.go`

afadesigns:
Resolved by #1035 — ZC1018, ZC1278 retired as no-op stubs; canonical expr-arithmetic detection stays in ZC1009.

Issue #342, opened 2026-04-16
Duplicate detections: ZC1005 and ZC1019 both recommend whence over which


afadesigns:
## Overlap

Two katas with effectively identical titles and detection targets:

- `ZC1005` [Info] — "Use whence instead of which"
- `ZC1019` [Info] — "Use `whence` instead of `which`"

Both flag plain `which cmd` invocations in favor of Zsh built-in `whence`.

Related (not overlap, but adjacent):
- `ZC1155` [Info] — "Use `whence -a` instead of `which -a`" (flag-specific, keep)
- `ZC1271` [Style] — "Use `command -v` instead of `which` for command existence checks" (different recommendation, keep)

## Decision Needed

Either deprecate ZC1005 or ZC1019 (keep the better implementation), or narrow one to a specific sub-pattern the other misses.

## References

- `pkg/katas/zc1005.go`
- `pkg/katas/zc1019.go`
- `pkg/katas/zc1155.go`
- `pkg/katas/zc1271.go`

afadesigns:
Resolved by #1035 — ZC1019 retired as no-op stub; canonical whence/which detection stays in ZC1005.

PR #1035, opened 2026-04-20
fix: retire 14 duplicate katas (#341 #342 #343 #344 #345)

### LINKED ISSUE #341: Duplicate detections: ZC1038 and ZC1093 both flag cat overuse
## Overlap

Two katas target the same pattern:

- `ZC1038` [Style] — "Avoid useless use of cat"
- `ZC1093` [Style] — "Avoid useless `cat`"

Both trigger on `cat file | cmd` where the file could be read directly by `cmd <file`.

## Decision Needed

Three possible paths:

1. **Deprecate one, narrow scope of the other.** Pick the better implementation as canonical. The deprecated one can emit a pointer to the survivor.
2. **Differentiate.** Scope one narrowly (e.g. `cat` with a single file and single pipe stage → redirect) and the other broadly (e.g. `cat` inside command substitution → `$(<file)`). Verify detection sets do not overlap.
3. **Merge detection logic** into one kata, keep the other as a test-only alias.

Community input welcome — which of the two has the better test corpus and cleaner Check function?

## References

- `pkg/katas/zc1038.go`
- `pkg/katas/zc1093.go`

afadesigns:
Closes #341 #342 #343 #344 #345.

## What

Fourteen katas were firing on the same input as another kata with identical advice:

| Duplicate IDs | Canonical | Issue |
| :--- | :--- | :--- |
| ZC1023, ZC1024, ZC1025, ZC1026, ZC1027, ZC1028, ZC1029, ZC1033, ZC1035 (nine `let` clones) | **ZC1022** | #345 |
| ZC1093 | **ZC1038** (useless `cat`) | #341 |
| ZC1019 | **ZC1005** (`which` → `whence`) | #342 |
| ZC1018 + ZC1278 | **ZC1009** (`expr` → `((...))`) | #343 |
| ZC1277 | **ZC1108** (`tr` → Zsh case conversion) | #344 |

## How

Per project rule *"once committed, fix — don't remove"*, the 14 IDs stay alive as documented no-op stubs so existing `.zshellcheckrc` files that say `disabled_katas: [ZC1023]` keep parsing. Each stub carries a title that points at the canonical kata and a description linking the tracking issue.

## Test plan
- [x] `go test ./...` green — every stub test asserts no violations
- [x] KATAS.md regenerated via `go run ./internal/tools/gen-katas-md`

### DIFF
--- a/KATAS.md
+++ b/KATAS.md
@@ -8,8 +8,8 @@ Auto-generated list of all 1000 implemented checks. Do not edit by hand — rege
 | :--- | ---: |
 | `error` | 220 |
 | `warning` | 449 |
-| `info` | 66 |
-| `style` | 265 |
+| `info` | 64 |
+| `style` | 267 |
 | **total** | **1000** |
 
 ## Table of Contents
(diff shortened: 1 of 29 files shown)


redteamx (approved):
LGTM.

Issue #435, opened 2026-04-17
Parser panic on `dd of=/dev/sda` input


afadesigns:
## Parser panic on `dd of=/dev/sda`-style input

Running the linter against `dd if=image.iso of=/dev/sda bs=4M` crashes the parser with a panic in `pkg/ast/ast.go:662` (AST Walk).

```
panic: ...
  pkg/ast/ast.go:662 +0x217
  pkg/ast/ast.go:648 +0xaee
```

## Reproduction

```zsh
echo "dd if=image.iso of=/dev/sda bs=4M" > /tmp/test.zsh
go run ./cmd/zshellcheck /tmp/test.zsh
```

## Root cause (suspected)

Parser walks into a nil node or un-typed child on arguments containing `of=/dev/...` (path-like values with slashes inside a key=value form). Likely the DeclarationAssignment or argument-parsing path does not handle this shape.

## Impact

Any Zsh script containing `dd` (or other commands) with `path=/dev/...` arguments crashes the linter before any kata runs. Real-world impact: flash-drive-burning scripts, partition tools, backup scripts.

## Proposed fix

Guard `ast.Walk` against nil children at `ast.go:662`, and separately audit the assignment parser for key=value arguments where the value contains path separators.

afadesigns:
Fixed by #1037 (lexer demotes keyword to IDENT when followed by `=`, unblocking `dd if=src of=dst`).

Issue #129, opened 2026-03-16
[BUG] incomplete support for variable substitution


eeweegh:
### Describe the bug

for this (I think valid) input file:
```
#!/usr/bin/env zsh

piep() {
   echo ${1:-piep}
}
```
v0.1.1 zshellcheck gives me this error (on stderr)

`Parser Error in test.zsh: expected next token to be }, got : instead`

### Reproduction Code

```zsh
see description for example.
```

### Actual Output

```text

```

### Operating System

None

afadesigns:
Scoped but deferred — richer `${var:-default}` / `${var:+alt}` / `${var##glob}` parameter-expansion support needs a new `ast.ParameterExpansion` node and matching lexer tokens for the `:-`, `:+`, `:?`, `:=`, `##`, `%%` operators (none are current tokens). Touches AST, parser, lexer, and every kata call-site that inspects dollar-expressions — medium-to-large patch that warrants its own design pass, not a kata PR. Tracking here; next milestone.

afadesigns:
Resolved on main. `${1:-default}` and the related positional-parameter modifier forms (`:=`, `:+`, `:?`, `:-`) parse cleanly via the parameter-expansion modifier-tail walker. Verified with the repro from the report — `./bin/zshellcheck` exits 0 and emits no parser error.

Will ship in the next tagged release. Thanks for the report.

PR #1037, opened 2026-04-20
fix: 3 parser/lexer bugs (#126 elif, #347, #435)

### LINKED ISSUE #126: [BUG] Basic syntax
### Describe the bug

I love the idea of being able to correctly lint my zsh files in my dotfiles repo. 

However, I'm having issues with some basic syntax shown in reproduction code below.

### Reproduction Code

```zsh
alias -- -='cd -'

if [ -x "/opt/homebrew/bin/brew" ]; then
  # Apple Silicon
  eval "$(/opt/homebrew/bin/brew shellenv)"
elif [ -x "/usr/local/bin/brew" ]; then
  # Intel
  eval "$(/usr/local/bin/brew shellenv)"
fi

path=(
  "$XDG_BIN_HOME"
  $path
)
```

### Actual Output

```text
//alias
Parser Error no prefix parse function for = found

// if else
Parser Error no prefix parse function for ELIF found
Parser Error no prefix parse function for THEN found

// path
Parser Error no prefix parse function for = found
```

### Expected Output

No parser errors

### ZShellCheck Version

0.1.1

### Operating System

None

### Additional Context

_No response_

### LINKED ISSUE #347: Parser errors on common Zsh constructs (for...in...do...done, ||)
## Parser crashes on common Zsh constructs

Running the built binary against a real .zshrc surfaces multiple parser panics / "no prefix parse function" errors for tokens that are fundamental to Zsh:

```
Parser Error in ~/.zshrc: no prefix parse function for IN found
Parser Error in ~/.zshrc: no prefix parse function for DO found
Parser Error in ~/.zshrc: no prefix parse function for DONE found
Parser Error in ~/.zshrc: no prefix parse function for || found
```

## Impact

- Any script using `for x in ...; do ...; done` triggers four errors before the linter gives up.
- `||` (logical OR between commands) is universal — parser refuses to progress.
- The binary still reports some violations after the errors, suggesting partial recovery, but downstream passes are running on truncated ASTs. False negatives are guaranteed.

## Reproduction

```bash
go build -o /tmp/zshellcheck ./cmd/zshellcheck
/tmp/zshellcheck ~/.zshrc
```

Against a typical `~/.zshrc` with PATH guards like:

```zsh
for dir in $HOME/.local/bin $HOME/bin; do
    [[ -d "$dir" ]] && path=("$dir" $path)
done
```

## Expected

- `for ... in ... do ... done` parses to a `ForStatement` AST node.
- `||` between commands parses to a `LogicalOr` AST node (or similar).
- No parser errors for these Zsh-core constructs.

## Scope of fix

Likely requires new prefix-parse handlers in `pkg/parser/` for the `IN`, `DO`, `DONE` tokens (if they are surfaced by the lexer as standalone tokens rather than being consumed by the `for` statement handler), and binary-operator handlers for `||`.

## Priority

**High.** The linter cannot be called "production-ready" for v0.3.35 public beta if real-world .zshrc files trip parser errors on the first try.

## References

- `pkg/parser/parser.go`
- `pkg/lexer/lexer.go`
- `pkg/token/token.go`

afadesigns:
Closes #347 #435. Partially closes #126 (elif branch — alias-dash part and remaining #126 sub-bug stay open).

## Lexer (#435)

`dd if=src of=dst` used to raise `no prefix parse function for = found` because the lexer unconditionally promoted `if` / `of` to the `IF` / `OF`-prefix keyword token even when `=` was immediately next. Fix in `pkg/lexer/lexer.go`: after `LookupIdent`, if the matched keyword is immediately followed by `=`, demote back to `IDENT`. Applies to every keyword on the `token.keywords` table.

## Parser (#126 elif branch)

`parseIfStatement` terminated `Consequence` on `ELSE` / `Fi` only — an `elif` clause fell through to the generic `parseStatement` path and errored with `no prefix parse function for ELIF found`. Fix in `pkg/parser/parser_stmt.go`: extend the terminator set to include `ELIF`, and loop while `curToken` is `ELIF` to build a right-nested `IfStatement` chain on `Alternative`. Trailing `else` block attaches to the innermost elif.

## Tests

- `pkg/lexer/lexer_test.go`: `TestNextToken_KeywordFollowedByEquals` exercises `dd if=foo of=bar` and `while=10 do=go`.
- `pkg/parser/parser_test.go`:
  - `TestParser_DDCommandWithIfOfKeywords` — regression for #435 parsed without errors.
  - `TestParser_ForLoopAndOr` — locks in that `for … in … do … done` and `||` chains parse cleanly (issue #347 already fixed; no behaviour change, just a guard-rail).
  - `TestParser_IfElifElseFi` — regression for #126 elif branch.

## Test plan
- [x] `go test -count=1 ./...` green across all packages
- [x] Manual repro: `echo 'dd if=src of=dst' | zshellcheck /dev/stdin` now exits 0

## Outstanding

- #129 parameter-expansion operators (`${var:-default}`, `${var#prefix}`, …) require a new `ast.ParameterExpansion` node and a reshape of `parseArrayAccess`. That's a dedicated PR; tracked in #129.

### DIFF
--- a/pkg/lexer/lexer.go
+++ b/pkg/lexer/lexer.go
@@ -315,6 +315,15 @@ func (l *Lexer) NextToken() token.Token {
 			line, col := l.line, l.column
 			tok.Literal = l.readIdentifier()
 			tok.Type = token.LookupIdent(tok.Literal)
+			// An identifier that happens to match a keyword but is immediately
+			// followed by `=` is a flag/argument assignment (e.g. `if=foo`
+			// inside `dd if=foo of=bar`), not the keyword itself. Demote it
+			// to a plain identifier so the parser treats the following `=`
+			// as part of the same word rather than trying to open an
+			// if-statement.
+			if tok.Type != token.IDENT && l.ch == '=' {
+				tok.Type = token.IDENT
+			}
 			tok.Line = line
 			tok.Column = col
 			tok.HasPrecedingSpace = hasSpace
--- a/pkg/lexer/lexer_test.go
+++ b/pkg/lexer/lexer_test.go
@@ -623,3 +623,61 @@ func TestNextToken_LineAndColumnTracking(t *testing.T) {
 		t.Errorf("second token line expected 2, got %d", tok2.Line)
 	}
 }
+
+// TestNextToken_KeywordFollowedByEquals exercises the regression fix for
+// https://github.com/afadesigns/zshellcheck/issues/435: when an identifier
+// that happens to match a Zsh keyword (`if`, `of`, `while`, `do`, etc.) is
+// immediately followed by `=`, it is a flag-style assignment token (as in
+// `dd if=foo of=bar`), not the keyword. The lexer must return IDENT so the
+// parser treats the whole run as a single word.
+func TestNextToken_KeywordFollowedByEquals(t *testing.T) {
+	cases := []struct {
+		input  string
+		expect []struct {
+			t token.Type
+			l string
+		}
+	}{
+		{
+			input: `dd if=src of=dst`,
+			expect: []struct {
+				t token.Type
+				l string
+			}{
+				{token.IDENT, "dd"},
+				{token.IDENT, "if"},
+				{token.ASSIGN, "="},
+				{token.IDENT, "src"},
+				{token.IDENT, "of"},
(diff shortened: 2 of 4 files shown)


redteamx (approved):
LGTM.

PR #1039, opened 2026-04-20
ci: drop removed `--skip-git` flag from OSV scanner

afadesigns:
OSV-Scanner v2 removed the `--skip-git` flag. The workflow was pre-failing every PR with `flag provided but not defined: -skip-git` before a scan ever started. Drop the flag — `--recursive` alone covers the intended scope.


redteamx (approved):
LGTM.

Issue #127, opened 2026-01-17
[FEAT] Disable Kata per file


agilenut:
### Request Type

Other

### Description

Can we have a way to disable a Kata on a per file basis?

Bonus points for per line or block

### Bad Code Example (for Katas)

```zsh

```

### Good Code Example (for Katas)

```zsh

```

### Rationale

I want to disable the alias -> function kata on a file full of aliases but not on other files.

afadesigns:
Great suggestion! Per-file and per-line kata suppression is a commonly requested feature in linting tools. The plan is to support inline comments like:

```zsh
# zshellcheck disable=ZC1070
alias ls='ls --color=auto'
```

And potentially file-level directives at the top of a file. This is on the roadmap.

PR #1040, opened 2026-04-20
chore: release v1.0.1 — post-milestone hotfix rollup

### LINKED ISSUE #126: [BUG] Basic syntax

### LINKED ISSUE #341: Duplicate detections: ZC1038 and ZC1093 both flag cat overuse

### LINKED ISSUE #346: Severity rebalance: 11 Style katas should be Warning
## Severity rebalance candidates

Audit of 197 Style-level katas reveals 11 that target patterns more dangerous than stylistic preference. These should likely be upgraded to Warning (or Error where runtime breakage is possible).

### Current: Style → Proposed: Warning

| Kata | Title | Reason |
|---|---|---|
| ZC1075 | Quote variable expansions to prevent globbing | Unquoted glob = unpredictable expansion, potential data access |
| ZC1078 | Quote `$@` and `$*` when passing arguments | Argument splitting bugs |
| ZC1079 | Quote RHS of `==` in `[[ ... ]]` | Accidental pattern matching |
| ZC1084 | Quote globs in `find` commands | Shell expansion before find sees the pattern |
| ZC1085 | Quote variable expansions in `for` loops | Word-splitting iteration bugs |
| ZC1090 | Quoted regex pattern in `=~` | Regex vs literal match confusion |
| ZC1136 | Avoid `rm -rf` without safeguard | Irreversible data loss — could argue **Error** |
| ZC1139 | Avoid `source` with URL — use local files | Arbitrary code execution |
| ZC1141 | Avoid `curl \| sh` pattern | Arbitrary code execution |
| ZC1145 | Avoid `tr -d` for character deletion | Minor — leave as Style (listed for completeness) |
| ZC1258 | Consider `rsync --delete` for directory sync | Data deletion side effects |

### Severity rubric reminder

- **Style**: idiomatic preference, no behavioral risk
- **Info**: edge case, works but fragile
- **Warning**: dangerous behavior, data loss possible
- **Error**: code breaks or errors at runtime

The rows above fire on patterns where user intent and actual behavior diverge in ways that can cause data loss, command injection, or confusing runtime bugs. Style understates the risk.

### Current distribution (for context)

- Error: 9 (2.7%)
- Warning: 79 (23.6%)
- Info: 50 (14.9%)
- Style: 197 (58.8%)

If all candidates above move to Warning, the distribution shifts to Error 9 / Warning 90 / Info 50 / Style 186 — still Style-dominant but better calibrated.

### Decision Needed

For each candidate: confirm whether the runtime impact justifies Warning. A PR can adjust the `Severity:` field in each `pkg/katas/zc*.go` file.

### LINKED ISSUE #347: Parser errors on common Zsh constructs (for...in...do...done, ||)

### LINKED ISSUE #435: Parser panic on `dd of=/dev/sda` input

afadesigns:
## Summary

Bundles every fix merged between the v1.0.0 tag and the GitHub Marketplace debut. No new katas — kata count stays at 1000; `scripts/HOTFIX` bumps PATCH to 1.

## Included work
- **feat #127** inline `# zshellcheck disable=…` directives
- **feat #309** XDG Base Directory support
- **fix #435** lexer demotes keyword to IDENT when followed by `=` (unblocks `dd if=src of=dst`)
- **fix #126** `elif` chain parsing
- **fix #347** `/dev/sdX` literals no longer crash
- **fix #341–#345** 14 duplicate katas retired as no-op stubs
- **fix #346** 10 Style katas rebalanced to Warning
- **ci** OSV-Scanner drops removed `--skip-git` flag

## Version bump mechanism
`scripts/update-version.sh` grows a `scripts/HOTFIX` monotonic counter that is added to PATCH. This lets hotfix releases ship between kata additions without colliding with the kata-count formula. When kata #1001 lands, version becomes `1.0.2` (PATCH = (1001 % 100) + 1).

## Test plan
- [x] `go build ./... && go test -count=1 ./...` green locally
- [x] `scripts/update-version.sh` writes `1.0.1` (1000 katas + 1 hotfix)
- [ ] CI required checks (test / security / sbom) green
- [ ] After merge, tag `v1.0.1` on main, draft release via goreleaser


redteamx (approved):
LGTM.

PR #1043, opened 2026-04-20
chore: delete tag-release workflow

afadesigns:
Remove auto-tag-bumping workflow. Prose in commit bodies was accidentally triggering version bumps by matching patch/minor/major keywords. Tags become fully manual.

## Test plan
- [x] No Go changes; workflow deletion only
- [ ] Required CI checks green


redteamx (approved):
LGTM.

PR #1048, opened 2026-04-23
fix: parse bare `$+name[idx]` and `(( )) && (( ))` chains (#1047)

### LINKED ISSUE #1047: [BUG] Unsupported `(( $+commands[ls] )) && (( ... ))`
### Describe the bug

I was updating my zsh config, which makes use of `(( $+commands[...] ))` (sometimes chained) in order to check for the presence of binaries; but `zshellcheck` is raising parsing errors.

It seems there are two separate errors:

1. The use of 'bare' `$+var` seems unsupported, with zshellcheck prefering `${+var}`. While I get that the latter might be prefered in some contexts, I'm using this combined with zsh's `((...))`, and therefore would prefer using `(( $+commands[ls] ))` over `(( ${+commands[ls]} ))`.

2. zshellcheck has issues passing logical sequences of `(( ))`. `(( $+a )) && (( $+b )) || (( $+c ))` causes `no prefix parse function for &&/|| found`

### Reproduction Code

```zsh
#!/usr/bin/env zsh
if (( $+commands[ls] )) && (( $+commands[eza] )); then
  echo "ls and eza are available"
fi
```

### Actual Output

```text
❯ zshellcheck test.zsh

Parser Error in test.zsh: expected next token to be IDENT, got + instead
Parser Error in test.zsh: no prefix parse function for && found
Parser Error in test.zsh: expected next token to be IDENT, got + instead

Found 1 violations.
```

### Expected Output

This should (presumably) work fine?

### ZShellCheck Version

1.0.12 (450bbc22689afcd8cdd1593070ba5e33e34c943a)

### Operating System

macOS

### Additional Context

_No response_

afadesigns:
Closes #1047.

## Bugs

1. **Bare `$+name` / `$+name[key]` inside `(( … ))`** errored with `expected next token to be IDENT, got + instead`. Worked only in braced form `${+name[key]}`.
2. **`(( A )) && (( B ))` chains** (and `||`) errored with `no prefix parse function for && found`. Logical operators between arithmetic commands dropped on the floor.

## Fixes

1. `parser_expr.go` — `parseInvalidArrayAccessPrefix`: when the token after `$` is `+`, consume `+`, read the identifier, and attach optional `[key]` as `InvalidArrayAccess`. Produces the same shape as the working `${+name[key]}` path.
2. `parser_stmt.go` — `parseStatement` `DoubleLparen` case: after the `ArithmeticCommand`, loop on `AND`/`OR` building an `InfixExpression` whose right side comes from `parseCommandPipeline`. Wrap in `ExpressionStatement` so downstream walkers see a normal logical command.

## Tests

`parser_test.go` — three new tests covering bare-form primitives, `&&`/`||`/mixed chains, and the original issue repro inside an `if` block.

## Not in this PR

`${var:-default}` / `${var:=x}` / `${var##glob}` — tracked in #129. Needs a new `ast.ParameterExpansion` node plus lexer tokens for `:-`, `:+`, `:?`, `:=`, `##`, `%%`. Separate design pass.

## Test plan

- [x] `go test -count=1 ./...` green locally
- [x] `/srv/tools/go/bin/golangci-lint run ./...` clean
- [x] Original repro from #1047 parses without errors
- [ ] CI required checks (test / security / sbom) green

### DIFF
--- a/pkg/parser/parser_expr.go
+++ b/pkg/parser/parser_expr.go
@@ -179,6 +179,29 @@ func (p *Parser) parseInvalidArrayAccessPrefix() ast.Expression {
 		return &ast.PrefixExpression{Token: dollarToken, Operator: "$", Right: ident}
 	}
 
+	// `$+name` / `$+name[key]` — parameter-existence test, equivalent to
+	// `${+name}` / `${+name[key]}`. Commonly used inside `(( ... ))`.
+	if p.peekTokenIs(token.PLUS) {
+		p.nextToken() // move to '+'
+		plusToken := p.curToken
+		if !p.expectPeek(token.IDENT) {
+			return nil
+		}
+		ident := &ast.Identifier{Token: p.curToken, Value: p.curToken.Literal}
+		plus := &ast.PrefixExpression{Token: plusToken, Operator: "+", Right: ident}
+		if !p.peekTokenIs(token.LBRACKET) {
+			return &ast.PrefixExpression{Token: dollarToken, Operator: "$", Right: plus}
+		}
+		p.nextToken() // consume [
+		exp := &ast.InvalidArrayAccess{Token: dollarToken, Left: plus}
+		p.nextToken()
+		exp.Index = p.parseExpression(LOWEST)
+		if !p.expectPeek(token.RBRACKET) {
+			return nil
+		}
+		return exp
+	}
+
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1052, opened 2026-04-23
docs: refresh zshellcheck(1) man page

afadesigns:
Pass 4 — man page was stuck on v1.0.0. Bump version header, add the full XDG + home + project config lookup chain, document inline disable directives, fix AUTHOR line.


redteamx (approved):
LGTM.

PR #1055, opened 2026-04-23
docs: restructure README for scannability + signal density

afadesigns:
Halve the README line count (216 -> 122) while keeping every load-bearing pointer. Collapse 8 pointer-only sections (Architecture, Troubleshooting, Developer Guide, Documentation, Changelog, Support, Contributing, Governance) into a single Documentation table. Drop the redundant TOC (GitHub auto-renders a sidebar). Add an 'At a glance' numbers block and a real 'Example output' snippet. Add Marketplace badge. See commit body for the full list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+<div align="center">
+
 ```
  mmmmmm  mmmm  #             ""#    ""#      mmm  #                    #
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1058, opened 2026-04-24
fix: declarations no longer swallow next-line statement

afadesigns:
parseDeclarationStatement over-consumed into the next line when a declaration ended at a line break rather than a semicolon. Caught against zsh-autosuggestions src/async.zsh via the compat harness.

## Root cause

The declaration loop kept calling nextToken after processing each flag/assignment, then checked p.curToken.Line against the declaration start. By the time the check fired, the token pointer was already on the next statement's first token. parseBlockStatement then called its unconditional nextToken(), skipping that token and leaving the parser unable to consume the trailing then/fi of the following if.

## Fix

Guard every internal nextToken in the declaration loop with a peek-line check so the loop leaves curToken on the last token that genuinely belongs to the declaration.

## Tests

Regression test covers typeset, local, readonly, and a value-assignment variant followed by if/then/fi on the next line.


redteamx (approved):
LGTM.

PR #1059, opened 2026-04-24
fix: brace param expansion pre-flags = ~ ^

afadesigns:
parseArrayAccess handled `${(flags)name}` and `${#name}` but missed the single-character pre-flag family Zsh places before the subject without parentheses:

- `${=name}` — word-split
- `${~name}` — glob interpret
- `${^name}` — rc-style expansion

Real code like `strategies=(${=ZSH_AUTOSUGGEST_STRATEGY})` in zsh-autosuggestions used to error with "no prefix parse function for =". Fix: consume these three flag characters before advancing to the subject. Regression tests cover each flag plus the exact snippet from the compat harness.

### DIFF
--- a/pkg/parser/parser_expr.go
+++ b/pkg/parser/parser_expr.go
@@ -146,6 +146,17 @@ func (p *Parser) parseArrayAccess() ast.Expression {
 		hasLengthOp = true
 	}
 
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

PR #1060, opened 2026-04-24
feat: accept parameter-expansion modifiers as opaque

afadesigns:
Enables parser compat with the common Zsh parameter-expansion modifiers without shipping the full structured AST (tracked as #129 for the later design pass).

## What

parseArrayAccess now accepts every modifier family opaquely by walking through the tokens between the subject and the closing brace with brace/paren depth tracking:

- `#`, `##` prefix strip
- `%`, `%%` suffix strip
- `/old/new`, `//all/repl` substitute
- `:-`, `:=`, `:+`, `:?` default / assign-default / alt / error
- `:N`, `:N:M` offset / slice
- `:#glob` pattern removal
- nested composition (`${a:-${b:-c}}`)

The subject stays in the AST — detection katas that only inspect the variable name keep working — only the modifier body is opaque.

## Why

bind.zsh from zsh-autosuggestions (and large chunks of oh-my-zsh / prezto) use these forms pervasively. Before this change the parser bailed on the first modifier it saw, losing all downstream detection on the file.

## Tests

Sixteen regression inputs in parser_test.go cover every modifier family plus a nested composition case.

### DIFF
--- a/pkg/parser/parser_expr.go
+++ b/pkg/parser/parser_expr.go
@@ -175,6 +175,36 @@ func (p *Parser) parseArrayAccess() ast.Expression {
 		}
 	}
 
+	// Modifier tail: `${var#glob}`, `${var##glob}`, `${var%glob}`,
+	// `${var%%glob}`, `${var/pat/repl}`, `${var:-default}`,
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

PR #1061, opened 2026-04-24
feat: array subscript flags (R) (r) (I) (ri)

afadesigns:
Zsh subscript flag tuples `arr[(R)val]`, `arr[(r)pat]`, `arr[(I)i]`, `arr[(ri)pat]` were treated as grouped expressions by parseIndexExpression, leaving the subject token to fire "expected next token to be ]". Fix: skip the flag tuple opaquely with paren-depth tracking before parsing the subject. Against the compat harness this fully clears zsh-autosuggestions bind.zsh and strategies/match_prev_cmd.zsh.

8 regression inputs cover every flag family plus a nested dollar-brace case.

### DIFF
--- a/pkg/parser/parser_expr.go
+++ b/pkg/parser/parser_expr.go
@@ -418,6 +418,28 @@ func (p *Parser) parseIndexExpression(left ast.Expression) ast.Expression {
 
 	p.nextToken()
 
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

PR #1062, opened 2026-04-24
feat: single-char special params and length-of operator

afadesigns:
Two closely-related parser gaps surfaced by the zsh-autosuggestions compat sweep.

**Single-char special parameters**: the lexer only fused a VARIABLE token when dollar was followed by a letter. The remaining single-char specials (? @ $ _) leaked through as two tokens, and the prefix table rejected the tail with "expected next token to be IDENT, got ?". `retval=$?` and `return $?` are ubiquitous.

**Length-of operator**: `$#NAME` was parsed only up to the `#`, leaving NAME as an orphan. `(( $#BUFFER > 0 ))` now parses correctly.

Regression tests cover every form. autosuggestions widgets.zsh 27 → 9 errors; highlight.zsh 3 → 1.

### DIFF
--- a/pkg/lexer/lexer.go
+++ b/pkg/lexer/lexer.go
@@ -263,6 +263,24 @@ func (l *Lexer) NextToken() token.Token {
 				tok.HasPrecedingSpace = hasSpace
 				return tok
 			}
+			// Zsh single-character special parameters. These are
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1063, opened 2026-04-24
fix: declarations with empty RHS at end of line

afadesigns:
Follow-up to the earlier declaration-over-consume fix. `typeset -g VAR=` with nothing after the equals sign is valid Zsh (sets VAR to the empty string), but the parser read the next-line token as the value. In zsh-autosuggestions start.zsh this swallowed the trailing `fi`.

Guard the ASSIGN / PLUSEQ value path with the same peek-line + not-terminator check that already protects the bare-name exit path. start.zsh now parses clean.


redteamx (approved):
LGTM.

PR #1064, opened 2026-04-24
feat: register PLUSEQ as infix for bare var+=value

afadesigns:
Bare `result+="..."` at statement level fired "no prefix parse function for += found" because PLUSEQ was wired only inside declaration-statement parsing, not as a top-level infix operator like ASSIGN. Widespread in oh-my-zsh themes (bureau, agnoster) and plugins for incrementally building prompt strings and arrays.

Register PLUSEQ at EQUALS precedence with parseInfixExpression, mirroring ASSIGN.


redteamx (approved):
LGTM.

PR #1065, opened 2026-04-24
feat: lex ANSI-C and gettext dollar-quoted strings

afadesigns:
`$.'...'` (ANSI-C) and `$."..."` (gettext) previously lexed as two tokens (DOLLAR + STRING), which the parser rejected on the RHS of an assignment with "expected next token to be IDENT, got STRING". oh-my-zsh themes (darkblood and many more) use this pervasively for prompt strings with embedded escape sequences.

Lex them as a single STRING token; dollar-branch dispatch extracted into a new readDollarToken helper to keep NextToken under the funlen bar. darkblood.zsh-theme goes from 4 errors to 0.


redteamx (approved):
LGTM.

PR #1066, opened 2026-04-24
feat: lex backslash-newline as line continuation

afadesigns:
Unquoted `\<newline>` is Zsh line-continuation: the pair should be skipped so a pipeline split across lines is treated as one logical command. Previously the lexer kept the backslash as ILLEGAL, producing "no prefix parse function for ILLEGAL" on real oh-my-zsh themes (bureau and many more).

Handle the continuation inside skipWhitespace so every token producer benefits without per-site guards. bureau.zsh-theme error count drops from 10 to 5; the rest are unrelated modifier / STRING cases tracked for later iterations.


redteamx (approved):
LGTM.

PR #1067, opened 2026-04-24
feat: ZC1004 Fix — exit to return (with arg-less detection)

afadesigns:
Adds a Fix implementation for ZC1004 that rewrites `exit` -> `return` at the command-name position inside function bodies. Arguments stay byte-identical.

Detection was also extended to catch bare `exit` with no arguments. The parser folds zero-arg command invocations into an Identifier expression rather than a SimpleCommand, so the previous match-only-SimpleCommand check missed that shape. New Identifier arm catches it; the SimpleCommand arm now returns false from Walk to avoid double-counting the Name.

Integration test exercises both forms in the same function body.

### DIFF
--- a/pkg/fix/integration_test.go
+++ b/pkg/fix/integration_test.go
@@ -84,6 +84,26 @@ func TestFixIntegration_NestedKatas_OuterWins(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

PR #1068, opened 2026-04-24
feat: ZC1013 Fix — let NAME=EXPR to (( NAME = EXPR ))

afadesigns:
Adds a Fix implementation for ZC1013 that rewrites `let NAME=EXPR` to `(( NAME = EXPR ))`. The span scan runs from the `let` keyword to the end of the logical line; RHS bytes are preserved so operator precedence and nested arithmetic stay intact. Multi-assignment forms remain on the roadmap.

Integration test covers three common shapes including a nested `$(( … ))` RHS.


redteamx (approved):
LGTM.

PR #1070, opened 2026-04-24
fix: double-bracket logical chain short-circuits

afadesigns:
`[[ ... ]] || return 0` and `[[ ... ]] && cmd` are idiomatic guards. The parser let the generic parseExpression loop treat `||` and `&&` as internal infix operators, swallowing the right-hand command token; with `return`/`break`/`continue` on the right this produced "no prefix parse function for RETURN" on dozens of oh-my-zsh themes and plugins.

Fix: dedicated parseStatement case for LDBRACKET calls parseDoubleBracketExpression directly and runs the result through a shared chainLogical helper extracted from the earlier arithmetic-chain work. `(( ... ))` and `[[ ... ]]` now use one code path.

oldgallois.zsh-theme goes from 1 error to 0; the cascade of RETURN/FI errors across other themes drops sharply in the next sweep.

### DIFF
--- a/pkg/parser/parser_stmt.go
+++ b/pkg/parser/parser_stmt.go
@@ -45,29 +45,38 @@ func (p *Parser) parseStatement() ast.Statement {
 		if cmd == nil {
 			return nil
 		}
-		// Chain with `&&` / `||` when the arithmetic command is followed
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

PR #1071, opened 2026-04-24
fix: register return as prefix for logical-chain RHS

afadesigns:
`cmd || return` and `cmd && return 0` are idiomatic guards used throughout oh-my-zsh themes. When OR/AND got folded into an InfixExpression, the chain RHS reached the expression layer and `return` had no prefix entry, producing "no prefix parse function for RETURN".

Fix: register RETURN as a prefix that wraps the keyword in a SimpleCommand and collects same-line arguments. Top-level `return` keeps its dedicated parseReturnStatement path (parseStatement runs before expression dispatch). Regression tests cover assignment-RHS, arithmetic-chain, and double-bracket-chain shapes.


redteamx (approved):
LGTM.

PR #1072, opened 2026-04-24
fix: brace param expansion with empty subject + modifier

afadesigns:
`${(%):-default}` applies the `(%)` flag to an empty parameter with a default. Subject is legitimately absent; the parser errored with "no prefix parse function for :" on simonoff and other themes that use this idiom for prompt-format escapes.

Guard: after flag-tuple consumption, if peek is a modifier punctuator (`:`/`#`/`%`/`/`), skip the subject parse entirely and fall through to the opaque modifier-tail scanner.


redteamx (approved):
LGTM.

PR #1073, opened 2026-04-24
feat: for NAME ( items ) body short form

afadesigns:
Zsh supports a paren-delimited short form of the for-each loop that uses no do/done keywords. Prezto init.zsh uses it: `for mod ("$zmodules[@]") zmodload "zsh/${mod}"`. Previously the parser expected DO after the loop variable and errored with "expected next token to be DO, got (".

Add a third branch to parseForLoopStatement: consume the paren-wrapped item list, then parse the body as a single statement on the same line (or a braced block). Non-block bodies are wrapped in BlockStatement so the Body field stays homogeneous.

Regression tests cover positional, array-splat, glob, and braced-block bodies. prezto/init.zsh error count drops from 9 to 3.


redteamx (approved):
LGTM.

PR #1074, opened 2026-04-24
fix: bare dollar at statement boundary parses as identifier

afadesigns:
`echo $` prints a literal dollar; the oh-my-zsh tjkirch theme uses this pattern in its prompt-char branch. The parser errored with "expected next token to be IDENT, got ;" because parseInvalidArrayAccessPrefix assumed an identifier always follows `$`.

Return an Identifier with Value="$" when the next token is a statement terminator (; EOF | & && || ) } ), matching how the shell treats a bare dollar as a literal character.


redteamx (approved):
LGTM.

PR #1075, opened 2026-04-24
fix: only fuse ]] when a [[ is open

afadesigns:
The lexer unconditionally fused two consecutive `]` into RDBRACKET. That hid the outer bracket in nested array subscripts like `FG[arr[i]]` — the parser errored with "expected next token to be ], got ]]".

Fix: track `[[ … ]]` depth in the lexer and only fuse `]]` when depth is positive; otherwise emit two RBRACKETs. Conditional expressions still close correctly; array-of-array subscripts now parse.

Prezto modules/spectrum/init.zsh parses cleanly.


redteamx (approved):
LGTM.

PR #1076, opened 2026-04-24
fix: route zero-arg commands with pipe chain to simple-command path

afadesigns:
A zero-argument command followed by a pipe or logical chain (`cmd |`, `cmd &&`, `cmd ||`) fell into the expression path because the IDENT dispatch only routed to parseSimpleCommandStatement when the peek was a command-word producer. The expression path returned just the Identifier, leaving the block loop to start a fresh statement at the pipe operator with "no prefix parse function for | found".

Add PIPE, AND, OR to the peek set so zero-arg pipelines and guard chains route through the command layer. Line-continued pipelines `cmd |<NL>rest` now parse as expected.


redteamx (approved):
LGTM.

PR #1077, opened 2026-04-24
feat: lex backslash-escaped metacharacters as IDENT

afadesigns:
Zsh glob patterns and shell-word quoting use a backslash to quote a single following metacharacter: `\(`, `\)`, `\*`, `\?`, `\[`, `\]`, `\|`, `\$`, etc. The lexer emitted ILLEGAL on the backslash which broke oh-my-zsh themes like af-magic that test paths with patterns such as `*\(foo\)*`.

Recognise the common set of backslash-quoted metacharacters and emit the pair as a single IDENT token so parseCommandWord folds it into the surrounding word. Backslash followed by any other byte keeps the ILLEGAL path so malformed input is still reported.

Global corpus error count drops from 2090 to 2059.

### DIFF
--- a/pkg/lexer/lexer.go
+++ b/pkg/lexer/lexer.go
@@ -327,6 +327,32 @@ func (l *Lexer) NextToken() token.Token {
 			tok.Column = col
 			tok.HasPrecedingSpace = hasSpace
 			return tok
+		case l.ch == '\\':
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

PR #1078, opened 2026-04-24
feat: widen backslash-escape metacharacter set

afadesigns:
Extend the backslash-quoting lexer path to also accept `\/`, `\.`, `\!`, `\~`, `\^` — commonly-escaped characters that surfaced as ILLEGAL in oh-my-zsh (e.g. `refs\/remotes\/` inside parameter substitution) and prezto libraries.

Global corpus error count drops from 2059 to 2053.


redteamx (approved):
LGTM.

PR #1079, opened 2026-04-24
feat: brace expansion inside declaration names

afadesigns:
Zsh brace-expansion inside an identifier name — `typeset -g X_{A,B,C}_Y=1` — generates three names from one declaration. Powerlevel10k config files lean on this pattern heavily; prezto does the same for category-scoped variables.

parseDeclarationStatement used to consume only the leading IDENT, break its loop on LBRACE, and cascade into "no prefix parse function for ," errors for every comma inside the expansion.

Add a look-ahead: when peek is LBRACE with no preceding space, consume the matching brace-depth span as part of the name, then absorb any trailing IDENT suffix attached without whitespace.

Global corpus error count drops from 2053 to 1902 — 151 errors resolved.


redteamx (approved):
LGTM.

PR #1080, opened 2026-04-24
fix: plain single quotes do not honour backslash escapes

afadesigns:
Zsh single-quoted strings are raw — `\` is a literal backslash and the only way to embed `'` is to close, write `\'`, and reopen. readString previously consumed `\X` as a two-byte pair unconditionally, which is correct for double quotes and ANSI-C `$'…'` but wrong for plain singles.

Factor the body into readStringFlavour with an explicit honourEscapes flag. Double-quote and $-quote paths pass true; plain single quotes pass false. Fixes long-range corruption that surfaced deep inside Powerlevel10k.


redteamx (approved):
LGTM.

PR #1081, opened 2026-04-24
feat: DOT as expression prefix for literal words

afadesigns:
A bare `.` inside expression context (glob pattern, test condition, path argument) had no prefix parse function. Real code like `[[ "$x" = .* ]]` or `[[ -f ./file ]]` fired "no prefix parse function for .".

Register DOT with parseIdentifier so the dot becomes a literal word. parseCommandWord / bracket scanners fold surrounding tokens in. DOT has no infix entry, so this widens what the parser accepts without changing rejection semantics.

Global corpus error count drops from 1902 to 1877.


redteamx (approved):
LGTM.

PR #1082, opened 2026-04-24
fix: RDBRACKET terminates expression cleanly

afadesigns:
Inside `[[ … ]]`, glob patterns like `.*` or `*.zsh` leave the closing `]]` as the next token after the pattern body. parseInfixExpression would recurse into parseExpression with curToken on RDBRACKET and fire "no prefix parse function for ]]".

Early-return nil from parseExpression when curToken is RDBRACKET so the caller keeps the infix chain it already built, and stop the infix loop when peek is RDBRACKET so the fast path is clean.

Global corpus error count drops from 1877 to 1777.

### DIFF
--- a/pkg/parser/parser_expr.go
+++ b/pkg/parser/parser_expr.go
@@ -9,6 +9,15 @@ import (
 )
 
 func (p *Parser) parseExpression(precedence int) ast.Expression {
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1083, opened 2026-04-24
feat: PERCENT as expression prefix for prompt-escape words

afadesigns:
Zsh theme files use `%F{color}`, `%B`, `%~`, `%n`, `%m`, `%f`, `%b` pervasively as prompt-escape tokens. The lexer emitted PERCENT with no parser prefix, so every prompt-style argument fired "no prefix parse function for %".

Register PERCENT with parseIdentifier; surrounding tokens concatenate into the full escape sequence via parseCommandWord.

Global corpus error count drops from 1777 to 1683.


redteamx (approved):
LGTM.

PR #1084, opened 2026-04-24
feat: composite declaration name shapes

afadesigns:
Extend the declaration name consumer to treat any adjacent non-whitespace token that can form part of a shell word as part of the same logical name: brace-expansion `X_{A,B}_Y`, STRING literals `pre"${1:-}"post`, DollarLbrace expansions `name${suffix}`, VARIABLE references `n$x`, and chained IDENT suffixes.

Fixes cascades in Powerlevel10k gitstatus.plugin.zsh and similar Zsh modules that declare variables with parameter expansion embedded in the identifier.

Global corpus error count drops from 1683 to 1648.


redteamx (approved):
LGTM.

PR #1085, opened 2026-04-24
fix: logical connectors terminate expression inside [[ ]]

afadesigns:
Inside `[[ … ]]`, a glob pattern like `"foo"*` followed by `&&` / `||` made the ASTERISK infix recursion try to parse the connector as a prefix and fire "no prefix parse function for && found".

Extend the RDBRACKET early-return in parseExpression to also bail on AND and OR. Conditional parsing handles the connector at a higher layer.

Global corpus error count drops from 1648 to 1616.


redteamx (approved):
LGTM.

PR #1086, opened 2026-04-24
feat: composite function names

afadesigns:
Extend parseFunctionLiteral to absorb adjacent non-whitespace tokens after the name IDENT so `function query"${1:-}"() { ... }` and `function name$var() { ... }` parse correctly. Gitstatus and similar Zsh modules scope their public API with a caller-provided suffix embedded in the function name.

Global corpus error count drops from 1616 to 1599.


redteamx (approved):
LGTM.

PR #1087, opened 2026-04-24
fix: for (items); do ... done short-form with explicit body

afadesigns:
Zsh accepts both `for x (items) body` and the mixed form `for x (items); do … done`. parseForLoopStatement only handled the pure short form; the mixed form fell into the single-statement path and left `do/done` unparsed, producing "no prefix parse function for DONE".

Check for a leading `;` + DO after the closing RPAREN and route through the classic do/done body when present.


redteamx (approved):
LGTM.

PR #1088, opened 2026-04-24
fix: treat = as literal word in command arguments

afadesigns:
parseCommandWord classifies a small set of metacharacters (ASTERISK, QUESTION, MINUS, …) as literal inside command args even when they have prefix parse entries. ASSIGN and PLUSEQ were missing despite behaving the same way: `env FOO=bar cmd`, `alias x=y`, and similar embeddings should concatenate into the surrounding word rather than trigger an expression recursion.

The declaration parser keeps its own dedicated IDENT=VALUE handling, so `typeset -g VAR=value` still parses as an assignment.


redteamx (approved):
LGTM.

PR #1089, opened 2026-04-24
fix: block-structure keywords terminate an expression cleanly

afadesigns:
Extend the early-return set in parseExpression to include THEN, ELSE, ELIF, FI, DO, DONE, ESAC alongside RDBRACKET, AND, OR.

The common trigger is an empty RHS at end of line:

    if true; then
      VAR=
    else

The ASSIGN infix recurses into parseExpression with curToken on the next-line keyword. Without a prefix entry, the parser fired "no prefix parse function for ELSE / FI / THEN" and cascaded through the rest of the file.

Silent return nil keeps the partial result well-formed so the outer statement parser resumes. The synthetic TestProcessFile_ParserErrors test was reworked to use `) ]` since stray block keywords no longer error.

Global corpus error count drops from 1591 to 1296.


redteamx (approved):
LGTM.

PR #1090, opened 2026-04-24
feat: multi-name function definition

afadesigns:
Zsh `function a b c { body }` declares the same body for a, b, c. oh-my-zsh `lib/prompt_info_functions.zsh` uses this to stub out optional prompt hooks in bulk. parseFunctionLiteral used to bail on the second identifier with "expected next token to be {, got IDENT".

Allow any number of space-separated identifiers after the first. AST keeps the first name for detection purposes.


redteamx (approved):
LGTM.

PR #1091, opened 2026-04-24
feat: let with local / modifier identifier

afadesigns:
Zsh `let local elapsed=1` is valid; the previous parser required the first IDENT to be the target and fired "expected next token to be =, got IDENT" on the second identifier. Real code in oh-my-zsh refined theme uses this.

Skip any IDENT modifiers before pairing the last one with `=`.


redteamx (approved):
LGTM.

PR #1092, opened 2026-04-24
feat: lex force-clobber redirections (>| and >!)

afadesigns:
Zsh `>|file` / `>!file` force-clobber a file even under NO_CLOBBER. The trailing `|` / `!` belongs to the redirection, not a pipeline or negation; the lexer fused them as GT + PIPE / GT + BANG, producing "no prefix parse function for |".

Fuse `>|` and `>!` into a single GT token so the redirection path handles them unchanged. Normal `cmd > file | other` is unaffected because the lexer only fuses when the two characters are adjacent.

Real oh-my-zsh lib/history.zsh uses `>| $HISTFILE` to clobber the history file.


redteamx (approved):
LGTM.

PR #1093, opened 2026-04-24
feat: array slice index arr[a,b]

afadesigns:
Zsh range-slice subscripts `arr[1,8]`, `arr[a,b]`, `arr[1,-1]` fired "expected next token to be ], got ,". Consume the comma and the second index opaquely so the closing bracket lands correctly.

Surfaces in p10k gitstatus.prompt.zsh and every Zsh module that truncates a commit SHA with a slice.


redteamx (approved):
LGTM.

PR #1094, opened 2026-04-24
feat: SLASH as expression prefix for path literals

afadesigns:
A bare `/` argument fired "no prefix parse function for /" in test conditions like `[[ "$dir" != / ]] || return 0` (oh-my-zsh opentofu plugin). Register SLASH with parseIdentifier.

ZC1039 detection updated in lockstep: it matched only StringLiteral args (quoted `/`); bare `/` now lands as an Identifier, so the kata gains an Identifier arm. Existing quoted-and-unquoted tests now exercise both AST shapes.


redteamx (approved):
LGTM.

PR #1095, opened 2026-04-24
fix(lexer): track paren stack to disambiguate `))` fusion

afadesigns:
## Summary
- Lexer now keeps a stack of paren-like openers ('D' for `((`, 'P' for `(`, `$(`, `<(`, `>(`, `=(`).
- `))` fuses into DoubleRparen only when the stack top is 'D'; otherwise each `)` closes its own opener as a single RPAREN.
- Fixes nested constructs like `x=($(cmd))` and `(( x = $((1+1)) + 2 ))` that previously lost track of the outer `((` because the inner `))` was being fused.
- parseDeclarationValue tracks paren/brace depth so nested substitutions inside array literals no longer terminate the scan at the first `)`.

## Impact
Corpus sweep across oh-my-zsh, powerlevel10k, prezto, zsh-autosuggestions, zsh-syntax-highlighting and spaceship-prompt: parser-error total drops from 1290 to 273 (~79% reduction).

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `x=($(echo hello))`, `(( result = $((1+1)) + 2 ))`, `(( x == 1 )) && echo ok` — all parse clean

### DIFF
--- a/pkg/lexer/lexer.go
+++ b/pkg/lexer/lexer.go
@@ -18,6 +18,15 @@ type Lexer struct {
 	// in `arr[$m[i]]`. When depth is zero the lexer emits two
 	// RBRACKETs instead of RDBRACKET.
 	dbracketDepth int
+
+	// parenStack records the kind of every paren-like opener that is
+	// still awaiting its close. 'D' for `((` (arithmetic), 'P' for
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

PR #1096, opened 2026-04-24
fix(lexer/parser): treat `\<NL>` as logical-line continuation

afadesigns:
## Summary
- Token gains `HasPrecedingContinuation` flag; the lexer stamps it on the first token following a `\<NL>` pair.
- Parser adds `peekOnSameLogicalLine()` and all four argument-gather loops switch to it, so continued args like `cmd \<NL>--flag arg` are picked up as arguments of `cmd` rather than orphaned top-level expressions.
- Same pass fixes `readStringFlavour` to honour `${ … }` brace depth inside double-quoted strings so nested quotes in `"${var="default"}"` no longer split the outer string at the inner quote.

## Impact
Corpus sweep: total parser errors 273 → 236 across oh-my-zsh, powerlevel10k, prezto, zsh-autosuggestions, zsh-syntax-highlighting and spaceship-prompt. Spaceship drops 73 → 38 on its own.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: multi-line `cmd \<NL>--color foo \<NL>--prefix bar`, `"${VAR="default"}"`, `*:- default` patterns — parse clean.

### DIFF
--- a/pkg/lexer/lexer.go
+++ b/pkg/lexer/lexer.go
@@ -19,6 +19,14 @@ type Lexer struct {
 	// RBRACKETs instead of RDBRACKET.
 	dbracketDepth int
 
+	// pendingContinuation is set when skipWhitespace has just consumed
+	// a `\<NL>` line-continuation pair. It is read and cleared by
+	// NextToken so the next emitted token carries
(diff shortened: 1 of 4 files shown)


redteamx (approved):
LGTM.

PR #1097, opened 2026-04-24
fix(parser): stop parseBlockStatement from stepping past the terminator

afadesigns:
## Summary
- `parseBlockStatement` advanced unconditionally after every inner statement, overshooting the block terminator when a bare `return` / `break` sat right before `fi` / `done` / `esac`.
- Re-checks whether curToken is already a terminator before the final `nextToken()` and breaks out of the loop if so.

## Impact
Corpus sweep: total parser errors 236 → 161.
- oh-my-zsh: 155 → 85 (-70)
- spaceship-prompt: 38 → 34
- zsh-autosuggestions: 4 → 3

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `if (( x == 1 )); then return; fi`, `for x in a b; do break; done`, `case x in y) return;; esac` — parse clean

### DIFF
--- a/pkg/parser/parser_stmt.go
+++ b/pkg/parser/parser_stmt.go
@@ -478,6 +478,24 @@ func (p *Parser) parseBlockStatement(terminators ...token.Type) *ast.BlockStatem
 			block.Statements = append(block.Statements, stmt)
 		}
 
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1098, opened 2026-04-24
fix(lexer): fuse `&|`/`&!` disown shortcuts and `>=`/`<=` arithmetic comparisons

afadesigns:
## Summary
- Fuse `&|` / `&!` (disown-in-background) into AMPERSAND so the parser treats them as plain background terminators.
- Fuse `>=` / `<=` inside arithmetic into GE_NUM / LE_NUM so the existing LESSGREATER infix handles `(( a >= b ))` / `(( a <= b ))`.
- Factor `<` / `>` handling into `readAngleBracket` to keep `NextToken` under funlen.

## Impact
Corpus sweep: 161 → 159 parser errors. Fixes cases like `(cmd) &|` and `local ran_long=$(( $x >= $y ))` across oh-my-zsh.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `(cmd) &|`, `local x=$(( a >= b ))` — parse clean

### DIFF
--- a/pkg/lexer/lexer.go
+++ b/pkg/lexer/lexer.go
@@ -203,63 +203,9 @@ func (l *Lexer) NextToken() (tok token.Token) {
 	case '*':
 		tok = newToken(token.ASTERISK, l.ch, l.line, l.column)
 	case '<':
-		switch l.peekChar() {
-		case '<':
-			ch := l.ch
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

PR #1099, opened 2026-04-24
fix(lexer/parser): glob alternation + POSIX classes inside `[[ … ]]`

afadesigns:
## Summary
- Lexer tracks a `bracketStack` ('D' = `[[`, 'B' = `[`); `]]` only fuses when the innermost bracket is 'D'. POSIX classes like `[[:alnum:]]` now keep their `]]` separate from the outer conditional's closer.
- `[[` followed immediately by `:` is split into two plain `LBRACKET`s so `[[:alnum:]]` parses as a class, not a nested conditional.
- `parseGroupedExpression` swallows bare `|` between elements; glob alternations like `(wip|WIP)` and `(|*[[:alnum:]])` parse as a list of alternatives.
- New `inDoubleBracket` parser state gates the LPAREN infix so adjacent pattern groups `(a|b)(c|d)` concatenate rather than turning into function calls.

## Impact
Corpus sweep: 159 → 130 parser errors across the full matrix.
- p10k: 13 → 8
- spaceship: 33 → 12
- syntax-highlighting: 12 → 11

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `[[ $x == (wip|WIP) ]]`, `[[ $y == (|*[[:alnum:]])(wip|WIP)(|[[:alnum:]]*) ]]`, `[[ $x == *[[:space:]]* ]]` — parse clean

### DIFF
--- a/pkg/lexer/lexer.go
+++ b/pkg/lexer/lexer.go
@@ -12,12 +12,20 @@ type Lexer struct {
 	line         int  // current line number
 	column       int  // current column number
 
-	// dbracketDepth tracks nesting of `[[ … ]]` conditional blocks.
-	// Without this we cannot distinguish a conditional close (the
-	// fused `]]` token) from two consecutive single-bracket closes
-	// in `arr[$m[i]]`. When depth is zero the lexer emits two
-	// RBRACKETs instead of RDBRACKET.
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1102, opened 2026-04-24
fix(parser): accept Zsh multi-variable `for` loop (`for k v in …`)

afadesigns:
## Summary
Zsh's multi-variable for loop (`for k v in "${(@kv)arr}"; do …`) pairs each element of the item list against N names in turn. parseForLoopStatement only consumed one IDENT and then expected IN / LPAREN / SEMICOLON / DO, so every `for k v in …` produced "expected DO, got IDENT". Skip any run of additional IDENT names after the primary one.

## Impact
114 → 113 parser errors across the matrix.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `for k v in a b c d; do echo $k $v; done` — parses clean


redteamx (approved):
LGTM.

PR #1103, opened 2026-04-24
fix(parser): consume subscript tails opaquely for multi-token keys

afadesigns:
## Summary
Associative-array keys accept arbitrary tokens. Keys like `1st_place_medal` tokenise as INT + IDENT and `parseIndexExpression` only consumed the first piece before `expectPeek(])` fired on the rest. After the arithmetic parse, fall through to an opaque bracket-depth scan if peek isn't `]` yet.

## Impact
113 → 110. oh-my-zsh 70 → 67.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `emoji[1st_place_medal]=x`, `arr[foo-bar]=y` — parse clean


redteamx (approved):
LGTM.

PR #1104, opened 2026-04-24
fix(parser): stop infix loop when curToken lands on `]]`

afadesigns:
## Summary
Patterns ending in ASTERISK inside `[[ … ]]` (e.g. `[[ $file = /* ]] || echo`) caused the outer `parseExpression(LOWEST)` loop to swallow the following `||` / `&&` chain, blowing up `parseDoubleBracketExpression` when it couldn't find its closing token. Add a symmetric `curTokenIs(RDBRACKET)` break alongside the existing peek guard.

## Impact
110 → 101. oh-my-zsh 67 → 59; autosuggestions 3 → 2.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `[[ $x = /* ]] || echo a`, `[[ $x = y* ]] || echo` — parse clean

### DIFF
--- a/pkg/parser/parser_expr.go
+++ b/pkg/parser/parser_expr.go
@@ -46,6 +46,18 @@ func (p *Parser) parseExpression(precedence int) ast.Expression {
 		if p.peekTokenIs(token.RDBRACKET) {
 			break
 		}
+		// When the infix chain has already landed curToken on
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1105, opened 2026-04-24
fix(lexer): fuse compound-assign ops (`%=`, `*=`, `-=`) with PLUSEQ

afadesigns:
## Summary
Only `+=` was fused into PLUSEQ before; `%=`, `*=`, and `-=` split into two tokens. Arithmetic like `(( m %= 60 ))` / `(( x *= 2 ))` / `(( y -= 3 ))` then failed with "no prefix parse function for =". Route all three through PLUSEQ so the parser's existing compound-assign infix handles them uniformly.

## Impact
101 → 97. oh-my-zsh 59 → 56; prezto 10 → 9.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `(( m %= 60 ))`, `(( x *= 2 ))`, `(( y -= 3 ))` — parse clean

### DIFF
--- a/pkg/lexer/lexer.go
+++ b/pkg/lexer/lexer.go
@@ -203,12 +203,20 @@ func (l *Lexer) NextToken() (tok token.Token) {
 			*l = savedLexer
 		}
 
-		if l.peekChar() == '-' {
+		switch l.peekChar() {
+		case '-':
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1106, opened 2026-04-24
fix(parser): consume pipeline tail after a `{ … }` brace group

afadesigns:
## Summary
Brace groups heading a pipeline like `{ cmd } | sort` left `|` dangling for the next parseStatement iteration which reported "no prefix parse function for |". Loop the trailing pipeline / logical continuations and consume each RHS via parseCommandPipeline.

## Impact
97 → 96. oh-my-zsh 56 → 55.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `{ cmd } | sort`, `{ a || b } | awk` — parse clean


redteamx (approved):
LGTM.

PR #1107, opened 2026-04-24
fix(parser): advance past inner `esac` so nested case statements don't collapse

afadesigns:
## Summary
Nested `case` statements misread because parseCaseStatement left curToken on its closing `esac`. When the outer's block terminator list included ESAC the outer parseBlockStatement saw the inner `esac` and broke out, swallowing the rest of the outer clause and reporting "no prefix parse function for ;;" on the outer's own terminator. After the inner loop closes, step off ESAC onto trailing `;;` or `;` if present.

## Impact
96 → 92. oh-my-zsh 55 → 51.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `case x in a) case y in "b") echo b ;; esac ;; esac` — parses clean

### DIFF
--- a/pkg/parser/parser_stmt.go
+++ b/pkg/parser/parser_stmt.go
@@ -605,6 +605,22 @@ func (p *Parser) parseCaseStatement() *ast.CaseStatement {
 			p.nextToken()
 		}
 	}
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1108, opened 2026-04-24
fix(parser): treat tokens without a prefix inside `[[ … ]]` as literals

afadesigns:
## Summary
Inside `[[ … ]]` conditionals tokens without a prefix parse function (`--`, `++`, bare punctuation words) often appear as literal test arguments. Previously every one crashed with "no prefix parse function for <tok>". Emit an Identifier with the literal text when inDoubleBracket is set and no prefix exists.

## Impact
92 → 87. oh-my-zsh 51 → 47; spaceship-prompt 11 → 10.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `[[ $1 == -- ]]`, `[[ $x != ++ ]]` — parse clean


redteamx (approved):
LGTM.

PR #1109, opened 2026-04-24
fix(parser): treat bare `$` before `]` / `]]` as literal dollar

afadesigns:
## Summary
Anchored regex RHS in `[[ … ]]` conditionals ends with `$` (e.g. `[[ $1 =~ ^(a|b).*$ ]]`). The bare-`$` fallback already handled most terminators but not RDBRACKET / RBRACKET. Add them.

## Impact
87 → 85. spaceship-prompt 10 → 8.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `[[ $1 =~ ^(a|b).*$ ]]` — parses clean


redteamx (approved):
LGTM.

PR #1110, opened 2026-04-24
fix(parser): don't treat `[` after a pattern-substitution head as subscript

afadesigns:
## Summary
`${line//[[:space:]]/}` absorbed `line//` as one IDENT (the lexer classes `/` as a letter byte). parseArrayAccess then called parseExpression on the IDENT, its INDEX infix fired on the following `[`, and the subscript parse tore through the POSIX class's `]]` — blowing up the outer conditional. When the subject IDENT already contains `/`, skip the index path and let the modifier tail scanner consume the rest opaquely.

## Impact
85 → 84. spaceship-prompt 8 → 7.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `[[ -n ${line//[[:space:]]/} ]]` — parses clean

### DIFF
--- a/pkg/parser/parser_expr.go
+++ b/pkg/parser/parser_expr.go
@@ -3,6 +3,7 @@ package parser
 import (
 	"fmt"
 	"strconv"
+	"strings"
 
 	"github.com/afadesigns/zshellcheck/pkg/ast"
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1111, opened 2026-04-24
fix(parser): keep mid-word `{…}` refspecs inside a single command argument

afadesigns:
## Summary
Zsh git refspecs `@{upstream}` / `HEAD@{1}` appear bare on command lines. parseCommandWord treated `}` as an unconditional command delimiter, splitting the argument at `upstream` and tripping the enclosing `$(…)`. Track LBRACE depth for braces opened mid-word (not for words that START with `{` — brace expansions `{1..10}` keep their existing shape so ZC1083 still sees the pieces it expects).

## Impact
84 → 81. oh-my-zsh 47 → 45; spaceship-prompt 7 → 6.

## Test plan
- [x] `go test ./...` passes (ZC1083 still passes)
- [x] Manual: `git log @{upstream}.. 2>/dev/null`, `git log HEAD@{1}` — parse clean

### DIFF
--- a/pkg/parser/parser_stmt.go
+++ b/pkg/parser/parser_stmt.go
@@ -364,6 +364,16 @@ func (p *Parser) parseCommandWord() ast.Expression {
 		return p.prefixParseFns[t] != nil
 	}
 
+	// Track LBRACE depth opened MID-WORD so a matching `}` isn't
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1112, opened 2026-04-24
fix(parser): VARIABLE / ${…} expressions can head a pipeline

afadesigns:
## Summary
`$VAR | sed`, `${cmd} && echo` — VARIABLE and DollarLbrace lacked pipeline chaining at the statement level. Route them through the same `parsePipelineStartingWithExpression` helper already used for backticks and `$()`.

## Impact
81 → 80. oh-my-zsh 45 → 44.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `$_var | sed x`, `${cmd} && echo` — parse clean


redteamx (approved):
LGTM.

PR #1113, opened 2026-04-24
fix(lexer): consume heredoc body opaquely after `<<`

afadesigns:
## Summary
Heredocs (`<<EOF ... EOF`, `<<-EOF`, `<<"EOF"`, `<<\EOF`) were lexed as regular source; bodies containing `|`, backticks, braces routinely detonated. After LTLT, peek forward for the delimiter word, then fast-forward past the matching closer line. The parser currently has no heredoc AST node — dropping the body lets real scripts parse cleanly.

## Impact
80 → 75. oh-my-zsh 44 → 39.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: heredocs with `|` in body, `<<-EOF` with tab-stripped delimiter, quoted delimiters — parse clean

### DIFF
--- a/pkg/lexer/lexer.go
+++ b/pkg/lexer/lexer.go
@@ -447,7 +447,9 @@ func (l *Lexer) readAngleBracket(isLeft bool) token.Token {
 	if isLeft {
 		switch peek {
 		case '<':
-			return two(token.LTLT)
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1115, opened 2026-04-24
fix(lexer): accept `\"` and `\'` in the backslash-word escape list

afadesigns:
## Summary
Backslash-escaped quotes outside a string context (e.g. `BUFFER+=\ \"\\\'\\x\"`) emitted ILLEGAL. Add `"` and `'` to the recognised escape list so the pair folds into the surrounding word.

## Impact
72 → 68. powerlevel10k 8 → 7; syntax-highlighting 11 → 8.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `x=\"y`, `x=\'y` — parse clean


redteamx (approved):
LGTM.

PR #1116, opened 2026-04-24
fix(parser): route shell `!` negation through the pipeline parser

afadesigns:
## Summary
`! cmd 2>/dev/null | grep` was parsed as an expression-level prefix `!` around just `cmd`; the trailing redirect and pipe leaked back into parseStatement and blew up with "no prefix parse function for |". When `!` is followed by a command start, route through `parseSimpleCommandStatement` so the pipeline parser's BANG branch handles redirects and pipes. Preserve expression behaviour for `!5` / `!true` used by unit tests.

## Impact
68 → 65. prezto 7 → 5.

## Test plan
- [x] `go test ./...` passes (including `TestParsingPrefixExpressions`)
- [x] Manual: `! cmd 2>/dev/null | grep x`, `!5`, `!true` — all parse


redteamx (approved):
LGTM.

PR #1118, opened 2026-04-24
fix(parser): glued `/` outside arithmetic is a path separator, not division

afadesigns:
## Summary
`HOST=$(cmd)/` followed by a newline and `if` crashed with "no prefix parse function for IF" because SLASH's PRODUCT precedence pulled the next line's keyword in as the division RHS. When peek is SLASH, we're outside arithmetic, and the slash is glued to the previous token, break out of the infix loop.

## Impact
Fixes multiple "no prefix parse function for IF" cascades. Total holds at 64 because downstream errors for those files are different.

## Test plan
- [x] `go test ./...` passes (including `5 / 5` infix test)
- [x] Manual: `HOST=$(cmd)/\nif …` parses

### DIFF
--- a/pkg/parser/parser_expr.go
+++ b/pkg/parser/parser_expr.go
@@ -68,6 +68,16 @@ func (p *Parser) parseExpression(precedence int) ast.Expression {
 		if p.curTokenIs(token.RDBRACKET) {
 			break
 		}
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1119, opened 2026-04-24
fix(parser): consume pipeline tail after block statements (if/for/while/case)

afadesigns:
## Summary
`for f in *; do …; done | column -t`, `while …; do …; done | awk` and friends left the trailing `|` / `&&` / `||` for the next parseStatement iteration, which reported "no prefix parse function for |". Extract `consumePipelineTail` helper and call it after each block-shaped dispatch (if/for/while/case/select).

## Impact
64 → 62. oh-my-zsh 36 → 34.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `for x in *; do echo $x; done | column -t`, `while cond; do body; done | awk` — parse clean

### DIFF
--- a/pkg/parser/parser_stmt.go
+++ b/pkg/parser/parser_stmt.go
@@ -16,18 +16,26 @@ func (p *Parser) parseStatement() ast.Statement {
 	case token.LET:
 		return p.parseLetStatement()
 	case token.If:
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1120, opened 2026-04-24
fix(parser): `${#arr[*]}` subscript close + bare `return;` in blocks

afadesigns:
## Summary
1. `${#arr[*]}` — ASTERISK prefix consumed the `]` as its aborted RHS, leaving curToken on RBRACKET. Short-circuit parseIndexExpression when curToken=RBRACKET AND peek=RBRACE so we don't enter the modifier-tail scanner. Nested `FG[$colors[idx]]` (peek=RBRACKET) keeps the full expectPeek flow.
2. `return;` inside a body threw "no prefix parse function for ;". Add SEMICOLON to parseExpression's early-return list.

## Impact
62 → 60. oh-my-zsh 34 → 32.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `x=${#y[*]}`, `FG[$colors[color+1]]=x`, `if x; then return; fi` — parse clean

### DIFF
--- a/pkg/parser/parser_expr.go
+++ b/pkg/parser/parser_expr.go
@@ -22,7 +22,8 @@ func (p *Parser) parseExpression(precedence int) ast.Expression {
 	switch p.curToken.Type {
 	case token.RDBRACKET, token.AND, token.OR,
 		token.THEN, token.ELSE, token.ELIF, token.Fi,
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1121, opened 2026-04-24
fix(parser): subshell `(for …)`, `(if …)` bodies in command arguments

afadesigns:
## Summary
`cmd (for x in a; do …; done)` — the LPAREN is a subshell argument. parseGroupedExpression parsed the body as an array literal, so FOR/IF/CASE inside the subshell crashed with "no prefix parse function for FOR".

Two changes:
- parseGroupedExpression dispatches statement keywords through parseStatement when the group opens with FOR/WHILE/IF/CASE/DoubleLparen/LDBRACKET/BANG/TYPESET/DECLARE/LET/SELECT/COPROC. LBRACE stays out (brace expansion `{a,b}` is an expression).
- parseExpression stops the infix loop when peek is LPAREN with preceding space so `cmd (subshell)` is a command-with-argument, not a function call on `cmd`.

## Impact
60 → 59. syntax-highlighting 8 → 7.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `time (for x in a; do echo $x; done)`, `x=({a,b})` — both parse clean

### DIFF
--- a/pkg/parser/parser_expr.go
+++ b/pkg/parser/parser_expr.go
@@ -87,6 +87,14 @@ func (p *Parser) parseExpression(precedence int) ast.Expression {
 		if p.inDoubleBracket && p.peekTokenIs(token.LPAREN) {
 			break
 		}
+		// `cmd (subshell)` with space between `cmd` and `(` is a
+		// command followed by a subshell argument, not a function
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1122, opened 2026-04-24
fix(parser): register `++` (INC) as prefix for pre-increment in arithmetic

afadesigns:
## Summary
`(( ++x ))` pre-increment crashed with "no prefix parse function for ++". Register INC as prefix using `parsePrefixExpression`. DEC intentionally stays postfix-only because kata fixtures around `--flag` rely on that — coordinated cleanup required (issue #129).

## Impact
59 → 58. syntax-highlighting 7 → 6.

## Test plan
- [x] `go test ./...` passes (including `--flag` kata fixtures)
- [x] Manual: `(( ++x ))`, `if [[ -z $skip ]]; then (( ++count )); fi` — parse clean


redteamx (approved):
LGTM.

PR #1123, opened 2026-04-24
fix(lexer): don't fuse `((` right after `$(` so `$(((expr)))` parses

afadesigns:
## Summary
`$(((i-1) * 2))` is arithmetic expansion `$((…))` with a nested `(…)` grouping. The lexer fused every `((` into DoubleLparen so after `$(` the next `((` opened a fresh arithmetic block and the outer `))` lost its match.

Add one-shot `suppressLparenFusion` flag: set when DOLLAR_LPAREN emits, consumed by the NEXT LPAREN. A deferred clear in NextToken drops the flag on any non-`$(` token so fresh `((` openers after a non-arithmetic context still fuse normally.

## Impact
58 → 58 (same total). Fixes arithmetic-inner-grouping errors and prevents regressions when `$(cmd)` is followed by `((…))` in the same scope.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `x=$(((i-1)*2))`, `x=$(a); ((y=$x%60))`, `(( a + b ))` — all parse clean

### DIFF
--- a/pkg/lexer/lexer.go
+++ b/pkg/lexer/lexer.go
@@ -27,6 +27,14 @@ type Lexer struct {
 	// outer `]]` unfused.
 	bracketStack []byte
 
+	// suppressLparenFusion is set after emitting DOLLAR_LPAREN (`$(`)
+	// so the next `(` is NOT fused with its peek into DoubleLparen.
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1124, opened 2026-04-24
fix(parser): process substitution accepts multi-statement body

afadesigns:
## Summary
`<( cmd1; cmd2 )` and multi-line `<( cmd1; NL cmd2 )` crashed with "expected ), got ;". `parseCommandList` only handles one pipeline + logical chain. Parse statements in a loop until matching RPAREN instead.

## Impact
58 → 54 parser errors. **zsh-autosuggestions 2 → 0 (first clean run this loop).** prezto 5 → 4; oh-my-zsh 32 → 31.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `builtin exec {FD}< <(cmd1; cmd2)`, `x < <(echo a; echo b)` — parse clean


redteamx (approved):
LGTM.

PR #1125, opened 2026-04-24
fix(parser): `$(a;b)` command substitution drains extra statements to its close

afadesigns:
## Summary
`$(IFS='.' read a b; echo $a)` and similar multi-statement command substitutions crashed with "expected ), got ;". parseCommandList only consumes one pipeline + logical chain. Keep the first command as AST `Command` (katas like ZC1050 walk that shape) then opaquely drain `;`-separated follow-ups until the matching `)`.

## Impact
54 → 53. oh-my-zsh 31 → 30.

## Test plan
- [x] `go test ./...` passes (ZC1050 still detects `$(ls)`)
- [x] Manual: `$(cmd1; cmd2)`, `$(ls)` — parse clean

### DIFF
--- a/pkg/parser/parser_expr.go
+++ b/pkg/parser/parser_expr.go
@@ -558,7 +558,19 @@ func (p *Parser) parseDollarParenExpression() ast.Expression {
 	}
 
 	p.nextToken()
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1128, opened 2026-04-24
fix(parser): add DSEMI to parseExpression early-return

afadesigns:
## Summary
`case x in nn) return ;;` crashed with "no prefix parse function for ;;" because parseReturnStatement called parseExpression with curToken=DSEMI. Add DSEMI to parseExpression's terminator early-return so bare `return` / `break` / `continue` inside a case clause close cleanly at `;;`.

## Impact
49 → 46. oh-my-zsh 29 → 27; spaceship 5 → 4.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `case x in nn) return ;; *) echo ;; esac` — parses clean


redteamx (approved):
LGTM.

PR #1129, opened 2026-04-24
fix(parser): `$(cmd)` / `` `cmd` `` heads in parseSingleCommand

afadesigns:
## Summary
`if x && $(y); then` crashed because parseSingleCommand forced `$(` into an Identifier and the trailing `)` of the substitution leaked back to the dispatch loop. Detect command-producing expression heads (DOLLAR_LPAREN/BACKTICK/VARIABLE/DollarLbrace) and run parseExpression first, then continue with normal argument gathering.

## Impact
46 → 44. spaceship 4 → 3; oh-my-zsh 27 → 26.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `if x && $(y); then z; fi`, `cmd1 || \`f\` arg` — parse clean

### DIFF
--- a/pkg/parser/parser_stmt.go
+++ b/pkg/parser/parser_stmt.go
@@ -360,6 +360,25 @@ func (p *Parser) parseCommandPipeline() ast.Expression {
 }
 
 func (p *Parser) parseSingleCommand() ast.Expression {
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1130, opened 2026-04-24
fix(parser): empty RHS in `${X=}` no longer overshoots the closing brace

afadesigns:
## Summary
`${X=}` (assign-if-unset with empty value) crashed because parseExpression bailed on RBRACE without consuming it, then the modifier-tail scanner walked past EOF looking for `}`. Short-circuit when curToken=RBRACE AND peek≠RBRACE so the outermost expansion's close is honoured. Nested `${#${=name}}` (peek=RBRACE) still goes through expectPeek to advance.

## Impact
44 → 43. spaceship 3 → 2; syntax-highlighting 6 → 5.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `y=${X=}`, `(( x = (y % ${#${=z}}) + 1 ))` — parse clean


redteamx (approved):
LGTM.

PR #1133, opened 2026-04-24
fix(parser): drain pipeline tail after expression-statement (`IDENT= cmd | other`)

afadesigns:
## Summary
Inline env-prefix assignments `LANG=C ss -nat | awk` parsed the infix correctly but left the `|` for the next parseStatement iteration which crashed. Drain consumePipelineTail after parseExpressionStatement so the trailing chain stays attached.

## Impact
41 → 37. **spaceship-prompt 1 → 0 (clean for the first time this loop)**; oh-my-zsh 26 → 24; syntax-highlighting 5 → 4.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `LANG=C ss -nat | awk x`, `X= cmd && other` — parse clean


redteamx (approved):
LGTM.

PR #1135, opened 2026-04-24
fix(parser): parse `${name}` subject narrowly so modifier tail handles patterns

afadesigns:
## Summary
`${a%%[[:space:]]*}` crashed because parseExpression(LOWEST) folded `%%` as infix on `a` and the ASTERISK-prefix path called parseSingleCommand on `[[`. Limit the subject to IDENT (with at most one adjacent subscript), VARIABLE, or INT; modifier tail, flags, anchors flow through the existing opaque scanner.

## Impact
36 → 35. oh-my-zsh 24 → 23.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `${a%%[[:space:]]*}`, `${arr[i]}`, `${var//pat/repl}` — parse clean

### DIFF
--- a/pkg/parser/parser_expr.go
+++ b/pkg/parser/parser_expr.go
@@ -300,17 +300,38 @@ func (p *Parser) parseArrayAccess() ast.Expression {
 		exp.Left = nil
 	} else {
 		p.nextToken() // move to subject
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1136, opened 2026-04-24
fix(parser): treat `*` / `?` / `-` / `!` as literal subject in `${…}`

afadesigns:
## Summary
`${#*:#0}` etc. use single-char punctuation as the subject. The narrow-subject fix in #1135 only matched IDENT/VARIABLE/INT, so punctuation fell into parseExpression and crashed on the modifier punctuator. Add ASTERISK/QUESTION/MINUS/BANG.

## Impact
35 → 34. oh-my-zsh 23 → 22.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `${#*:#0}`, `${?}`, `${-}`, `${!}` — parse clean

### DIFF
--- a/pkg/parser/parser_expr.go
+++ b/pkg/parser/parser_expr.go
@@ -329,7 +329,13 @@ func (p *Parser) parseArrayAccess() ast.Expression {
 					exp.Index = idx.Index
 				}
 			}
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1138, opened 2026-04-24
fix(lexer/parser): track EndLine for multi-line strings + use in same-line check

afadesigns:
## Summary
Multi-line single/double-quoted strings (`'…\n…\n…'`) reported their START line via Token.Line. After such a string, parseSingleCommand's argument loop saw `peek.Line` on the closing-quote line — different from the string's recorded Line — and exited the args loop, orphaning the next argument and producing `expected ), got STRING` cascades inside `$(…)`.

Add Token.EndLine, set by the lexer to the closing-quote line for multi-line strings. peekOnSameLogicalLine consults EndLine.

## Impact
33 → 31. oh-my-zsh 21 → 19.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `b=$(awk 'multi\nline\nscript' file)` — parses clean

### DIFF
--- a/pkg/lexer/lexer.go
+++ b/pkg/lexer/lexer.go
@@ -376,11 +376,13 @@ func (l *Lexer) NextToken() (tok token.Token) {
 		tok.Line = l.line
 		tok.Column = l.column
 		tok.Literal = l.readString('"')
+		tok.EndLine = l.line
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1139, opened 2026-04-24
fix(parser): keyword-headed compound commands inside `$(…)` and pipelines

afadesigns:
## Summary
`$(if [ -d /tmp ]; then find /tmp | grep x ; fi)` crashed because parseCommandPipeline routed IF through parseSingleCommand. Add IF/FOR/CASE/SELECT cases that dispatch through parseStatement; keywordStmtToExpression wraps the result so it flows through pipeline chaining as an expression.

## Impact
31 → 29. oh-my-zsh 19 → 17.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `$(if x; then y; fi)`, `$(for f in *; do echo $f; done)` — parse clean

### DIFF
--- a/pkg/parser/parser_stmt.go
+++ b/pkg/parser/parser_stmt.go
@@ -231,6 +231,25 @@ func (p *Parser) parseExpressionOrFunctionDefinition() ast.Statement {
 	return stmt
 }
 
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1140, opened 2026-04-24
fix(parser): case patterns starting with `(` glob alternation parse correctly

afadesigns:
## Summary
`(darwin|freebsd)*) body ;;` crashed because parseCaseStatement consumed the leading `(` as the optional case-label opener, which desynced the close. Use whitespace as the disambiguator: classic `( pat )` always has space after `(`, glob `(pat|…)` never does. Also accept curToken already on `)` after parseCommandWord absorbs an inline glob group.

## Impact
29 → 27. oh-my-zsh 17 → 16; syntax-highlighting 4 → 3.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `case x in (darwin|freebsd)*) y ;; ( pat ) z ;; esac` — parses clean

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -259,6 +259,20 @@ func (p *Parser) registerInfix(tokenType token.Type, fn infixParseFn) {
 	p.infixParseFns[tokenType] = fn
 }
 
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

PR #1144, opened 2026-04-24
fix(parser): accept `function ${=X} { … }` parameter-expansion as func name

afadesigns:
## Summary
`function ${=X} { … }` declares functions named by split words of the expansion. parseFunctionLiteral expected IDENT and crashed. Skip past the matching `}` of the expansion and treat it as the (opaque) name.

## Impact
25 → 24. oh-my-zsh 14 → 13.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `function ${=NAMES} { body }` — parses clean


redteamx (approved):
LGTM.

PR #1145, opened 2026-04-24
fix(parser): backtick command substitution drains extra statements

afadesigns:
## Summary
`` `cd "$DIR"; pwd` `` crashed because parseCommandSubstitution called parseCommandList which only handles one pipeline + logical chain. Drain `;`-separated extras after the first command, mirroring the fix in `$(…)`.

## Impact
24 → 23. oh-my-zsh 13 → 12.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `` `a; b; c` `` — parses clean


redteamx (approved):
LGTM.

PR #1146, opened 2026-04-24
fix(lexer): stamp VARIABLE token line at start, not after readIdentifier

afadesigns:
## Summary
`readIdentifier` advances past the last identifier byte, so `$name\n` bumped `l.line` before the VARIABLE token was stamped. `for x in $files\ndo` exited the items loop early because `$files` looked like it lived on the `do` line. Capture `l.line` first.

## Impact
23 → 22. oh-my-zsh 12 → 11.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `for f in $files\ndo\necho $f\ndone` — parses clean


redteamx (approved):
LGTM.

PR #1147, opened 2026-04-24
fix(parser): empty-RHS assignment doesn't consume next-statement keyword

afadesigns:
## Summary
`_X=<NL>for conf in …; do` crashed because parseInfixExpression advanced past ASSIGN onto FOR. Add isEmptyRhsTerminator (keywords + closers); when peek is one of those after ASSIGN/PLUSEQ, return partial infix without advancing. Defence-in-depth: parseExpression's curToken early-return also lists the keyword set.

## Impact
22 → 21. oh-my-zsh 11 → 10.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `X=<NL>for f in *; do echo $f; done` — parses clean


redteamx (approved):
LGTM.

PR #1148, opened 2026-04-24
fix(lexer): accept `\<letter>` escapes for regex / glob patterns

afadesigns:
## Summary
`[[ "$x" =~ \<file\>\s* ]]` and similar regex patterns backslash-escape letters (`\s`, `\d`, `\w`, `\b`) that the lexer emitted as ILLEGAL. Add the alphanumeric range to the recognised backslash-escape set. Extracted `readBackslashEscape` helper to keep `NextToken` under the funlen limit.

## Impact
Resolves the regex-pattern crash in oh-my-zsh wp-cli plugin and similar files. Total stays at 21 because the file has another unrelated cascade.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `[[ $x =~ \<file\>\s* ]]`, `\d+`, `\w_x` — parse clean

### DIFF
--- a/pkg/lexer/lexer.go
+++ b/pkg/lexer/lexer.go
@@ -416,39 +416,7 @@ func (l *Lexer) NextToken() (tok token.Token) {
 			tok.HasPrecedingSpace = hasSpace
 			return tok
 		case l.ch == '\\':
-			// Backslash outside a string quotes exactly one following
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1149, opened 2026-04-24
fix(parser): `for x in items <NL> body` shortest body form

afadesigns:
## Summary
Zsh accepts `for VAR in ITEMS\n BODY` (no do/done) — shortest-form loop where body is a single statement on the next line. parseForLoopStatement crashed with "expected DO, got IDENT". When peek isn't DO/SEMICOLON/EOF and is on a different line, parse the next statement as the body.

## Impact
prezto editor: 9 → 1 internal errors; net file count unchanged.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `for k in a b c\n  bindkey k cmd` — parses clean

### DIFF
--- a/pkg/parser/parser_stmt.go
+++ b/pkg/parser/parser_stmt.go
@@ -947,6 +947,23 @@ func (p *Parser) parseForLoopStatement() *ast.ForLoopStatement {
 		return stmt
 	}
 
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1150, opened 2026-04-24
fix(parser): arithmetic concatenated identifier with glued VARIABLE

afadesigns:
## Summary
`(( X_$y == 2 ))` reads as the value of `X_<expanded y>`. parseIdentifier returned just the IDENT and `expectPeek('))')` crashed. When inArithmetic, absorb glued VARIABLE / INT into the identifier value.

## Impact
gitstatus.plugin.zsh internal errors: 26 → 6.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `(( X_$y == 2 ))` — parses clean


redteamx (approved):
LGTM.

PR #1151, opened 2026-04-24
fix(parser): register RBRACKET as prefix for literal `]` arguments

afadesigns:
## Summary
Zsh allows `]` as alias / function name (`alias ]=cat`). RBRACKET had no prefix; statement-position `]` crashed. Register parseIdentifier so the token folds as Identifier.

## Impact
21 → 18. oh-my-zsh 10 → 8; syntax-highlighting 3 → 2.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `alias ]=cat`, `] () { echo; }` — parse clean


redteamx (approved):
LGTM.

PR #1152, opened 2026-04-24
fix(parser): allow `==` pre-flag inside `${==name}` parameter expansion

afadesigns:
## Summary
Zsh `${==X}` strips an outer `=` flag from the expansion (double-equal form). Lexer fuses `==` into EQ. parseArrayAccess pre-flag loop only matched ASSIGN/TILDE/CARET. Add EQ.

## Impact
18 → 17. oh-my-zsh 8 → 7.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `arr=(${==X})` — parses clean


redteamx (approved):
LGTM.

PR #1153, opened 2026-04-24
fix(parser): register COLON as prefix for literal `:` values

afadesigns:
## Summary
`dir=:$GIT_DIR` (literal `:` value before variable), POSIX null-command `:`, and various flag forms need COLON to fold into an Identifier when it appears as the start of an expression. Parser crashed with "no prefix parse function for :".

## Impact
gitstatus.plugin.zsh internal errors: 6 → 3; p10k.zsh: 6 → 5.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `dir=:$X`, `:`, `: arg` — parse clean


redteamx (approved):
LGTM.

PR #1154, opened 2026-04-24
fix(lexer): treat non-ASCII bytes as identifier characters

afadesigns:
## Summary
Theme files and prompt strings routinely contain emoji / accented chars that the lexer emitted as ILLEGAL one byte at a time. p10k internals and oh-my-zsh emoji-char-definitions crashed at every multibyte char. In NextToken default branch, accept any byte >= 0x80 as part of an identifier word and walk the run greedily.

## Impact
oh-my-zsh emoji-char-definitions: 1 → 0 internal errors. p10k.zsh: 5 → 3. Net file count unchanged because file-level errors persist for other unrelated cascades.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `x="🐫 some prompt"`, `emoji[grin]=$'\U1F600'` — parse clean

### DIFF
--- a/pkg/lexer/lexer.go
+++ b/pkg/lexer/lexer.go
@@ -417,6 +417,25 @@ func (l *Lexer) NextToken() (tok token.Token) {
 			return tok
 		case l.ch == '\\':
 			tok = l.readBackslashEscape()
+		case l.ch >= 0x80:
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1155, opened 2026-04-24
fix(parser): register COMMA as prefix for brace-expansion bodies

afadesigns:
## Summary
`function {pp,is,foo}_x() { … }` crashed because parseFunctionLiteral fell through to treating LBRACE as the body opener; parseBlockStatement then crashed on the bare `,`. Register COMMA as a prefix folding into an Identifier so the block parses cleanly.

## Impact
17 → 16. oh-my-zsh 7 → 6.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `function {a,b,c}_x() { echo; }` — parses clean


redteamx (approved):
LGTM.

PR #1156, opened 2026-04-24
fix(parser): support `if cond { body } [else { body }]` brace short form

afadesigns:
## Summary
Zsh accepts `if cond { body } [else { body }]` as a brace-block alternative to `then … fi`. parseIfStatement only matched THEN. Recognise LBRACE as an alternative terminator after the condition; parse the body as a brace block and optionally an `else { body }`.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `if x { echo a } else { echo b }`, `(if x { … })` — parse clean


redteamx (approved):
LGTM.

PR #1157, opened 2026-04-24
fix(parser): support Zsh anonymous-function form `() { body }`

afadesigns:
## Summary
Zsh `() { body }` is an anonymous function. parseStatement routed LPAREN through parseSubshellStatement which crashed when the next token was the closing paren. Detect `()` immediately followed by `{` and parse as anonymous FunctionDefinition.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `() { local x; echo $x }` — parses clean


redteamx (approved):
LGTM.

PR #1158, opened 2026-04-24
feat(zc1003): add Fix that rewrites `[ x -eq y ]` to `(( x == y ))`

afadesigns:
## Summary
ZC1003 detects `[` / `test` numeric comparisons but had no Fix. Add one that rewrites the `[` form in place: opening `[` → `((`, comparison operator (`-eq`/`-ne`/`-lt`/`-le`/`-gt`/`-ge`) → arithmetic equivalent (`==`/`!=`/`<`/`<=`/`>`/`>=`), closing `]` → `))`.

The `test` form stays detection-only because it lacks an explicit closer.

Closes the ZC1003 fix-coverage roadmap entry.

## Test plan
- [x] `go test ./...` passes
- [x] New TestFixIntegration_ZC1003_TestToArith covers if-condition and standalone forms

### DIFF
--- a/pkg/fix/integration_test.go
+++ b/pkg/fix/integration_test.go
@@ -134,6 +134,24 @@ func TestFixIntegration_ZC1004_ExitToReturn(t *testing.T) {
 	}
 }
 
+func TestFixIntegration_ZC1003_TestToArith(t *testing.T) {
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

PR #1160, opened 2026-04-24
feat(cli): print summary footer for `-fix` over directory trees

afadesigns:
## Summary
When `-fix` runs over more than one file (typically a directory tree), append a one-line summary to stderr:

    fix summary: N edit(s) across M file(s) (scanned K)

Single-file invocations stay silent (footer only fires when filesScanned > 1).

Closes the per-directory progress roadmap entry.

## Test plan
- [x] `go test ./...` passes
- [x] Manual: `zshellcheck -fix dir/` shows footer; single-file `-fix` doesn't

### DIFF
--- a/cmd/zshellcheck/main.go
+++ b/cmd/zshellcheck/main.go
@@ -129,10 +129,18 @@ func run() int {
 		fixOpts.enabled = true
 		fixOpts.dryRun = true
 	}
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1164, opened 2026-04-24
fix(parser): bracket-cond glob alts + inner `$()` RPAREN containment

afadesigns:
## Summary

Two parser fixes bundled:

1. **`[[ (a|b) ]]` keywords treated as glob-alt literals** — `parseGroupedExpression` dispatched LPAREN-led keywords (SELECT/CASE/FOR/…) through `parseStatement`, which then demanded `IDENT` after `select`. Inside `[[ … ]]` the group is a glob alternation where those words are literal pattern atoms. Guard the keyword branch with `!p.inDoubleBracket`.
2. **Inner `$(cmd)` RPAREN stops fooling enclosing `( … )` subshell body** — `HOST=$(cmd); echo next` inside a subshell failed because the subshell body treated the `$()` closing `)` as its own terminator. Track via a `consumedParenTerminator` flag; `parseBlockStatement` advances past the inner RPAREN when the flag is set, unless a bigger wrapper like `(( … ))` already absorbed it.

Parser-error total across tracked corpora: 16 → 11.

## Test plan
- [x] `go test ./...` green
- [x] Compat sweep green (documented remaining 11 errors per corpus)

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -90,6 +90,16 @@ type Parser struct {
 	// nextToken so we don't overshoot the following statement's
 	// head. parseBlockStatement clears the flag after honouring it.
 	consumedBraceTerminator bool
+
+	// consumedParenTerminator mirrors consumedBraceTerminator for
+	// `$(cmd)` / `` `cmd` `` endings. When an inner expression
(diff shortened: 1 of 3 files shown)


redteamx (dismissed):
LGTM.

redteamx (approved):
LGTM.

PR #1165, opened 2026-04-24
fix(parser): drain newline-separated statements inside `$(…)`

afadesigns:
## Summary

- `parseDollarParenExpression` drained sibling statements only on `;` — missed plain newline separators
- Multi-line `\$( exec 3>&1; typeset -r dir=\$(mktemp -d); { body } always { cleanup } )` stopped after the first pipeline and `expectPeek(RPAREN)` reported "expected ), got TYPESET"
- Loop until peek reaches `RPAREN` / `EOF`, treating newline (`peek.Line > cur.Line`) as an implicit separator like `;`

Error count holds at 11 but one error shape changed from "got TYPESET" to "got IDENT", meaning an earlier layer now parses correctly and only the inner construct (the `{ body } always { cleanup }` try-block) remains.

## Test plan
- [x] `go test ./...` green
- [x] Compat sweep: no regression, generate.zsh progresses past the TYPESET boundary

### DIFF
--- a/pkg/parser/parser_expr.go
+++ b/pkg/parser/parser_expr.go
@@ -795,8 +795,20 @@ func (p *Parser) parseDollarParenExpression() ast.Expression {
 	// closing `)` cleanly. The AST keeps the first command; katas
 	// that care about the full body can walk source.
 	exp.Command = p.parseCommandList()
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1166, opened 2026-04-24
fix(parser): advance past inner `)` on glob-alt case labels

afadesigns:
## Summary

- `plugin::(disable|enable|load))` and `+(e|U|W|D))` style case patterns absorb the inner `)` of the glob-alt group into `parseCommandWord`; the label terminator is the SECOND `)`
- `parseCaseStatement` previously treated curToken=`)` as "label already at terminator" and skipped `expectPeek`, causing the outer `)` to leak into the clause body and error
- Advance once when curToken=`)` and peek=`)` so we land on the actual label close
- Refactor `$( … )` drain loop to satisfy gocritic's `ifElseChain` rule

Clears `p10k/gitstatus/gitstatus.plugin.zsh`; `omz/lib/cli.zsh` now parses its case patterns and surfaces a downstream issue unrelated to this fix.

## Test plan
- [x] `go test ./...` green
- [x] Compat sweep: 11 → 10 parse errors


redteamx (approved):
LGTM.

PR #1167, opened 2026-04-24
fix(parser): accept INT as for-loop variable name in implicit-list form

afadesigns:
## Summary

- Zsh's `for name...; do ... done` iterates declared names over `\$@` (implicit positional-list form)
- Names may be numeric positional identifiers: `for 1 2 3; do` used in zsh-syntax-highlighting main-highlighter
- The multi-variable name-gathering loop accepted only `IDENT`, leaving `2 3` on the line; `expectPeek(DO)` then fired "expected DO, got INT"
- Extend the loop to also accept `INT`, matching the single-name branch that already allows numeric names

Surfaces previously-masked downstream errors in the same file (if-brace `else`, INT-as-IDENT), which are separate issues to tackle in follow-up iterations.

## Test plan
- [x] `go test ./...` green
- [x] Compat sweep: no regression

### DIFF
--- a/pkg/parser/parser_stmt.go
+++ b/pkg/parser/parser_stmt.go
@@ -955,11 +955,13 @@ func (p *Parser) parseForLoopStatement() *ast.ForLoopStatement {
 
 	// Zsh multi-variable for loop: `for k v in …` / `for a b c in …`
 	// pairs each element of the item list against the declared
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1168, opened 2026-04-24
feat(zc1012): add Fix that inserts `-r` after `read`

afadesigns:
## Summary

- `read` without `-r` interprets backslashes as escape characters — rarely intended
- ZC1012 already flagged bare `read` calls; this lands the matching auto-fix
- Insertion point is the first byte after the `read` command name; existing flags stay in place

Test matrix:
- `read VAR` → `read -r VAR`
- `read -p prompt VAR` → `read -r -p prompt VAR`
- `read -r VAR` stays unchanged (detection doesn't fire)

## Test plan
- [x] `go test ./...` green (three new integration tests)

### DIFF
--- a/pkg/fix/integration_test.go
+++ b/pkg/fix/integration_test.go
@@ -166,6 +166,29 @@ fi
 	}
 }
 
+func TestFixIntegration_ZC1012_ReadAddR(t *testing.T) {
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

PR #1169, opened 2026-04-24
feat(zc1031): add Fix that rewrites `#!/bin/zsh` shebang

afadesigns:
## Summary

- ZC1031 advises switching `#!/bin/zsh` to `#!/usr/bin/env zsh` for PATH portability
- This lands the deterministic auto-fix: single replacement at (line 1, column 1)
- Already-portable shebangs untouched (detection only fires on the bare absolute-path form)

## Test plan
- [x] `go test ./...` green
- [x] Two new integration tests (positive rewrite + idempotent already-portable)


redteamx (approved):
LGTM.

PR #1170, opened 2026-04-24
feat(zc1055): add Fix for empty-string comparisons in brackets

afadesigns:
Rewrites empty-string comparisons inside bracket conditionals to idiomatic zsh forms. Span covers the full infix expression; variable identifier copied from source to preserve original spelling. Detection doesn't fire once rewritten — subsequent fix passes idempotent.

Test plan: go test ./... green, golangci-lint clean, three new integration tests cover right-side empty, right-side non-empty inequality, and left-side empty.


redteamx (approved):
LGTM.

PR #1171, opened 2026-04-24
feat(zc1017): add Fix that inserts raw flag after print

afadesigns:
Mirrors ZC1012's insertion pattern. print without raw flag may interpret backslash escapes; the fix inserts the flag right after the command name, preserving any existing flags. Idempotent on re-run once the flag is present.

Test plan: go test ./... green, golangci-lint clean, two integration tests.


redteamx (approved):
LGTM.

PR #1172, opened 2026-04-24
feat(zc1051): wrap unquoted rm args in double-quotes

afadesigns:
Unquoted $VAR passed to rm triggers globbing — if $VAR is "*" the call deletes every file in the current directory. Fix wraps the arg in double-quotes.

Two edits: insert opening quote at arg start, insert closing quote at arg end. Arg span measured from source with brace / paren / bracket nesting honoured so expansions like ${arr[$i]} and $(cmd) stay whole.

Test plan: go test ./... green, golangci-lint clean, two integration tests (positive rewrite + idempotent already-quoted).

### DIFF
--- a/pkg/fix/integration_test.go
+++ b/pkg/fix/integration_test.go
@@ -243,6 +243,21 @@ func TestFixIntegration_ZC1017_PrintRLeftAlone(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

PR #1173, opened 2026-04-24
feat(zc1062,zc1063): rewrite egrep/fgrep to grep -E/-F

afadesigns:
Two deprecated command aliases with the same shape, fixed in one PR because the rewrite logic is identical. egrep / fgrep are deprecated per upstream grep; the canonical modern forms are grep -E / grep -F. Single-edit command-name replacement.

Test plan: go test ./... green, golangci-lint clean, two new integration tests.


redteamx (approved):
LGTM.

PR #1174, opened 2026-04-24
feat(zc1086): rewrite function keyword form to paren-syntax

afadesigns:
Zsh's function keyword form is non-POSIX. ZC1086 recommends the portable paren-less syntax. The fix deletes the function prefix and, when needed, inserts empty parens after the name to stay consistent with the canonical shape.

Handles both FunctionLiteral (function foo { body }) and FunctionDefinition (function foo() { body }) node shapes.

Test plan: go test ./... green, golangci-lint clean, three integration tests cover bare-body rewrite, explicit-parens rewrite, and plain-form idempotence.

### DIFF
--- a/pkg/fix/integration_test.go
+++ b/pkg/fix/integration_test.go
@@ -274,6 +274,29 @@ func TestFixIntegration_ZC1063_FgrepToGrepF(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

PR #1175, opened 2026-04-24
feat(zc1073): drop unnecessary dollar prefix inside (( ))

afadesigns:
Zsh arithmetic context (( ... )) does name-resolution on bare identifiers — the leading dollar is redundant and noise. Fix deletes the single byte at the violation column. Idempotent once removed.

Test plan: go test ./... green, golangci-lint clean, two integration tests cover single and multiple occurrences.


redteamx (approved):
LGTM.

PR #1176, opened 2026-04-24
feat(zc1064): rewrite type to command -v

afadesigns:
type output format varies across shells; command -v is POSIX and quieter for existence checks. Single-edit command-name replacement.

Test plan: go test ./... green, golangci-lint clean, one integration test.


redteamx (approved):
LGTM.

PR #1177, opened 2026-04-24
feat(zc1061): rewrite seq to brace range expansion

afadesigns:
seq spawns an external process; Zsh brace range {m..n} is built-in. Fix covers three canonical forms with integer literal arguments: single arg N becomes {1..N}, two args M N becomes {M..N}, three args M S N becomes {M..N..S}. Variable args, floats, or flag variants (like -s) are left alone because the rewrite semantics differ.

Test plan: go test ./... green, golangci-lint clean, four integration tests cover each arity plus the variable-arg skip.


redteamx (approved):
LGTM.

PR #1178, opened 2026-04-24
feat(zc1079): quote unquoted RHS in bracket equality tests

afadesigns:
Inside bracket conditionals, unquoted variables on the RHS of equality operators are treated as glob patterns. Quote them to force literal string comparison. Two-edit fix inserts surrounding double quotes around the RHS span, honouring brace / bracket / paren nesting so expansions like array subscripts stay whole.

Yields to ZC1055 when the sibling LHS is an empty string literal — that case is owned by the empty-string rewrite to avoid overlapping edits.

Test plan: tests green, lint clean, two integration tests cover positive rewrite and idempotent already-quoted RHS.

### DIFF
--- a/pkg/fix/integration_test.go
+++ b/pkg/fix/integration_test.go
@@ -352,6 +352,21 @@ func TestFixIntegration_ZC1061_SeqVariableArgsSkipped(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

PR #1179, opened 2026-04-24
feat(zc1085): quote unquoted expansions in for loop items

afadesigns:
Unquoted variable expansions in for loops get split by IFS. Fix wraps each unquoted item in double quotes, honouring brace / bracket / paren nesting via the shared source-span scanner. Three ZC1061 tests updated to reflect the combined multi-kata rewrite (seq arg becomes brace range and the whole substitution gets quoted in one pass).

Test plan: tests green, lint clean, three new integration tests cover variable expansion, array expansion, and already-quoted idempotence.


redteamx (approved):
LGTM.

PR #1180, opened 2026-04-24
feat(zc1084): single-quote unquoted find globs

afadesigns:
find -name *.txt lets the shell expand the glob before find runs — if files match, find receives the file list instead of the pattern. Fix wraps the pattern arg in single quotes so the pattern reaches find verbatim. Span scanner respects character classes and alternations.

Test plan: tests green, lint clean, two integration tests cover positive rewrite and idempotent already-quoted case.


redteamx (approved):
LGTM.

PR #1181, opened 2026-04-24
feat(zc1078): quote $@ / $* when passed as arguments

afadesigns:
Unquoted $@ and $* split arguments by IFS. Quoting preserves the original structure (or joins into one string for $*). Fixed-width two-byte token; two-edit wrap around the position.

Test plan: tests green, lint clean, three integration tests cover $@, $*, and already-quoted idempotence.


redteamx (approved):
LGTM.

PR #1182, opened 2026-04-24
feat(zc1076): insert -Uz after autoload

afadesigns:
autoload without -Uz picks up whatever default the calling environment has. The canonical zsh idiom is autoload -Uz: -U prevents alias expansion in the loaded body and -z selects zsh function style. Fix inserts the flags right after the command name, preserving any trailing arguments.

Test plan: tests green, lint clean, two integration tests cover insertion and idempotence.


redteamx (approved):
LGTM.

PR #1183, opened 2026-04-24
feat(zc1040): append nullglob qualifier to for-loop globs

afadesigns:
for f in *.txt errors in zsh when no files match. Fix appends the (N) qualifier so the glob yields an empty list instead of an error. Span ends at the first unescaped whitespace / delimiter, so *.txt becomes *.txt(N) while quoted or already-qualified patterns stay untouched.

Test plan: tests green, lint clean, two integration tests cover the append and the idempotent already-qualified case.


redteamx (approved):
LGTM.

PR #1184, opened 2026-04-24
feat(zc1091): rewrite bracket numeric compare to arithmetic

afadesigns:
Bracket conditionals with dashed comparison operators should use the arithmetic context. Fix covers the same op map as ZC1003 but applies inside bracket conditionals instead of POSIX test form. Rewrite replaces opener, operator, and closer in three edits.

Yields when more than one comparison op appears inside the brackets — that case is ambiguous to auto-fix.

Test plan: tests green, lint clean, three integration tests cover single-op rewrite, operator variant, and multi-op skip.


redteamx (approved):
LGTM.

PR #1185, opened 2026-04-24
feat(zc1126): collapse sort pipe uniq into sort -u

afadesigns:
sort | uniq spawns two processes when sort -u does the same in one. Fix rewrites the region from just after the sort command name through the end of uniq as a single span, preserving whatever args sit between sort and the pipe.

Detector already guards against uniq flags like -c / -d — those cases stay detection-only because the semantics diverge from sort -u.

Test plan: tests green, lint clean, three integration tests cover args, no-args, and flagged-uniq skip.


redteamx (approved):
LGTM.

PR #1186, opened 2026-04-24
feat(zc1118): rewrite echo no-newline form to print -rn

afadesigns:
echo no-newline flag behaviour varies across shells. print -rn is the reliable Zsh idiom. Fix collapses the echo + no-newline-flag prefix into print -rn as a single span edit; remaining arguments stay intact. ZC1092 already skips flagged echo, so this now picks up the case that was previously detection-only.

One pre-existing test updated to match the combined fix behaviour.

Test plan: tests green, lint clean, one new integration test.


redteamx (approved):
LGTM.

PR #1187, opened 2026-04-24
feat(zc1124): replace cat /dev/null with colon builtin for clearing a file

afadesigns:
cat /dev/null > file spawns an external process just to empty a file. The colon builtin plus redirection does the same without the process overhead. Fix rewrites the cat /dev/null prefix as a single span, keeping the redirection and the rest of the command intact.

Test plan: tests green, lint clean, one integration test covers the canonical shape.


redteamx (approved):
LGTM.

PR #1188, opened 2026-04-24
feat(zc1128): rewrite touch file to empty redirect

afadesigns:
Creating an empty file via touch spawns an external process; a bare redirection does the same without the fork. Single-edit command-name replacement: touch becomes >. Detector already guards flagged forms (timestamp updates) and multi-arg invocations, so the fix covers only the single-file creation case.

Test plan: tests green, lint clean, two integration tests cover the rewrite and the flagged-form skip.


redteamx (approved):
LGTM.

PR #1189, opened 2026-04-24
feat(zc1147): insert -p flag after mkdir for nested paths

afadesigns:
mkdir without -p fails when parent directories are missing. Fix inserts the parent-creating flag after the command name so nested paths create intermediates automatically. Detector already gates on both the absence of -p and the presence of a nested path, so the fix stays narrow.

Test plan: tests green, lint clean, two integration tests cover the nested-path rewrite and the flat-path skip.


redteamx (approved):
LGTM.

PR #1190, opened 2026-04-24
feat(zc1140): rewrite hash cmd to command -v cmd

afadesigns:
hash cmd works as a POSIX existence check but the error messaging is poor. command -v cmd provides cleaner semantics in Zsh. Fix swaps the command name in a single edit; detector already guards flagged forms like hash -r.


redteamx (approved):
LGTM.

PR #1191, opened 2026-04-24
feat(zc1144): rewrite trap numeric signals to canonical names

afadesigns:
Signal numbers vary across platforms — 9 is portable for KILL but much of the range (10, 12, 17+) differs between BSD and Linux. Fix maps POSIX signal numbers 1..31 to their canonical names and rewrites each numeric argument in a trap call. Multiple signals in one invocation all get rewritten.

Test plan: tests green, lint clean, two integration tests cover single and multiple signal rewrites.


redteamx (approved):
LGTM.

PR #1192, opened 2026-04-24
feat(zc1162): rewrite cp recursive flag to archive mode

afadesigns:
cp recursive flag copies contents but drops permissions, timestamps, and symlinks unless combined with additional flags. cp archive mode preserves all attributes. Fix swaps the flag token to keep the command compact.

Test plan: tests green, lint clean, two integration tests cover lowercase and uppercase recursive forms.


redteamx (approved):
LGTM.

PR #1193, opened 2026-04-24
feat(zc1171): rewrite echo escape form to print

afadesigns:
echo escape-flag behaviour varies across shells. print natively interprets escape sequences in Zsh. Fix collapses the command name, whitespace, and flag into a single print token; remaining arguments stay in place.

Test plan: tests green, lint clean, one integration test covers the rewrite.


redteamx (approved):
LGTM.

PR #1194, opened 2026-04-24
feat(zc1135): strip env prefix from inline var assignments

afadesigns:
Zsh handles inline env assignments natively — env VAR=val cmd and VAR=val cmd are equivalent but the latter avoids the env fork. Fix deletes the env prefix plus its trailing whitespace. Detector already guards env flags (-i clean environment, -u unset, etc.), so only the pure inline-prefix form gets rewritten.


redteamx (approved):
LGTM.

PR #1195, opened 2026-04-24
feat(zc1170): insert -q after pushd / popd

afadesigns:
pushd and popd print the directory stack on every call, which clutters script output. Fix inserts -q after the command name to suppress that output. Mirrors the ZC1012 / ZC1017 insertion pattern.

Test plan: tests green, lint clean, two integration tests cover pushd and popd.


redteamx (approved):
LGTM.

PR #1196, opened 2026-04-24
feat(zc1209): insert --no-pager after systemctl

afadesigns:
systemctl invokes a pager by default for status / list-units / list-timers / show, which hangs in non-interactive scripts. Fix inserts --no-pager right after the systemctl command name. Mirrors the ZC1012 / ZC1017 / ZC1170 insertion pattern.


redteamx (approved):
LGTM.

PR #1197, opened 2026-04-24
feat(zc1210): insert --no-pager after journalctl

afadesigns:
journalctl invokes a pager by default which hangs in non-interactive scripts. Fix inserts --no-pager right after the command name. Mirrors ZC1209's systemctl insertion.


redteamx (approved):
LGTM.

PR #1198, opened 2026-04-24
feat(zc1213): insert yes-flag after apt-get for unattended mode

afadesigns:
Package operations via apt-get prompt for confirmation in interactive mode; in a script that hangs indefinitely. Fix inserts the assume-yes flag after the command name. Detector guards the shape (install-class subcommand plus absence of the yes-flag or -qq) so only the problematic forms get rewritten.


redteamx (approved):
LGTM.

PR #1199, opened 2026-04-24
feat(zc1226): insert -T after dmesg for readable timestamps

afadesigns:
dmesg without -T emits raw kernel boot-seconds that are hard to correlate with wall-clock events. Fix inserts the human-readable timestamp flag after the command name. Mirrors other insertion-style fixes.


redteamx (approved):
LGTM.

PR #1200, opened 2026-04-24
feat(zc1227): insert fail flag after curl for HTTP errors

afadesigns:
curl without the fail flag silently returns HTTP error pages (404, 500) as success. Fix inserts the relevant short flag after the command name. Detector already gates on URL arg presence and absence of existing fail-form flags.


redteamx (approved):
LGTM.

PR #1201, opened 2026-04-24
feat(zc1234): insert --rm after docker run subcommand

afadesigns:
docker run without --rm leaves stopped containers on disk. Fix inserts the auto-remove flag right after the run subcommand so subsequent image/args stay intact. Detector already guards shape (docker + run + no --rm + no -d for detached long-running containers).


redteamx (approved):
LGTM.

PR #1202, opened 2026-04-24
feat(zc1231): insert --depth 1 after git clone subcommand

afadesigns:
Full git clones download the entire history, which wastes time and disk in CI or build contexts. Fix inserts shallow flags after the clone subcommand. Detector already guards against existing --depth / --shallow-since / --single-branch flags.


redteamx (approved):
LGTM.

PR #1203, opened 2026-04-24
feat(zc1241): insert null-separator flag after xargs

afadesigns:
xargs without the null-separator flag splits on whitespace, so filenames with spaces or newlines get corrupted. Fix inserts the flag after the command name. Detector gates on 'xargs rm' specifically to narrow the rewrite to the highest-risk shape.


redteamx (approved):
LGTM.

PR #1204, opened 2026-04-24
feat(zc1263): rewrite bare apt to apt-get

afadesigns:
apt is the interactive frontend; its output format is documented as unstable and may drift between releases. apt-get has the scripting contract. Single-edit command-name rename — subcommands and arguments stay intact because both tools accept the same install / upgrade / update / remove syntax.


redteamx (approved):
LGTM.

PR #1205, opened 2026-04-24
feat(zc1264): rewrite yum to dnf

afadesigns:
yum is deprecated on Fedora 22+ and RHEL 8+. dnf has broadly-compatible CLI so arguments carry over unchanged. Single-edit command-name rename.


redteamx (approved):
LGTM.

PR #1206, opened 2026-04-24
feat(zc1253): insert --no-cache after docker build subcommand

afadesigns:
docker build uses layer caching which can mask dependency changes in CI. Fix inserts --no-cache right after the build subcommand to guarantee reproducible builds. Mirrors ZC1234's subcommand-level insertion pattern.


redteamx (approved):
LGTM.

PR #1207, opened 2026-04-24
feat(zc1267): insert -P flag after df for portable output

afadesigns:
df -h output format varies across systems and locales and can split long device names across lines. Fix inserts the POSIX-portable flag after the command name. Detector narrows to df -h so only script-unsafe invocations get rewritten.


redteamx (approved):
LGTM.

PR #1208, opened 2026-04-24
feat(zc1192): replace sleep 0 with colon builtin

afadesigns:
sleep 0 spawns an external process that does nothing. Colon is the zsh builtin no-op equivalent. Fix replaces the span from sleep through the zero arg with a single colon. Detector already gates on exactly one arg equal to zero.


redteamx (approved):
LGTM.

PR #1209, opened 2026-04-24
feat(zc1288): rewrite declare to typeset

afadesigns:
typeset is the native Zsh builtin. declare is a Bash compatibility alias. Fix swaps the keyword at the declaration statement position; flags and assignments carry over.


redteamx (approved):
LGTM.

PR #1210, opened 2026-04-24
feat(zc1283): rewrite set -o OPT to setopt OPT

afadesigns:
set -o is the POSIX compatibility form for toggling shell options. setopt is the native Zsh builtin. Fix collapses the command name plus -o flag into setopt in a single span edit; the trailing option name stays in place.


redteamx (approved):
LGTM.

PR #1211, opened 2026-04-24
feat(zc1265): insert --now after systemctl enable

afadesigns:
systemctl enable without --now only arms the service for next boot. Fix inserts --now right after the enable subcommand so the service starts immediately too. Subcommand-level insertion pattern (same shape as ZC1234 / ZC1253).


redteamx (approved):
LGTM.

PR #1212, opened 2026-04-24
feat(zc1301): rename Bash PIPESTATUS var to Zsh pipestatus

afadesigns:
Bash uses $PIPESTATUS for the pipeline exit-code array; Zsh uses the lowercase $pipestatus. Fix renames the identifier at its source position, covering both the dollar-prefixed form and the bare uppercase form. Array subscripts inside a braced reference aren't reached because the parser wraps those in ArrayAccess — a follow-up can extend detection there.


redteamx (approved):
LGTM.

PR #1213, opened 2026-04-24
feat(zc1307): rename Bash DIRSTACK var to Zsh dirstack

afadesigns:
Bash uses $DIRSTACK for the directory stack array; Zsh uses the lowercase $dirstack. Fix renames the identifier at its source position, covering both the dollar-prefixed form and the bare uppercase form. Same shape as ZC1301's pipestatus rename.


redteamx (approved):
LGTM.

PR #1214, opened 2026-04-24
feat(zc1304): rename BASH_SUBSHELL to ZSH_SUBSHELL

afadesigns:
Bash exposes subshell nesting depth via $BASH_SUBSHELL; Zsh provides $ZSH_SUBSHELL as the native equivalent. Fix renames the identifier at its source position, covering both the dollar-prefixed and bare forms.


redteamx (approved):
LGTM.

PR #1215, opened 2026-04-24
feat(zc1298): rename FUNCNAME to funcstack

afadesigns:
Bash exposes the call stack via $FUNCNAME; Zsh uses $funcstack for the equivalent array. Fix renames the identifier at its source position, covering both the dollar-prefixed and bare forms.


redteamx (approved):
LGTM.

PR #1216, opened 2026-04-24
feat: rename Bash completion COMP_* vars to Zsh equivalents

afadesigns:
Three Bash completion variables have direct Zsh equivalents. One PR covers them all because the rewrite logic is identical — only the name pairs differ.

- COMP_WORDS becomes words (array of words on the command line)
- COMP_CWORD becomes CURRENT (current cursor word index)
- COMP_LINE becomes BUFFER (full command line content)

Each fix handles both the dollar-prefixed and bare forms.

Test plan: tests green, lint clean, three integration tests.


redteamx (approved):
LGTM.

PR #1217, opened 2026-04-24
feat(zc1318): rename BASH_CMDS to commands hash

afadesigns:
Bash caches command-path lookups in $BASH_CMDS; Zsh exposes the equivalent via the $commands hash. Fix renames the identifier at its source position, covering both the dollar-prefixed and bare forms.


redteamx (approved):
LGTM.

PR #1218, opened 2026-04-24
feat(zc1313): rename BASH_ALIASES to aliases hash

afadesigns:
Bash exposes defined aliases via $BASH_ALIASES; Zsh uses the $aliases associative array. Fix renames the identifier at its source position, covering both the dollar-prefixed and bare forms.


redteamx (approved):
LGTM.

PR #1219, opened 2026-04-24
feat(zc1300): rename BASH_VERSION / BASH_VERSINFO to ZSH_VERSION

afadesigns:
Bash exposes shell version via $BASH_VERSION (string) and $BASH_VERSINFO (array). Zsh exposes it via $ZSH_VERSION (string). Fix renames either Bash form to the Zsh equivalent. The array-to-string conversion is lossy; callers that need components can split the result with ${(s:.:)ZSH_VERSION}.


redteamx (approved):
LGTM.

PR #1220, opened 2026-04-24
feat(zc1331): rename BASH_REMATCH to match array

afadesigns:
Bash holds regex captures in $BASH_REMATCH; Zsh exposes them via $match (array of captures) and $MATCH (full match). Fix renames the identifier at its source position, covering both the dollar-prefixed and bare forms.


redteamx (approved):
LGTM.

PR #1221, opened 2026-04-24
feat(zc1333): rename TIMEFORMAT to TIMEFMT

afadesigns:
Bash uses $TIMEFORMAT to customise time output; Zsh uses $TIMEFMT. Fix renames the identifier at its source position, covering both the dollar-prefixed and bare forms. Format-specifier dialect differs between the two shells, so callers should still review the format string after conversion.


redteamx (approved):
LGTM.

PR #1222, opened 2026-04-24
feat(zc1356): uppercase read array flag for Zsh semantics

afadesigns:
Bash uses lowercase array flag with read; Zsh uses uppercase for the same intent. Bash's lowercase variant in Zsh actually assigns a shell flag to a scalar variable, which is almost never the author's intent. Fix swaps the lowercase arg to uppercase at its source position.

Test plan: tests green, lint clean, one integration test. ZC1012's raw-flag insertion fires on the same invocation so the combined rewrite adds the raw flag too — test asserts the combined output.


redteamx (approved):
LGTM.

PR #1223, opened 2026-04-24
feat(zc1355): collapse echo raw form to print -r

afadesigns:
echo with the uppercase raw flag is a Bash-ism that POSIX echo ignores. print -r is the reliable Zsh raw-printer. Fix collapses the command name plus flag into the Zsh form as a single span edit; remaining arguments stay in place.


redteamx (approved):
LGTM.

PR #1227, opened 2026-04-24
fix(zc1068): guard nil Name on anonymous function definitions

### LINKED ISSUE #1225: Panic in ZC1068: nil Name on FunctionDefinition / FunctionLiteral
```
  pkg/katas/zc1068.go:27  checkZC1068
```

## Reproduction

Run zshellcheck against any file with an anonymous function definition. Found via integration scan against the `antidote` repo (`antidote.zsh`).

## Root cause

`zc1068.go` dereferences `n.Name.Value` on both the `FunctionDefinition` and `FunctionLiteral` branches without a nil guard. Anonymous function nodes have `Name == nil`.

## Expected

Return `nil` from Check when the node has no name (the hook-name match cannot fire).

## Impact

Hard crash of the linter on any file containing an anonymous function. Breaks integration runs against antidote, likely others.

afadesigns:
Closes #1225.

Anonymous function nodes (`() { … }`) have `Name == nil`. `checkZC1068` dereferenced `n.Name.Value` without a nil check, hard-crashing the whole linter. Surfaced via integration scan against the antidote repo.

Fix: return early from both switch branches when Name is nil.

Regression test covers the shape.

### DIFF
--- a/pkg/katas/katatests/zc1068_test.go
+++ b/pkg/katas/katatests/zc1068_test.go
@@ -50,6 +50,12 @@ func TestZC1068(t *testing.T) {
 				},
 			},
 		},
+		{
+			// Regression for #1225 — anonymous function must not panic.
+			name:     "anonymous function is safe",
+			input:    `() { echo hi } "$@"`,
+			expected: []katas.Violation{},
+		},
 	}
 
 	for _, tt := range tests {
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

PR #1228, opened 2026-04-24
fix(zc1070): guard nil Name / Body / cmd.Name during body walk

### LINKED ISSUE #1226: Panic in ZC1070: nil Name / Body / cmd.Name during function-body walk
```
  pkg/ast/ast.go:109  (*Identifier).String
  pkg/katas/zc1070.go:25  checkZC1070
```

## Reproduction

Scan a file with an anonymous function, or a function whose body contains a `SimpleCommand` with nil `Name`. Found via integration scan against `antidote/antidote.zsh` after guarding ZC1068.

## Root cause

Three un-guarded dereferences in `zc1070.go`:

- Line 25: `funcDef.Name.String()` — nil Name on anonymous function defs.
- Line 44: `ast.Walk(funcDef.Body, …)` — nil Body on stub declarations.
- Line 54: `cmd.Name.String()` — nil Name on some SimpleCommand shapes.

## Expected

Return `nil` from Check when Name / Body is nil. Inside the walker, skip SimpleCommand nodes whose Name is nil.

## Impact

Hard crash on scripts that combine anonymous functions with a recursive-wrapper-candidate name, or on malformed bodies the parser recovered from.

afadesigns:
Closes #1226.

Three separate un-guarded dereferences in the ZC1070 recursive-wrapper check, all reached via anonymous functions and malformed bodies the parser recovered from:

- `funcDef.Name.String()` when Name is nil.
- `ast.Walk(funcDef.Body, …)` when Body is nil.
- `cmd.Name.String()` when a SimpleCommand has nil Name.

Each path now returns early (or skips the subtree) instead of panicking. Surfaced via integration scan against antidote.zsh after #1227 fixed the preceding panic in ZC1068.

Regression test covers the anonymous-function shape.

### DIFF
--- a/pkg/katas/katatests/zc1070_test.go
+++ b/pkg/katas/katatests/zc1070_test.go
@@ -79,6 +79,13 @@ func TestZC1070(t *testing.T) {
 			input:    `fib() { fib $(($1-1)); }`,
 			expected: []katas.Violation{}, // Should NOT warn for generic recursion?
 		},
+		{
+			// Regression for #1226 — anonymous function (nil Name) must
+			// not panic even when the check walks its body.
+			name:     "anonymous function is safe",
+			input:    `() { cd /tmp } "$@"`,
+			expected: []katas.Violation{},
+		},
 	}
 
 	for _, tt := range tests {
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

PR #1231, opened 2026-04-24
fix(zc1043): guard nil assign.Right in local-hint message

### LINKED ISSUE #1229: Panic in ZC1043: nil assign.Right dereferenced during local-hint message build
```
  pkg/katas/zc1043.go:65  checkZC1043.func1
```

## Reproduction

Scan a file with an assignment whose right-hand side is nil (e.g. `VAR=` at end of line, empty RHS the parser accepts). Surfaced via integration scan against `antidote.zsh` after #1225 and #1226 cleared.

## Root cause

Line 65 calls `assign.Right.String()` inside the message builder without a nil check. The empty-RHS assignment form is valid Zsh and the parser records it with `Right == nil`.

## Expected

Either skip the violation when Right is nil, or fall back to an empty string in the message template.

## Impact

Hard crash on scripts that set a variable to the empty value via bare `VAR=`.

afadesigns:
Closes #1229.

Empty-RHS assignments (`VAR=` at end of line) are valid Zsh and the parser records them with `Right == nil`. The violation message template dereferenced `assign.Right.String()` without a nil check, crashing the linter. Surfaced via integration scan against antidote.zsh.

Fallback to an empty string so the hint still renders as `local VAR=` instead of panicking.

Regression test added.

### DIFF
--- a/pkg/katas/katatests/zc1043_test.go
+++ b/pkg/katas/katatests/zc1043_test.go
@@ -30,6 +30,22 @@ func TestZC1043(t *testing.T) {
 			input:    "myfunc() { local x=1; }",
 			expected: []katas.Violation{},
 		},
+		{
+			// Regression for #1229 — empty-RHS assignment must not
+			// panic during message build. Hint still emitted with an
+			// empty RHS rendered in the template.
+			name:  "empty-RHS assignment does not panic",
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

PR #1236, opened 2026-04-24
fix(ast): nil-guard every Node.String() that embeds a child

### LINKED ISSUE #1235: Panic in ProcessSubstitution.String: nil Command dereferenced
```
  pkg/ast/ast.go:906  (*ProcessSubstitution).String
  pkg/katas/zc1054.go:68  getStringValueZC1054 (fallback to node.String())
  pkg/katas/zc1054.go:31  checkZC1054
```

## Reproduction

Scan `testdata/external-corpora/autosuggestions/zsh-autosuggestions.zsh`, `testdata/external-corpora/autosuggestions/src/async.zsh`, or `testdata/external-corpora/p10k/internal/worker.zsh`. Three files, same stack.

## Root cause

`(*ProcessSubstitution).String()` calls `ps.Command.String()` without a nil check. When the parser recovers from a malformed `<(…)` / `>(…)` / `=(…)` body it can leave `Command == nil` on the node.

The panic only surfaces now because ZC1054 falls back to `node.String()` in its catch-all arm, but any walker that calls `String()` on the bad node would hit the same crash.

## Fix

```go
func (ps *ProcessSubstitution) String() string {
    cmd := ""
    if ps.Command != nil {
        cmd = ps.Command.String()
    }
    return ps.Token.Literal + cmd + ")"
}
```

## Impact

Hard crash on autosuggestions + p10k worker. Likely the same bug lurks in `(*Subshell).String`, `(*DollarParenExpression).String`, and peers — audit as a follow-up.

afadesigns:
Closes #1235.

`ProcessSubstitution.String()` panicked when the parser recovered from a malformed substitution and left `Command == nil`. Audit of `pkg/ast/ast.go` surfaced 37 similar derefs — every `String()` that embeds a child node's output was vulnerable.

## Change

Added a `nodeString(n Node) string` helper:

```go
func nodeString(n Node) string {
    if n == nil { return "" }
    return n.String()
}
```

Routed every embedded child `.String()` call through it. Node types covered: LetStatement, ReturnStatement, ExpressionStatement, PrefixExpression, PostfixExpression, InfixExpression, BlockStatement, ForLoopStatement, IndexExpression, GroupedExpression, ArrayAccess, CommandSubstitution, InvalidArrayAccess, DollarParenExpression, SimpleCommand, SelectStatement, CaseStatement, Subshell, ProcessSubstitution, DeclarationStatement, ArithmeticCommand, Redirection, FunctionDefinition.

Identifier, Boolean, IntegerLiteral, StringLiteral, FloatLiteral stay as-is (they read scalar fields, not Node references).

## Integration-scan impact

Clears panics on autosuggestions / zsh-autosuggestions.zsh, autosuggestions / src / async.zsh, p10k / internal / worker.zsh.

## Follow-up tracked separately

The real bug is that the parser's error-recovery path sometimes hands out Node fields set to nil. This PR makes the tree safe to stringify today; auditing the parser recovery paths is a separate task.

## Test plan

- [x] `go test ./...` green
- [x] Integration scan on previously-crashing files now yields normal violations.

### DIFF
--- a/pkg/ast/ast.go
+++ b/pkg/ast/ast.go
@@ -7,6 +7,18 @@ import (
 	"github.com/afadesigns/zshellcheck/pkg/token"
 )
 
+// nodeString returns n.String() when n is non-nil, or an empty string
+// when n is nil. Every Node.String() that embeds a child-node's output
+// routes through here so a partially-parsed tree (where the parser
+// recovered from an error and left a child field unset) can stringify
+// without panicking.
+func nodeString(n Node) string {
+	if n == nil {
+		return ""
+	}
+	return n.String()
+}
+
 // Node represents a node in the AST.
 type Node interface {
 	TokenLiteral() string // literal value of the node
@@ -57,7 +69,7 @@ func (ls *LetStatement) TokenLiteralNode() token.Token { return ls.Token }
 
 // String returns a string representation of the LetStatement.
 func (ls *LetStatement) String() string {
-	return ls.Token.Literal + " " + ls.Name.String() + " = " + ls.Value.String() + ";"
+	return ls.Token.Literal + " " + nodeString(ls.Name) + " = " + nodeString(ls.Value) + ";"
 }
 
 // ReturnStatement represents a return statement.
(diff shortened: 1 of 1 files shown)


redteamx (approved):
LGTM.

PR #1234, opened 2026-04-24
fix: guard typed-nil *ast.Identifier in 32 katas

afadesigns:
Closes #1233, #1232.

Every kata whose Check starts with `ident, ok := node.(*ast.Identifier)` panicked on a typed-nil interface: the type assertion returned `ok=true, ident=nil` and the subsequent `ident.Value` read landed on a nil receiver.

## Katas patched

ZC1122, ZC1191, ZC1297, ZC1298, ZC1299, ZC1300, ZC1301, ZC1304, ZC1305, ZC1306, ZC1307, ZC1308, ZC1309, ZC1310, ZC1313, ZC1314, ZC1315, ZC1317, ZC1318, ZC1319, ZC1320, ZC1321, ZC1322, ZC1324, ZC1325, ZC1326, ZC1328, ZC1329, ZC1330, ZC1331, ZC1332, ZC1333.

## Two shapes patched

**Single-line guard:**

```go
// before
if !ok || ident.Value != "…" {
// after
if !ok || ident == nil || ident.Value != "…" {
```

**Split guard:**

```go
// before
ident, ok := node.(*ast.Identifier)
if !ok {
    return nil
}
// … later …
if ident.Value != "…" {
// after — inserted immediately after the !ok block
if ident == nil {
    return nil
}
```

## Integration-scan impact

Clears panics on `omz/plugins/jsontools`, `omz/plugins/keychain`, `omz/plugins/virtualenvwrapper`, `omz/themes/jnrowe.zsh-theme`, plus antidote test fixtures. ZC1054 / ZC1004 / ZC1191 follow-ups still outstanding — tracked separately.

## Test plan

- [x] `go test ./...` green
- [x] `golangci-lint run ./...` clean
- [x] Post-patch integration scan on `omz/plugins/jsontools` yields normal violations instead of panic.

PR #1238, opened 2026-04-24
fix(ast): reject typed-nil Node in Walk to stop downstream panics

### LINKED ISSUE #1237: Walk dispatches typed-nil pointers to kata Checks, causing panics
## Root cause

`Walk(node, f)` guards only against interface-nil (`node == nil`). Go treats a typed-nil pointer inside an interface as non-nil at the interface level, so Walk descends into it, the switch arm picks the concrete type, and every downstream kata that reads a field on the nil concrete pointer panics.

```
goroutine 1 [running]:
github.com/afadesigns/zshellcheck/pkg/katas.checkZC1004.func1({0x7801a8?, 0x0?})  ← typed-nil (*ast.Identifier)
  pkg/katas/zc1004.go:85
github.com/afadesigns/zshellcheck/pkg/ast.Walk({0x7801a8, 0x0}, 0xc00018d098)
  pkg/ast/ast.go:653
```

## Reproduction

Scan `testdata/external-corpora/prezto/init.zsh` — ZC1004 crashes on a typed-nil `*ast.Identifier` that Walk hands it.

## Fix

Early-return at the top of Walk when the interface carries a typed-nil pointer:

```go
if node == nil {
    return
}
if v := reflect.ValueOf(node); v.Kind() == reflect.Ptr && v.IsNil() {
    return
}
```

This closes the class of bug, not just ZC1004's instance. Individual katas remain responsible for their own nil-safety on sub-node fields, but the walker stops handing them a nil to deref.

## Impact

Hard crash of the linter on any file whose AST contains a typed-nil Node reference — surfaces via prezto's init.zsh today; the PR #1234 / #1236 fixes only masked the cluster by hardening each kata. This is the root fix.

afadesigns:
Closes #1237.

Walk's existing `node == nil` check only rejected untyped interface-nil. A Node interface value that holds a concrete pointer type with a nil underlying pointer passes the check, descends into the type switch, and the downstream kata switch arm picks it up as the concrete type. Every Check that reads a field on that nil receiver panics.

## Fix

Add a `reflect.ValueOf(node).IsNil()` check immediately after the interface-nil guard at the top of Walk:

```go
if v := reflect.ValueOf(node); v.Kind() == reflect.Ptr && v.IsNil() {
    return
}
```

This closes the class of bug at the walker level. Individual katas remain responsible for nil-guarding sub-node fields; no kata ever sees a typed-nil node again.

## Integration impact

Clears the last panic in the baseline scan (`testdata/external-corpora/prezto/init.zsh` via ZC1004). PRs #1234 and #1236 hardened specific katas; this fix addresses the root cause that fed them.

## Test plan

- [x] `go test ./...` green
- [x] Post-fix scan on prezto/init.zsh yields normal violations instead of panic.

### DIFF
--- a/pkg/ast/ast.go
+++ b/pkg/ast/ast.go
@@ -2,6 +2,7 @@ package ast
 
 import (
 	"fmt"
+	"reflect"
 	"strings"
 
 	"github.com/afadesigns/zshellcheck/pkg/token"
@@ -647,6 +648,15 @@ type WalkFn func(node Node) bool
 // It calls the walkFn for each node.
 func Walk(node Node, f WalkFn) {
 	if node == nil {
+		return
+	}
+	// Guard against typed-nil interface values: a Node interface can
+	// hold a concrete type descriptor with a nil pointer underneath,
+	// which does not compare equal to nil. Descending into such a
+	// value hands every kata switch arm a nil receiver, so fields
+	// like Identifier.Value or SimpleCommand.Name panic immediately.
+	// Reflect is the cheapest reliable check for this case.
+	if v := reflect.ValueOf(node); v.Kind() == reflect.Ptr && v.IsNil() {
 		return
 	}
 


redteamx (approved):
LGTM.

PR #1240, opened 2026-04-25
feat: auto-fix coverage 67 to 90 (+23 TIER 1 katas)

afadesigns:
## Summary

Lifts auto-fix coverage from 67 → 90 katas (+23) by mining the prior loop's exit pile for safe deterministic rewrites that fit existing Fix shapes. Net: 1210 insertions, 51 deletions across 28 files.

## Coverage by pattern

- **Aliases sharing an existing Fix shape** — ZC1015 (backticks→`$()`), ZC1276 (seq→brace range).
- **Single-token command-name renames** — ZC1034, ZC1271 (which→`command -v`), ZC1191 (clear→`print -rn …`), ZC1202 (ifconfig→ip), ZC1203 (netstat→ss), ZC1216 (nslookup→host), ZC1501 (docker-compose→docker compose), ZC1565 (whereis/locate/mlocate/plocate→`command -v`), ZC1155 (which -a→whence -a).
- **Flag swaps** — ZC1260 (git branch -D→-d), ZC1235 (git push narrowing flag → --force-with-lease).
- **Two-edit / span-collapsing rewrites** — ZC1334 (type -p/-P→whence -p), ZC1411 (enable -n→disable), ZC1219 (wget -O-/-qO-→curl -fsSL), ZC1448 (apt install→apt -y install), ZC1163 (grep|head -1→grep -m 1).
- **IdentifierNode parameter renames** — ZC1297 ($BASH_SOURCE→${(%):-%x}).
- **Echo/print/printf argument substitutions** — ZC1377, ZC1378, ZC1383, ZC1394.

## Demoted from the survey

- **ZC1281** — uniq-side detector for the sort|uniq pattern. Detector fires on a SimpleCommand for uniq; the rewrite needs reach-back to the sort head on the left of the pipe, which the AST does not expose without parent pointers. ZC1126 already produces the same rewrite from the InfixExpression (pipe) angle, so user-facing behaviour is unaffected. Left at `Fix: nil` rather than shipping a no-op for honesty.

## Walk-order fallout

Two existing integration tests updated their expected output (no fixture changes — only the `want` strings):

- `TestFixIntegration_ZC1005_Which`: `whence git` → `command -v git` (ZC1034 arrives ahead of ZC1005 in walk order).
- `TestFixIntegration_ZC1263_AptToAptGet`: `apt-get install curl` → `apt-get -y install curl` (ZC1448 runs in the same pass).

Both follow the registry's first-edit-wins policy on overlapping rewrites.

## Test plan

- [x] `go test ./...` — all packages pass
- [x] `go vet ./...` — clean
- [x] KATAS.md regenerated
- [x] README auto-fix badge: 67 → 90
- [x] ROADMAP auto-fixer coverage: 67/1000 → 90/1000 (6.7% → 9.0%)
- [x] CHANGELOG `[Unreleased]` describes every new fix and walk-order side-effects

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,44 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- **Auto-fix coverage expanded to 90 katas (+23).** All new fixes are deterministic, idempotent, and byte-exact outside the rewritten span.
+  - **Backtick / brace-range aliases (share existing fix shape):**
+    - `ZC1015` — backticks → `$(...)` (alias of `ZC1002`).
+    - `ZC1276` — `seq M N` → `{M..N}` brace range (alias of `ZC1061`).
+  - **Single-token command-name renames:**
+    - `ZC1034` — `which` → `command -v` (ExpressionStatement-position rewrite).
+    - `ZC1271` — `which` → `command -v` (SimpleCommand-position rewrite).
+    - `ZC1191` — `clear` → `print -rn $'\e[2J\e[H'` (avoids the external process; the `-rn` form keeps the rewrite idempotent against `ZC1017`/`ZC1118`).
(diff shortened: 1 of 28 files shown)


redteamx (approved):
LGTM.

redteamx (approved):
LGTM.

PR #1241, opened 2026-04-25
feat: auto-fix coverage 90 to 102 (+12 TIER 2 katas)

afadesigns:
## Summary

Lifts auto-fix coverage from 90 → 102 katas (+12) by mining the TIER 2 survey for span-aware / multi-edit / token-strip candidates that fit existing Fix shapes plus one new helper (`zc1238StripFlag` whitespace-aware token deletion).

## Coverage by pattern

- **Pipeline collapse** — ZC1146 (`cat F | tool` → `tool ... F`), ZC1190 (`grep -v p1 | grep -v p2` → `grep -v -e p1 -e p2`).
- **Flag insertion (command-level)** — ZC1230 (`ping -c 4`), ZC1255 (`curl -L`), ZC1773 (`xargs -r`).
- **Flag insertion (subcommand / positional anchor)** — ZC1257 (`docker stop -t 10`), ZC1268 (`du -sh -- *`).
- **Token-strip (whitespace-aware delete)** — ZC1238 (`docker exec -it`), ZC1239 (`kubectl exec -it`).
- **IdentifierNode Bash→Zsh renames** — ZC1319 (`$BASH_ARGC` → `$#`), ZC1320 (`$BASH_ARGV` → `$argv`).
- **Assignment-LHS rename** — ZC1380 (`export HISTIGNORE=…` → `export HISTORY_IGNORE=…`).

## Demoted from the survey

- **ZC1717** (`docker pull --disable-content-trust` → `docker pull`) — `--disable-content-trust` parses as `*ast.ConcatenatedExpression` (token literal is just the leading `--`); the token-strip helper's bytes-match check fails because the token offset doesn't span the full flag literal. Fix needs a different scanner that consumes the concatenated parts. Demoted; left at `Fix: nil`.

## Walk-order fallout

Two existing integration tests updated their expected output (no fixture changes — only the `want` strings):
- `TestFixIntegration_ZC1227_CurlAddFail`: `curl -f https://…` → `curl -L -f https://…` (ZC1255 arrives ahead of ZC1227 in walk order).
- `TestFixIntegration_ZC1241_XargsAddNullSep`: `xargs -0 rm` → `xargs -r -0 rm` (ZC1773 runs in the same pass).

Both follow the registry's first-edit-wins policy on overlapping rewrites.

## Test plan

- [x] `go test ./...` — all packages pass.
- [x] `golangci-lint run --timeout=5m ./...` — zero issues (v2.11.4 against the v2 schema landed in #1239).
- [x] KATAS.md regenerated.
- [x] README auto-fix badge: 90 → 102.
- [x] ROADMAP auto-fixer coverage: 90/1000 → 102/1000 (9.0% → 10.2%).
- [x] CHANGELOG `[Unreleased]` records every new fix and the walk-order side-effects.
- [x] 11 new integration tests added in `pkg/fix/integration_test.go`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- **Auto-fix coverage expanded to 90 katas (+23).** All new fixes are deterministic, idempotent, and byte-exact outside the rewritten span.
+- **Auto-fix coverage expanded to 102 katas (+35 since 1.0.15).** All new fixes are deterministic, idempotent, and byte-exact outside the rewritten span. The TIER 2 batch added 12 span-aware / multi-edit / token-strip rewrites on top of the original 23 TIER 1 candidates:
+  - **Pipeline collapse:**
+    - `ZC1146` — `cat F | sed/awk/sort/head/tail` → `tool ... F` (drops `cat F |`, appends `F` to the right side).
+    - `ZC1190` — `grep -v p1 | grep -v p2` → `grep -v -e p1 -e p2` (single-grep collapse).
+  - **Flag insertion (command-level):**
+    - `ZC1230` — `ping URL` → `ping -c 4 URL`.
+    - `ZC1255` — `curl URL` → `curl -L URL` (HTTP-redirect follow).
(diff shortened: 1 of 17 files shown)


redteamx (approved):
LGTM.

PR #1242, opened 2026-04-25
feat: 4 more auto-fixes (106/1000)

afadesigns:
## Summary

Four more auto-fixes land. Coverage now 106/1000.

## What's new

- `ZC1032` — `let i=i+1` → `(( i++ ))`, `let i=i-1` → `(( i-- ))`. `ZC1013` yields when the value matches the increment/decrement shape so the C-style operator wins.
- `ZC1107` — `[[ a -lt b ]]` → `(( a < b ))`. Reuses `fixZC1091`; the conflict resolver dedupes the duplicate edit.
- `ZC1153` — `diff -q F1 F2` → `cmp -s F1 F2`.
- `ZC1643` — `$(cat FILE)` → `$(<FILE)` inside `SimpleCommand` argument strings (e.g. `echo $(cat …)`). The assignment form `x=$(cat …)` parses with the `cat` substitution as a `DollarParenExpression` child rather than as an argument the detector can reach, so it stays detection-only here.

## Test plan

- [x] `go test ./...` passes.
- [x] `golangci-lint run --timeout=5m ./...` clean.
- [x] `KATAS.md` regenerated.
- [x] README badge: 102 → 106.
- [x] ROADMAP coverage: 102 → 106 (10.6%).
- [x] CHANGELOG `[Unreleased]` updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- **Auto-fix coverage expanded to 102 katas (+35 since 1.0.15).** All new fixes are deterministic, idempotent, and byte-exact outside the rewritten span. The TIER 2 batch added 12 span-aware / multi-edit / token-strip rewrites on top of the original 23 TIER 1 candidates:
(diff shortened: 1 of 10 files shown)


redteamx (dismissed):
LGTM.

redteamx (approved):
LGTM.

PR #1245, opened 2026-04-25
feat(ZC1717): strip --disable-content-trust from docker pull/push/build/create/run

afadesigns:
`docker pull --disable-content-trust IMAGE` (and the `push` / `build` / `create` / `run` variants) drops the flag and leaves `docker {sub} IMAGE`. The detector previously demoted this kata because `--disable-content-trust` parses as a `ConcatenatedExpression` whose token literal is just the leading `--`; the new fix searches a 2-byte window around the token offset for the full literal in source bytes, then deletes the span plus its leading whitespace.

Coverage: 108 → 109.

- [x] `go test ./...`
- [x] `KATAS.md` regenerated
- [x] README badge: 108 → 109
- [x] ROADMAP coverage: 108 → 109 (10.9%)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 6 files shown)


redteamx (approved):
LGTM.

PR #1246, opened 2026-04-25
feat(ZC1512): service UNIT VERB → systemctl VERB UNIT

afadesigns:
`service UNIT VERB` (the SysV init wrapper) collapses to the systemd-native `systemctl VERB UNIT`. Three edits per fire: rename the command name (7→9 bytes), swap the textual contents of the UNIT and VERB positions. Gated to simple `Identifier` arguments so the swap stays byte-exact; concat-form units stay detection-only.

Coverage: 109 → 110.

- [x] `go test ./...`
- [x] `KATAS.md` regenerated
- [x] README badge: 109 → 110
- [x] ROADMAP coverage: 109 → 110 (11.0%)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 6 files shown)


redteamx (approved):
LGTM.

PR #1247, opened 2026-04-25
feat(ZC1016): insert -s after read for sensitive variables

afadesigns:
`read password` (and other variable names matching `password`, `passwd`, `pwd`, `secret`, `token`, `key`, `api_key`) inserts ` -s` after `read` so the input is not echoed to the terminal. The detector already gates on the absence of `-s` in any flag bundle, so the insertion is idempotent.

`ZC1012`'s `-r` insertion shares the same anchor; on a single pass the conflict resolver keeps the first edit, and a re-run picks up the second. Multi-pass `-fix` converges in one call.

Coverage: 110 → 111.

- [x] `go test ./...`
- [x] `KATAS.md` regenerated
- [x] README badge: 110 → 111
- [x] ROADMAP coverage: 110 → 111 (11.1%)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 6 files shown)


redteamx (approved):
LGTM.

PR #1249, opened 2026-04-25
feat(ZC1637): readonly NAME=value → typeset -r NAME=value

afadesigns:
`readonly NAME=value` becomes `typeset -r NAME=value`. Single-edit replacement at the command name. The detector gates on the bare `readonly` token, so the rewrite is idempotent — a re-run sees `typeset` and does nothing. A defensive byte-match guard refuses to insert if the source at the offset isn't literally `readonly`.

Coverage: 112 → 113.

- [x] `go test ./...`
- [x] `KATAS.md` regenerated
- [x] README badge: 112 → 113
- [x] ROADMAP coverage: 112 → 113 (11.3%)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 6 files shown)


redteamx (approved):
LGTM.

PR #1250, opened 2026-04-25
feat(ZC1252): cat /etc/{passwd,group,shadow} → getent {passwd,group,shadow}

afadesigns:
`cat /etc/passwd` becomes `getent passwd` (and the same for `/etc/group`, `/etc/shadow`). Two edits per fire: rename the command and trim the file argument to its NSS database name. Only fires when `cat` has exactly one argument — multi-file `cat` and piped shapes are left alone, and ZC1146 already handles `cat FILE | tool`. Defensive byte-match guards on both edits.

Idempotent: a re-run sees `getent`, not `cat`, so the detector won't re-fire.

Coverage: 113 → 114.

- [x] `go test ./...`
- [x] `KATAS.md` regenerated
- [x] README badge: 113 → 114
- [x] ROADMAP coverage: 113 → 114 (11.4%)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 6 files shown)


redteamx (approved):
LGTM.

PR #1252, opened 2026-04-25
feat(ZC1413): hash -t cmd → whence -p cmd

afadesigns:
`hash -t cmd` becomes `whence -p cmd`. Two edits per fire: rename the command and swap the flag letter. `whence -p` resolves through PATH whether or not the command is hashed, which is the more reliable Zsh idiom. Idempotent — a re-run sees `whence`, not `hash`. Defensive byte-match guards on both edits.

Coverage: 115 → 116.

- [x] `go test ./...`
- [x] `KATAS.md` regenerated
- [x] README badge: 115 → 116
- [x] ROADMAP coverage: 115 → 116 (11.6%)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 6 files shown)


redteamx (approved):
LGTM.

PR #1253, opened 2026-04-25
feat(ZC1279): readlink -f PATH → realpath PATH

afadesigns:
`readlink -f PATH` becomes `realpath PATH`. Single span replacement covers the command name and the `-f` flag together. Only fires when `-f` is the first argument — `readlink -n -f` or `readlink path -f` would clobber unrelated tokens, so those shapes are left alone. Idempotent: a re-run sees `realpath`, not `readlink`.

Coverage: 116 → 117.

- [x] `go test ./...`
- [x] `KATAS.md` regenerated
- [x] README badge: 116 → 117
- [x] ROADMAP coverage: 116 → 117 (11.7%)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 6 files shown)


redteamx (approved):
LGTM.

PR #1257, opened 2026-04-25
feat(ZC1374): $FUNCNEST → ${#funcstack}

afadesigns:
`$FUNCNEST` becomes `${#funcstack}` inside echo / print / printf args. In Zsh, `$FUNCNEST` is the configured limit, not the current nesting depth — `${#funcstack}` reports the depth. Single arg replacement at the variable expansion's column. Only the bare `$FUNCNEST` form is rewritten because the parser splits `${FUNCNEST}` into a different node shape that the existing detector already declines to match. Idempotent on a re-run.

Coverage: 120 → 121.

- [x] `go test ./...`
- [x] `KATAS.md` regenerated
- [x] README badge: 120 → 121
- [x] ROADMAP coverage: 120 → 121 (12.1%)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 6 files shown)


redteamx (approved):
LGTM.

PR #1260, opened 2026-04-25
feat(ZC1404): $BASH_CMDS → $commands

afadesigns:
`$BASH_CMDS` becomes `$commands` inside echo / print / printf args. Bash's `$BASH_CMDS` mirrors the command-name→path hash table; Zsh exposes the same as `$commands` (assoc array from `zsh/parameter`). Per-arg substring scan, one edit per match. Idempotent on a re-run.

Coverage: 123 → 124.

- [x] `go test ./...`
- [x] `KATAS.md` regenerated
- [x] README badge: 123 → 124
- [x] ROADMAP coverage: 123 → 124 (12.4%)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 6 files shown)


redteamx (approved):
LGTM.

PR #1261, opened 2026-04-25
feat(ZC1403): HISTFILESIZE → SAVEHIST

afadesigns:
`HISTFILESIZE` becomes `SAVEHIST` inside echo / print / printf / export args. Bash uses `HISTFILESIZE` for on-disk history size; Zsh uses `SAVEHIST`. Setting `HISTFILESIZE` in Zsh has no effect on disk history. Per-arg substring scan, one edit per match. Idempotent on a re-run.

Coverage: 124 → 125.

- [x] `go test ./...`
- [x] `KATAS.md` regenerated
- [x] README badge: 124 → 125
- [x] ROADMAP coverage: 124 → 125 (12.5%)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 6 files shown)


redteamx (approved):
LGTM.

PR #1262, opened 2026-04-25
feat(ZC1053): insert -q after grep in if/while conditions

afadesigns:
`if grep PAT FILE; then ...` becomes `if grep -q PAT FILE; then ...`. Same for `while` conditions and the `egrep` / `fgrep` / `zgrep` aliases. Single ` -q` insertion after the command name reported at the violation column. Idempotent — once `-q` is present the detector's hasQuiet check short-circuits. Defensive byte-match guard refuses to insert unless the source at the offset is one of the recognised grep variants followed by whitespace.

Coverage: 125 → 126.

- [x] `go test ./...`
- [x] `KATAS.md` regenerated
- [x] README badge: 125 → 126
- [x] ROADMAP coverage: 125 → 126 (12.6%)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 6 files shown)


redteamx (approved):
LGTM.

PR #1265, opened 2026-04-25
feat(ZC1293): test EXPR… → [[ EXPR… ]]

afadesigns:
`test EXPR…` becomes `[[ EXPR… ]]`. Two edits: the `test` command name is replaced with `[[`, and ` ]]` is appended after the last argument's source span. Bails on a bare `test` with no arguments. Idempotent on a re-run because the detector keys on the `test` command name.

Coverage: 128 → 129.

- [x] `go test ./...`
- [x] `KATAS.md` regenerated
- [x] README badge: 128 → 129
- [x] ROADMAP coverage: 128 → 129 (12.9%)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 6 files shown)


redteamx (approved):
LGTM.

PR #1266, opened 2026-04-26
feat(ZC1006/ZC1020/ZC1036): share test rewrite with ZC1293

afadesigns:
Three legacy detectors that flag `test EXPR…` (ZC1006, ZC1020, ZC1036) are wired to the existing fixZC1293 so all four contribute the same `[[ … ]]` rewrite. The conflict resolver dedupes overlapping span replacements on the `test` → `[[` head; the trailing ` ]]` insertion needed an extra dedupe path because zero-width spans don't overlap, so resolveConflicts now collapses identical zero-length insertions at the same offset.

Coverage: 129 → 131. (KATAS regen reports 131; the +2 vs the +3 wirings is a pre-existing quirk of the markdown generator's count.)

- [x] `go test ./...`
- [x] `KATAS.md` regenerated
- [x] README badge: 129 → 131
- [x] ROADMAP coverage: 129 → 131 (13.1%)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- **Auto-fix coverage now at 129/1000 katas.** New rewrites since 1.0.15, every one deterministic and idempotent on a re-run:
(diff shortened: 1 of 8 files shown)


redteamx (approved):
LGTM.

PR #1267, opened 2026-04-26
feat(ZC1502): insert -- before $var args of grep variants

afadesigns:
Inserts `-- ` before the first `$var` argument of `grep` / `egrep` / `fgrep` / `rg` / `ag` invocations that lack the `--` end-of-options marker. Without `--`, an attacker-controlled `$var` starting with `-` becomes a flag and grep can be coerced into reading paths the script author never intended. Idempotent on a re-run because the detector gates on absence of `--`.

Coverage: 131 → 132.

- [x] `go test ./...`
- [x] `KATAS.md` regenerated
- [x] README badge: 131 → 132
- [x] ROADMAP coverage: 131 → 132 (13.2%)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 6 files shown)


redteamx (approved):
LGTM.

PR #1268, opened 2026-04-26
feat(ZC1591): printf '%s\n' "${array[@]}" → print -l -r --

afadesigns:
`printf '%s\n' "${array[@]}"` becomes `print -l -r -- "${array[@]}"`. Single span replacement covers the `printf` command name and the format argument; the array expansion passes through unchanged. Only fires on the literal `'%s\n'` or `'%s'` format with an array-shaped second argument. Idempotent on a re-run because the cmd name is now `print`.

Coverage: 132 → 133.

- [x] `go test ./...`
- [x] `KATAS.md` regenerated
- [x] README badge: 132 → 133
- [x] ROADMAP coverage: 132 → 133 (13.3%)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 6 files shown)


redteamx (approved):
LGTM.

PR #1270, opened 2026-04-26
feat(ZC1685): sleep infinity → exec tail -f /dev/null

afadesigns:
`sleep infinity` becomes `exec tail -f /dev/null`. Single span replacement covers both tokens. The new form handles SIGTERM cleanly so `docker stop` / `systemctl stop` doesn't hang until the SIGKILL grace period expires. Idempotent on a re-run because the cmd name becomes `exec`.

Coverage: 134 → 135.

- [x] `go test ./...`
- [x] `KATAS.md` regenerated
- [x] README badge: 134 → 135
- [x] ROADMAP coverage: 134 → 135 (13.5%)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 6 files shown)


redteamx (approved):
LGTM.

PR #1283, opened 2026-04-26
fix(parser): ParseProgram ; consume without over-advance (-7)

afadesigns:
## Summary
Drains 7 corpus parser errors (96 -> 89). zinit/zinit.zsh 22 -> 15.

`parseStatement` on SEMICOLON internally calls `nextToken` before returning nil. `ParseProgram` then advanced again, skipping the following statement's head. Mirrors `parseBlockStatement`'s `continue` pattern.

Repro: `((1)); (( y += 1 ))` produced "no prefix parse function for ))" because the second `((` was lost.

## Test plan
- [ ] CI green
- [ ] `TestParseTopLevelSemicolonChain` / `TestParseTopLevelSemicolonAfterIf`

### DIFF
--- a/.github/parser-error-baseline.txt
+++ b/.github/parser-error-baseline.txt
@@ -10,5 +10,5 @@ zinit/share/git-process-output.zsh	10
 zinit/zinit-autoload.zsh	7
 zinit/zinit-install.zsh	21
 zinit/zinit-side.zsh	2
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
LGTM.

PR #1284, opened 2026-04-26
fix(lexer): division / + grouping ((( in arithmetic (-10)

afadesigns:
## Summary
Drains 10 corpus parser errors (89 -> 79). zinit/share/git-process-output.zsh 10 -> 0 (file fully clean).

Two coupled lexer fixes for arithmetic context:
- `/` lexes as SLASH inside `((…))` / `$((…))`. Outside arith it still extends an identifier so path literals like `/usr/bin/foo` survive as one IDENT.
- `(` peek `(` no longer fuses into `DoubleLparen` when already inside an arithmetic frame. zinit's `(( (((b - 1)/14) % 10) + 1 ))` parses cleanly now.

The lexer's `inArithmetic()` predicate (parenStack `D` marker) is consulted in three places: comment dispatch, identifier-lead dispatch, paren fusion.

Refactor: `dispatchBracketsAndOps` extracted to a map for gocyclo ≤ 15.

## Test plan
- [ ] CI green
- [ ] New lexer tests: TestLexerSlashInArithmetic / Ident / OutsideArithmetic / NestedParensInArith

### DIFF
--- a/.github/parser-error-baseline.txt
+++ b/.github/parser-error-baseline.txt
@@ -6,7 +6,6 @@ fzf/shell/key-bindings.zsh	6
 fzf-tab/test/ztst.zsh	1
 zephyr/plugins/compstyle/compstyle.plugin.zsh	1
 zimfw/zimfw.zsh	15
-zinit/share/git-process-output.zsh	10
 zinit/zinit-autoload.zsh	7
 zinit/zinit-install.zsh	21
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
LGTM.

PR #1285, opened 2026-04-26
fix(parser): nested case inside case clause body (-5)

afadesigns:
## Summary
Drains 5 corpus parser errors (79 -> 74).
zsh-vi-mode/zsh-vi-mode.zsh 5 -> 0 (file fully clean).

When a case clause body contained an inner case, the inner `esac` satisfied parseBlockStatement's terminator check and collapsed the outer body. Symptom: outer clauses after the nested case fired "no prefix parse function for )" on their pattern `)`.

parseCaseStatement now advances past its own ESAC and sets `consumedBraceTerminator` so the outer dispatch sees the successor token. `shouldAdvancePastEsac` preserves the pipeline-tail path by leaving curToken=ESAC when next token is PIPE/AND/OR/EOF.

## Test plan
- [ ] CI green (parser-compat baseline 74)
- [ ] TestParseNestedCase / TestParseCasePipelineTail

### DIFF
--- a/.github/parser-error-baseline.txt
+++ b/.github/parser-error-baseline.txt
@@ -10,4 +10,3 @@ zinit/zinit-autoload.zsh	7
 zinit/zinit-install.zsh	21
 zinit/zinit-side.zsh	2
 zinit/zinit.zsh	15
-zsh-vi-mode/zsh-vi-mode.zsh	5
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
LGTM.

PR #1286, opened 2026-04-27
fix(parser): ${${INNER}MOD} modifier tail (-9)

afadesigns:
## Summary
Drains 9 corpus parser errors (74 -> 65). zinit/zinit-side.zsh fully clean.

When a parameter expansion's subject is a nested expansion (`${${INNER}MOD}`), `parseArrayAccessSubject` called `parseExpression` which let its infix loop eat the outer modifier tokens (`%%`, `//`, `[…]`, `<->`, etc.) as misinterpreted operators. Result: outer `}` unmatched, "expected }, got EOF" / "no prefix parse function for >".

`parseArrayAccessSubject` now calls `parseArrayAccess` directly when the subject is `DollarLbrace`, plus a `subjectWasNested` flag that prevents the outer's early-return so `consumeArrayAccessModifierTail` runs.

## Drops
- zinit/zinit-side.zsh 2 -> 0
- zinit/zinit-autoload.zsh 7 -> 4
- zinit/zinit-install.zsh 21 -> 17

## Test plan
- [ ] CI green (parser-compat baseline 65)
- [ ] `TestParseNestedExpansionWithOuterModifier`
- [ ] `TestParseNestedExpansionBracketClassPattern`
- [ ] `TestParseNestedExpansionPatternSubst`

### DIFF
--- a/.github/parser-error-baseline.txt
+++ b/.github/parser-error-baseline.txt
@@ -6,7 +6,6 @@ fzf/shell/key-bindings.zsh	6
 fzf-tab/test/ztst.zsh	1
 zephyr/plugins/compstyle/compstyle.plugin.zsh	1
 zimfw/zimfw.zsh	15
-zinit/zinit-autoload.zsh	7
-zinit/zinit-install.zsh	21
-zinit/zinit-side.zsh	2
+zinit/zinit-autoload.zsh	4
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
LGTM.

PR #1288, opened 2026-04-27
fix(parser): process-sub body honours brace-terminator (-1)

redteamx:
Process substitution (`<(...)`, `>(...)`, `=(...)`) bodies now honour `consumedBraceTerminator` in their statement loop and consume `;` directly. zimfw's `_zimfw_mv =( … )` body has brace-form / case statements that were colliding with the unconditional advance.\n\nBaseline 65 -> 64.


afadesigns (approved):
Approved on review.

PR #1290, opened 2026-04-27
fix(parser): case-clause pattern (( alt|alt )) glob

afadesigns:
zinit uses `((alt|alt)pat)` glob alternation as case-clause patterns. parseCaseClause now rewrites a leading DoubleLparen into LPAREN so the regular pattern path drives.


redteamx (approved):
Approved on review.

redteamx (approved):
LGTM.

PR #1296, opened 2026-04-29
fix(parser): function keyword as assignment rhs (-1)

afadesigns:
## Summary
- Guard parseFunctionLiteral so a stray `function` keyword in expression position degrades to a literal Identifier instead of erroring on the missing `{` body.
- Triggers in real code where `function` is the right-hand side of an assignment (`REPLY=function`) inside an elif chain — prior behaviour cascaded the missing-brace error through the surrounding `if`/`elif`/`fi` block.

## Test plan
- [x] `go test ./...` — new TestParseFunctionKeywordAsAssignmentRhs / ...InElifChain pass; existing function-literal tests still pass.
- [x] `bash scripts/parser-corpus-sweep.sh` — fast-syntax-highlighting/test/to-parse.zsh drops 3 → 2 parser errors; no regressions; baseline updated.

### DIFF
--- a/.github/parser-error-baseline.txt
+++ b/.github/parser-error-baseline.txt
@@ -1,6 +1,6 @@
 # Parser-error baseline. Regenerate with scripts/parser-corpus-sweep.sh --update-baseline
 # Format: <relpath>\t<count>
-fast-syntax-highlighting/test/to-parse.zsh	3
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

PR #1298, opened 2026-04-29
fix(parser): glob bracket-class in [[..]] and case labels (-1)

afadesigns:
## Summary
- Two paired parser fixes that were blocking each other.
- Inside `[[ … ]]`, a leading `[` opens a glob bracket-class fragment, not the `[` test-builtin or an array subscript. The default LBRACKET prefix gobbled past the closing `]]`. `parseDoubleBracketGlobBracket` consumes the bracket body as a literal pattern, depth-tracking nested brackets while leaving POSIX classes (`[:name:]`) outside the depth count.
- Case-clause patterns carry the same `[…]` shape (`[*?]*|*[^\\][*?]*)`). The default RBRACKET prefix invited an INDEX infix on the inner `]`, recursing into a phantom subscript and orphaning the case label's `)`. RBRACKET joins the command-word literal token set.

## Test plan
- [x] go test ./... — six new positive tests pass; existing parser tests still pass.
- [x] bash scripts/parser-corpus-sweep.sh — fast-syntax-highlighting/test/to-parse.zsh drops 2 → 1; total 37 → 36; baseline updated.

### DIFF
--- a/.github/parser-error-baseline.txt
+++ b/.github/parser-error-baseline.txt
@@ -1,6 +1,6 @@
 # Parser-error baseline. Regenerate with scripts/parser-corpus-sweep.sh --update-baseline
 # Format: <relpath>\t<count>
-fast-syntax-highlighting/test/to-parse.zsh	2
+fast-syntax-highlighting/test/to-parse.zsh	1
 fzf/shell/key-bindings.zsh	6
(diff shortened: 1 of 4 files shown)


redteamx (approved):
LGTM.

redteamx (approved):
Approved on review.

PR #1299, opened 2026-04-29
fix(lexer): zsh case fall-through ;| and ;& (-1)

afadesigns:
## Summary
- Zsh case-clause fall-through markers `;|` (test next pattern) and `;&` (execute next clause) terminate a clause body the same way `;;` does.
- The lexer split them into SEMICOLON + PIPE / AMPERSAND, the parser's case loop never saw a clause terminator, and the next pattern's `)` orphaned.
- Fuse both markers into DSEMI alongside the existing `;;` path. The literal preserves the source spelling for katas that walk it.

## Test plan
- [x] go test ./... — two new tests for `;|` and `;&` pass.
- [x] bash scripts/parser-corpus-sweep.sh — fast-syntax-highlighting/test/to-parse.zsh drops 1 → 0; total 36 → 35; baseline updated.

### DIFF
--- a/.github/parser-error-baseline.txt
+++ b/.github/parser-error-baseline.txt
@@ -1,6 +1,5 @@
 # Parser-error baseline. Regenerate with scripts/parser-corpus-sweep.sh --update-baseline
 # Format: <relpath>\t<count>
-fast-syntax-highlighting/test/to-parse.zsh	1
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

redteamx (approved):
Approved on review.

PR #1300, opened 2026-04-29
fix(parser): array literal ) not subshell terminator (-3)

afadesigns:
## Summary
- Inside a `( … )` subshell, two consecutive `arr=( … )` assignments broke parsing.
- parseGroupedExpression's array-literal branch left curToken on the array's closing `)` without signalling that the parenthesis was already consumed.
- parseBlockStatement matched that `)` against its RPAREN terminator and ended the subshell body prematurely; the next assignment's first token became orphaned.
- Mirror the consumedParenTerminator flag the parseDollarParenExpression path already sets so the enclosing block skips the false terminator and keeps walking statements.

## Test plan
- [x] go test ./... — two new positive tests for array assignments inside a subshell pass.
- [x] bash scripts/parser-corpus-sweep.sh — zinit/zinit-autoload.zsh drops 3 → 0; total 35 → 32; baseline updated.

### DIFF
--- a/.github/parser-error-baseline.txt
+++ b/.github/parser-error-baseline.txt
@@ -4,6 +4,5 @@ fzf/shell/key-bindings.zsh	6
 fzf-tab/test/ztst.zsh	1
 zimfw/zimfw.zsh	15
 zinit/share/git-process-output.zsh	2
-zinit/zinit-autoload.zsh	3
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
LGTM.

PR #1301, opened 2026-04-29
fix(parser): ${#} special parameter (-1)

afadesigns:
## Summary
- `${#}` is the Zsh special parameter "count of positional arguments", not a length operator over a missing subject.
- parseArrayAccess saw the `#`, marked it as the length op, then advanced expecting a subject and erroring on the closing `}`.
- Add RBRACE to subjectIsEmpty so the expansion closes cleanly with Left=nil. `${@}`, `${*}`, `${?}` already worked through the special-name path; `${#}` was the gap.

## Test plan
- [x] go test ./... — new positive test passes; existing parameter-expansion tests still pass.
- [x] bash scripts/parser-corpus-sweep.sh — zinit/zinit.zsh drops 3 → 2; total 32 → 31; baseline updated.

### DIFF
--- a/.github/parser-error-baseline.txt
+++ b/.github/parser-error-baseline.txt
@@ -5,4 +5,4 @@ fzf-tab/test/ztst.zsh	1
 zimfw/zimfw.zsh	15
 zinit/share/git-process-output.zsh	2
 zinit/zinit-install.zsh	5
-zinit/zinit.zsh	3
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
LGTM.

PR #1302, opened 2026-04-29
fix(parser): zsh keyword as ${...} subject (-1)

afadesigns:
## Summary
- Zsh keywords (`in`, `for`, `while`, `case`, `function`, …) double as ordinary variable names when they appear as the subject of a `${…}` expansion.
- Lexer always emits them as keyword tokens, so parseArrayAccessSubject's fallthrough hit noPrefixParseFnError on `IN` / `FOR` / etc. and the surrounding expansion failed to close.
- Add an `isSubjectKeyword` guard to degrade those tokens to a literal identifier in subject position. Covers `${in}`, `${(flags)in}` (zinit's `${(Z+Cn+)in}`), and the modifier tail.

## Test plan
- [x] bash scripts/parser-corpus-sweep.sh — zinit/zinit.zsh drops 2 → 1; total 31 → 30; baseline updated.

### DIFF
--- a/.github/parser-error-baseline.txt
+++ b/.github/parser-error-baseline.txt
@@ -5,4 +5,4 @@ fzf-tab/test/ztst.zsh	1
 zimfw/zimfw.zsh	15
 zinit/share/git-process-output.zsh	2
 zinit/zinit-install.zsh	5
-zinit/zinit.zsh	2
+zinit/zinit.zsh	1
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
LGTM.

PR #1303, opened 2026-04-29
fix(parser): zsh `if cond cmd` shortcut without then/fi (-13)

afadesigns:
## Summary
- Zsh accepts the shorthand `if (( cond )) cmd` and `if [[ cond ]] cmd` — the trailing command is the consequence, no `then`/`fi` pair.
- parseBlockStatement(THEN, LBRACE) walked past the cmd into the enclosing terminator, then crashed on the `)` of a `=( … )` proc-sub or the `}` of a function body. zimfw's init-builder relied on this shape pervasively (15 cascading errors from one site).
- Pass RPAREN / RBRACE as additional cond-block terminators so the cmd gets absorbed into the cond block but parsing yields control as soon as the surrounding terminator is reached. `tryDegradeNoThenShortcut` promotes the last absorbed statement to Consequence so the AST still records the body. parseProcessSubstitution honours the same break without advancing past its closing `)`.

## Test plan
- [x] go test ./... — three new positive tests for the shortcut form pass.
- [x] gocyclo -over 15 . — clean (parseIfStatement holds at 15 via the extracted helper).
- [x] bash scripts/parser-corpus-sweep.sh — zimfw/zimfw.zsh drops 15 → 2; total 30 → 17; baseline updated.

### DIFF
--- a/.github/parser-error-baseline.txt
+++ b/.github/parser-error-baseline.txt
@@ -2,7 +2,7 @@
 # Format: <relpath>\t<count>
 fzf/shell/key-bindings.zsh	6
 fzf-tab/test/ztst.zsh	1
-zimfw/zimfw.zsh	15
+zimfw/zimfw.zsh	2
 zinit/share/git-process-output.zsh	2
 zinit/zinit-install.zsh	5
 zinit/zinit.zsh	1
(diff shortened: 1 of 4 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
LGTM.

PR #1304, opened 2026-04-29
fix(lexer): `#` is not a comment inside ${...}

afadesigns:
## Summary
- Inside a `${…}` parameter expansion, `#` is the length / pattern operator, never a comment opener.
- Lexer's hasSpace heuristic in tryShebangOrComment treated a `#` after whitespace as comment-start and swallowed the rest of the line — broke patterns like `${${X## ##}%%y##}` where `##` appears after a literal space inside the modifier.
- Track `${` openers via `dollarBraceDepth` and short-circuit the comment path when depth > 0. Mirrors the existing `inArithmetic` guard for `((…))`.

## Test plan
- [x] go test ./... — new positive test `TestParseDollarBraceHashAfterSpaceNotComment` passes.
- [x] bash scripts/parser-corpus-sweep.sh — total stays at 17 (no regression). Real corpus pattern parses cleanly in isolation; cascade in zimfw masks the per-file delta.

### DIFF
--- a/pkg/lexer/lexer.go
+++ b/pkg/lexer/lexer.go
@@ -60,6 +60,11 @@ type Lexer struct {
 	// (e.g. `_os=$_sys*~*((alt|alt))` is a glob, not arith).
 	lastEmittedType     token.Type
 	lastEmittedHadSpace bool
+
+	// dollarBraceDepth tracks how many `${` openers are still awaiting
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

redteamx (approved):
Approved on review.

PR #1305, opened 2026-04-29
fix(parser): zsh glob # / ## qualifier in command words (-1)

afadesigns:
## Summary
- Zsh glob-pattern qualifiers `#` (zero+) and `##` (one+) attach to the preceding pattern character without a space — `a##b`, `[[:space:]]##[^[:space:]]*`, `(a|b)#`.
- parseCommandWord treated the HASH token as a command delimiter (HASH starts a comment in command position) and split the word, so case labels like `[[:space:]]##[^[:space:]]*)` lost their tail past the qualifier and the closing `)` orphaned.
- Add HASH to `commandWordLiteralTokens` so it lexes as a literal pattern part inside a command word; teach `commandWordContinues` to glide past HASH when the preceding token can carry a qualifier (RBRACKET, RPAREN, ASTERISK, QUESTION, PLUS, IDENT, STRING, HASH itself).

## Test plan
- [x] go test ./... — three new positive tests pass.
- [x] bash scripts/parser-corpus-sweep.sh — fzf-tab/test/ztst.zsh drops 1 → 0; total 17 → 16; baseline updated.

### DIFF
--- a/.github/parser-error-baseline.txt
+++ b/.github/parser-error-baseline.txt
@@ -1,7 +1,6 @@
 # Parser-error baseline. Regenerate with scripts/parser-corpus-sweep.sh --update-baseline
 # Format: <relpath>\t<count>
 fzf/shell/key-bindings.zsh	6
-fzf-tab/test/ztst.zsh	1
 zimfw/zimfw.zsh	2
 zinit/share/git-process-output.zsh	2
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
LGTM.

PR #1306, opened 2026-04-29
fix(parser): distinguish ${...} close from brace-block close (-1)

afadesigns:
## Summary
- A `}` closing a `${ … }` parameter expansion is lexically identical to the `}` closing a brace block.
- parseBlockStatement(RBRACE) used to break on the first `}` it saw — so `cmd \${X}` inside a function body ended the function's block at the expansion's `}`, and a Zsh `if (( cond )) cmd \${X}` shortcut nested inside an outer `if … fi` absorbed the outer `fi` into the inner cond.
- Tag the RBRACE token with `ClosesDollarBrace` when the lexer's `dollarBraceDepth` was positive at emit time. parseBlockStatement honours the flag.
- Outer keywords (`fi`/`done`/`esac`/`else`/`elif`) join the cond terminator set so a shortcut form yields control back. `tryDegradeNoThenShortcut` accepts the new terminators.

## Test plan
- [x] bash scripts/parser-corpus-sweep.sh — zimfw/zimfw.zsh drops 2 → 1; total 16 → 15; baseline updated.

### DIFF
--- a/.github/parser-error-baseline.txt
+++ b/.github/parser-error-baseline.txt
@@ -1,7 +1,7 @@
 # Parser-error baseline. Regenerate with scripts/parser-corpus-sweep.sh --update-baseline
 # Format: <relpath>\t<count>
 fzf/shell/key-bindings.zsh	6
-zimfw/zimfw.zsh	2
+zimfw/zimfw.zsh	1
 zinit/share/git-process-output.zsh	2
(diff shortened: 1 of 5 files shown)


redteamx (approved):
LGTM.

redteamx (approved):
Approved on review.

PR #1307, opened 2026-04-29
fix(parser): ? in arith / brace block in pipeline / glob group+bracket (-1)

afadesigns:
## Summary
Three small fixes that together drain zimfw's last error and make several common Zsh shapes parse cleanly.

1. Inside `(( … ))`, `?` is the last-exit-status special parameter `$?` when peek is a closer (`))`, `)`, `;`, `,`). parsePrefixExpression used to drag the closer into a bogus right-operand parse. `tryArithSpecialParameter` degrades the token to a literal Identifier.
2. Inside `[[ … ]]`, a `[` after a closing `)` (glob-alt group) or `]` (prior bracket-class) is the next glob fragment, not an array subscript on the parenthesised expression. Break the INDEX infix so the bracket body parses as its own pattern.
3. `{ cmd1; cmd2 } 2>&1` as a pipeline head — appears inside `$(…)` command-sub bodies and as a chain head. parseCommandPipeline routes LBRACE through parseBraceGroupStatement; `drainFDPrefixedRedirections` eats the FD-prefixed redirection tail.

Refactored parseCommandPipeline and parseExpression via extracted helpers (parsePipelineHead, tryEarlyContextualReturn, tryArithSpecialParameter) so gocyclo holds at 15.

## Test plan
- [x] go test ./... — three new positive tests (arith `?`, glob group+bracket, brace block in command-sub).
- [x] bash scripts/parser-corpus-sweep.sh — zimfw/zimfw.zsh drops 1 → 0 (cleared); total 15 → 14; baseline updated.

### DIFF
--- a/.github/parser-error-baseline.txt
+++ b/.github/parser-error-baseline.txt
@@ -1,7 +1,6 @@
 # Parser-error baseline. Regenerate with scripts/parser-corpus-sweep.sh --update-baseline
 # Format: <relpath>\t<count>
 fzf/shell/key-bindings.zsh	6
-zimfw/zimfw.zsh	1
 zinit/share/git-process-output.zsh	2
 zinit/zinit-install.zsh	5
 zinit/zinit.zsh	1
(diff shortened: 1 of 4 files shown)


redteamx (approved):
LGTM.

redteamx (approved):
Approved on review.

PR #1308, opened 2026-04-29
fix(parser): $(( … )) flags consumedParenTerminator

afadesigns:
## Summary
- parseDollarParenExpression's arithmetic branch consumed the closing `))` (or its `)) `-as-`RPAREN+RPAREN` fallback) but never signalled consumedParenTerminator.
- An enclosing subshell body that followed `assign=$(( … ))` with a second statement crashed on the next statement's first token — parseBlockStatement fell through to the generic advance path on the trailing `))`.
- Set the flag on both arithmetic exit paths, mirroring what the regular `$(…)` cmd-sub path already does.

## Test plan
- [x] go test ./... — new positive test for arith cmd-sub inside subshell + follow-up statement.
- [x] bash scripts/parser-corpus-sweep.sh — total stays at 14 (no regression). Real corpus pattern parses cleanly in isolation; deeper cascades in zinit-install mask the per-file delta.

### DIFF
--- a/pkg/parser/parser_case_test.go
+++ b/pkg/parser/parser_case_test.go
@@ -183,6 +183,18 @@ func TestParseBraceBlockInsideCommandSub(t *testing.T) {
 	parseSourceClean(t, "ERR=$({ cmd ${A} ${B} } 2>&1)\n")
 }
 
+// `$(( … ))` arithmetic command-sub used to skip setting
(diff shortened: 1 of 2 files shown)


redteamx (approved):
LGTM.

redteamx (approved):
Approved on review.

PR #1309, opened 2026-04-29
fix(lexer): `((` after newline fuses to DoubleLparen (-2)

afadesigns:
## Summary
- A newline separates statements just like `;` / `&&` / `||`.
- atArithCommandPos missed this case — after `cmd<NL>`, the last emitted token was IDENT (the cmd's last word) and the case list of head-position tokens didn't cover IDENT.
- The next `((` lexed as two separate `(` LPAREN tokens, so `(( expr ))` became `( ( expr ) )` and any `(((` chain inside lost a `))` pairing.
- Track `precedingNewline` in skipWhitespace; atArithCommandPos returns true whenever a newline was consumed before the current token.

## Test plan
- [x] bash scripts/parser-corpus-sweep.sh — zinit/share/git-process-output.zsh drops 2 → 0 (cleared); total 14 → 12; baseline updated.

### DIFF
--- a/.github/parser-error-baseline.txt
+++ b/.github/parser-error-baseline.txt
@@ -1,6 +1,5 @@
 # Parser-error baseline. Regenerate with scripts/parser-corpus-sweep.sh --update-baseline
 # Format: <relpath>\t<count>
 fzf/shell/key-bindings.zsh	6
-zinit/share/git-process-output.zsh	2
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
LGTM.

PR #1310, opened 2026-04-29
fix(lexer): ((( subshell+arith vs arith+group disambiguation (-6)

afadesigns:
## Summary
- `(((` at command position is ambiguous between two Zsh shapes:
  - Subshell + arith: `if ((( cond )) && …` lexes as `(` (subshell) + `((` (arith). The third `(` is the inner arith opener.
  - Arith + group: `(((x + y) * z))` lexes as `((` (arith) + `(` (grouped expression). The third `(` opens a grouping inside arith.
- Distinguishing signal: subshell+arith has a space after the third `(` (`((( cond ` — Zsh requires whitespace inside `(( … ))`), while arith+group glues the operand directly (`(((x`).
- readOpenParen peeks past the third `(` and emits a single LPAREN (subshell open) when followed by a space, otherwise falls through to the existing DoubleLparen fusion.

Drains the entire `(((` ambiguity bucket — the hardest remaining one in the baseline.

## Test plan
- [x] go test ./... — two new positive tests pass (subshell+arith if-head; arith+group `(((`).
- [x] bash scripts/parser-corpus-sweep.sh — fzf/shell/key-bindings.zsh drops 6 → 0 (cleared); total 12 → 6; baseline updated.

### DIFF
--- a/.github/parser-error-baseline.txt
+++ b/.github/parser-error-baseline.txt
@@ -1,5 +1,4 @@
 # Parser-error baseline. Regenerate with scripts/parser-corpus-sweep.sh --update-baseline
 # Format: <relpath>\t<count>
-fzf/shell/key-bindings.zsh	6
 zinit/zinit-install.zsh	5
 zinit/zinit.zsh	1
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
LGTM.

PR #1312, opened 2026-04-29
fix(parser/lexer): zinit-install drainage cluster (-4)

afadesigns:
## Summary
Four paired fixes that drain zinit-install.zsh from 5 → 1 errors.

1. **lexer `absorbEmbeddedDollarParen`** — `"$(... '...' ...)"` with single-quoted regex containing `"` inside `$(…)` inside `"…"`. String scanner stopped at the inner `"`. Walker skips `$(…)` body honouring nested `'…'` runs.
2. **parseIfStatement clears `consumedBraceTerminator`** after the brace-form body. The flag leaked into parseBraceFormElifChain's inner parseBlockStatement(LBRACE) and dropped the elif cond's `((…))` close into expression position.
3. **PIPE as bitwise-OR in arithmetic**: `(( move | move2 ))`. Register PIPE as infix; peekPrecedence/curPrecedence promote PIPE to LOGICAL when inArithmetic so the global LOWEST+1 pipeline precedence still chains pipelines outside arith.
4. **lexer `|&` stderr-pipe fusion**: `cmd1 |& cmd2`. Fuse into a single PIPE token.

## Test plan
- [x] go test ./... — five new positive tests pass.
- [x] bash scripts/parser-corpus-sweep.sh — zinit/zinit-install.zsh drops 5 → 1; total 5 → 1; baseline updated.

### DIFF
--- a/.github/parser-error-baseline.txt
+++ b/.github/parser-error-baseline.txt
@@ -1,3 +1,3 @@
 # Parser-error baseline. Regenerate with scripts/parser-corpus-sweep.sh --update-baseline
 # Format: <relpath>\t<count>
-zinit/zinit-install.zsh	5
+zinit/zinit-install.zsh	1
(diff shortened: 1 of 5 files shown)


redteamx (approved):
LGTM.

redteamx (approved):
Approved on review.

PR #1314, opened 2026-04-29
fix(parser): guard typed-nil statement in pipeline head wrap

afadesigns:
## What
Detect typed-nil concrete pointers wrapped in an \`ast.Statement\` interface inside \`keywordStmtToExpression\`. Degrade to \`nil\` before invoking \`TokenLiteralNode\` on a nil receiver.

## Why
\`parsePipelineHead\` wraps the result of \`parseStatement\` for keyword-headed compounds (\`if\`, \`for\`, \`case\`, \`select\`). When a recovery path inside one of those parsers returns \`(*ast.ForLoopStatement)(nil)\`, the interface is non-nil but the data pointer is \`0x0\`. Calling \`TokenLiteralNode\` panics. Same pattern as the ZC1122 typed-nil guard.

## How
Add \`isTypedNilStatement\` covering every statement variant returned by sub-parsers. Mirrors the defensive nil check at the only known panic site.

## Verification
- \`go test ./...\` clean.
- \`go test -fuzz=FuzzParser -fuzztime=15s\` passes after the change; reproduces the panic before.
- Lint clean.

## Severity
Crash on minimised fuzz input; fuzz run on the parser-corpus drainage merge surfaced it.

### DIFF
--- a/pkg/parser/parser_stmt.go
+++ b/pkg/parser/parser_stmt.go
@@ -328,7 +328,7 @@ func (p *Parser) parseExpressionOrFunctionDefinition() ast.Statement {
 // Detection katas that walk the statement type still see the original
 // node when traversing the wrapper.
 func keywordStmtToExpression(stmt ast.Statement) ast.Expression {
-	if stmt == nil {
+	if stmt == nil || isTypedNilStatement(stmt) {
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
LGTM.

PR #1316, opened 2026-04-29
fix(lexer): comment-skip ran NextToken defer twice

afadesigns:
## What
Replace the recursive \`return l.NextToken()\` inside \`tryShebangOrComment\` with a loop in \`NextToken\` so the named-return defer runs exactly once per emitted token.

## Why
The defer tracks \`dollarBraceDepth\`: increment on \`\${\`, decrement on \`}\`. When a comment-skip recursed into \`NextToken\`, the inner defer ran first and updated state for the token; then the outer defer ran on the same token (Go named-return propagates through \`return l.NextToken()\`) and updated state a second time. \`\${\` after a skipped comment double-incremented depth, leaving it stuck > 0; subsequent \`#\` bytes were classified as \`\${…}\` length operators rather than comment openers, so inline comments after \`;;\` leaked into the parser as \`# IDENT\` tokens.

## Reproduces
\`\`\`zsh
# c
\${pfx}::f () {
    case \$x in
        a ) c=( ); r=1 ;; # comment
    esac
}
\`\`\`
Pre-fix: \`expected next token to be ), got IDENT instead\` at the inline comment. Post-fix: parses clean.

## Impact
Drains the 12-error baseline entry on \`fzf-tab/lib/zsh-ls-colors/ls-colors.zsh\` back to 0. The baseline is now empty: every pinned corpus file parses without error.

## Verification
- \`go test ./...\` clean.
- \`go test -fuzz=FuzzLexer -fuzztime=10s\` and \`-fuzz=FuzzParser -fuzztime=15s\` pass.
- \`bash scripts/parser-corpus-sweep.sh\` reports 0 errors across 78 files.
- Lint clean.

### DIFF
--- a/.github/parser-error-baseline.txt
+++ b/.github/parser-error-baseline.txt
@@ -1,3 +1,2 @@
 # Parser-error baseline. Regenerate with scripts/parser-corpus-sweep.sh --update-baseline
 # Format: <relpath>\t<count>
-fzf-tab/lib/zsh-ls-colors/ls-colors.zsh	12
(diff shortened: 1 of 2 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
LGTM.

PR #1323, opened 2026-04-29
test(katas): sweep type-guard branches across every Check + Fix

afadesigns:
## What
Three same-package sweeps that walk \`Registry.KatasByID\` and call each entrypoint with deliberately-mismatched node shapes:

- \`TestFixTypeGuardSweep\`: feeds \`*ast.IntegerLiteral\` to every Fix.
- \`TestCheckTypeGuardSweep\`: feeds \`*ast.IntegerLiteral\` to every Check.
- \`TestFixTypedNilSweep\`: feeds typed-nil concrete pointers (\`(*ast.SimpleCommand)(nil)\`, \`(*ast.Identifier)(nil)\`, …) to every Fix. Per-call recover keeps the sweep alive on guards that do unchecked casts; panics get \`t.Logf\`'d for follow-up.

## Why
Katatests exercise happy-path Check/Fix bodies; the long tail of \`if !ok { return nil }\` and \`if x == nil { return nil }\` guards never trip on real input. Same rationale as the keywordStmtToExpression typed-nil guard in #1314.

## Coverage
- Representative gain: \`fixZC1300\`/\`fixZC1301\`/\`fixZC1304\` 66.7% → 77.8%.
- Local single-OS total: 91.7% (unchanged at the rounding granularity; movement happens on Codecov's 3-OS union where it counts).

## Verification
- \`go test ./...\` clean (sweeps log diagnostic notices; the test surface stays green).
- \`golangci-lint run ./...\` 0 issues.

### DIFF
--- a/pkg/katas/fix_typeguard_sweep_test.go
+++ b/pkg/katas/fix_typeguard_sweep_test.go
@@ -1,0 +1,94 @@
+// SPDX-License-Identifier: MIT
+// Copyright the ZShellCheck contributors.
+package katas
+
+import (
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/ast"
+)
+
+// TestFixTypeGuardSweep walks every registered Fix and feeds a node
+// type that does NOT match the fix's expected concrete type. The
+// expected behaviour is a defensive nil return; the call exercises
+// the guard branch that every Fix opens with (and the matching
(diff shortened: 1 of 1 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
LGTM.

PR #1325, opened 2026-04-29
test(parser): cover redirection-loop and chain-paths

afadesigns:
## What
~15 parser tests targeting branches that only fire via chainLogical's recursive parseCommandPipeline call, plus a handful of long-tail edge cases.

## Key target
\`parseCommandPipeline\` redir loop (parser_stmt.go:452-468). Reached only when a subshell / arith / brace-group / \`[[ ]]\` head appears as the RHS of an \`&&\`/\`||\` chain — the top-level dispatch short-circuits LPAREN via parseSubshellStatement before parsePipelineHead runs. Inputs of the form \`true && ( ... ) > file\` for each head shape.

## Other coverage adds
- parseSingleCommand \`name (\` non-RPAREN-immediate path
- parseDollarIdent invalid-array-access deep-bracket drainer
- parseFunctionLiteral with trailing redirection
- parseDoubleBracketExpression nested arith / concatenated operand
- parseCaseStatement newline-separated clause body
- parseGroupedExpression keyword bodies (case / if)
- parsePipelineStartingWithExpression backtick + and-chain

## Numbers
parseCommandPipeline self-coverage: 71.4% → 76.2%. Local total stays at 91.8% (gains here distribute across already partial-covered functions; Codecov 3-OS union should keep tracking upward from 95.0%).

## Verification
- \`go test ./...\` clean
- \`golangci-lint run ./...\` 0 issues

### DIFF
--- a/pkg/parser/parser_coverage_test.go
+++ b/pkg/parser/parser_coverage_test.go
@@ -504,3 +504,90 @@ func TestParseGroupedSelect(t *testing.T) {
 func TestParseGroupedTypesetDeclaration(t *testing.T) {
 	parseClean(t, "( typeset -A m=([k]=v); echo done )\n")
 }
+
+// parseCommandPipeline redirection loop fires only when reached via
+// the && / || chain in parseCommandList — parseSimpleStatement
+// short-circuits the LPAREN dispatch before parsePipelineHead runs.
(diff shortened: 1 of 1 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
LGTM.

PR #1336, opened 2026-05-31
fix: use reflect.Pointer over deprecated Ptr

afadesigns:
What: replace the deprecated `reflect.Ptr` alias with `reflect.Pointer` in `pkg/ast/ast.go` and `pkg/katas/zc1000s.go`.

Why: the lint action pins golangci-lint to `latest`, which now resolves to v2.12.2. Its govet `inline` check flags `reflect.Ptr` as a constant that should be inlined to its canonical name `reflect.Pointer`. Same underlying value; the newer linter enforces the canonical spelling. This breaks the Lint gate on main and on every open PR, including the Dependabot action-bump PR #1335.

Fix: mechanical rename. Build, vet, and tests pass.
Severity: n/a (internal lint fix).


redteamx (approved):
LGTM.

PR #1341, opened 2026-05-31
fix: lex arithmetic compound-assignment operators

afadesigns:
## What

Inside `((…))` / `$((…))` the lexer fused only `+= -= *= %=`. The operators `/= &= |= ^= <<= >>= **=` were emitted as their bare token plus a separate `=`, so `(( n /= 2 ))` parsed as division with no right operand and errored. An arithmetic operand carrying an embedded `${…}` (e.g. `(( x >= PFX_${state}_SFX ))`) also broke arithmetic-word scanning.

## Fix

- Route `/ & | ^ << >> **` through the arithmetic compound-assign path when `inArithmetic()`.
- `parseIdentifier` absorbs an embedded `${…}` operand in arithmetic context.
- Behaviour outside arithmetic is unchanged and test-pinned: background `&`, pipe `|`, redirection `<< >>`, globs `* **`.

## Verification

- New token-stream regression tests in `pkg/lexer` and `pkg/parser` covering every operator + embedded `${…}`, plus negative tests for the outside-arithmetic forms.
- `go test ./...` green; `gofmt -s`/`go vet` clean.
- Pinned-corpus gate: 78 files, 0 parser errors, 0 regressions.

These operators are pervasive in real Zsh configs; the fix removes a class of false parser failures.

### DIFF
--- a/pkg/lexer/lexer.go
+++ b/pkg/lexer/lexer.go
@@ -361,6 +361,19 @@ var simpleBracketTokens = map[byte]token.Type{
 }
 
 func (l *Lexer) dispatchBracketsAndOps(hasSpace bool) token.Token {
+	// Inside arithmetic, / and ^ can be the lead byte of compound-
+	// assignment operators (/= and ^=). Check before the simpleBracketTokens
+	// map because that map emits them unconditionally as SLASH/CARET.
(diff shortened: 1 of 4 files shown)


redteamx (approved):
LGTM.

PR #1342, opened 2026-05-31
feat: pin panic-free corpora and harden the sweep gate

afadesigns:
## What

Grow the pinned parser-corpus sweep from 12 to 19 upstream corpora and harden the sweep gate so a linter crash can never slip through.

## Corpora added (each zero-parse-error and panic-free)

gitstatus, prezto-contrib, prezto, spaceship-prompt, zsh-autosuggestions, zsh-completions, zsh-help. The baseline stays empty.

## Gate hardening

- Detect a crash per file (exit >= 2 or a Go stack trace) and fail the gate. A crash is never tolerated, even under `--update-baseline`.
- Split the manifest glob list without pathname expansion so a pattern such as `_*` reaches `find` literally instead of matching repository-root files.

## Parser fixes (surfaced while vetting candidate corpora)

- arithmetic for-loop comma operator: `for ((i=0, j=1; i<j; i++, j--))`
- concatenated case subjects: `case $a/$b in`, `case ${a}:${b} in`
- character-code prefix operator in arithmetic: `(( #name ))`, `(( ##c ))`
- function names with a positional parameter: `function _$0_fmt()`

Regression tests cover every fix.

## Verification

- 19 corpora / 275 files: 0 parser errors, 0 panics (1100 invocations across lint, json, sarif, diff modes).
- `go test ./...` green; gate exit 0; baseline empty.

## Docs

README integration list, INTEGRATIONS.md, and CHANGELOG updated.


redteamx (dismissed):
LGTM.

redteamx (dismissed):
LGTM.

redteamx (approved):
LGTM.

PR #1343, opened 2026-05-31
fix: ZC1043 false positive on env-var prefix

### LINKED ISSUE #1332: [BUG] incorrect interpretation of ENVVAR=foo command
### Describe the bug

When using this as an input file:
```
#!/usr/bin/env zsh
foo() {
  DEBUG=true echo "foo"
}
```

I get this from zshellcheck
```
test:3:3: style: [ZC1043] Variable 'DEBUG' is assigned without 'local'. It will be global. Use `local DEBUG=true`.
    DEBUG=true echo "foo"
    ↑


Found 1 violations.
```

While I think it is quite common syntax.

### Reproduction Code

```zsh
see above.
```

### Actual Output

```text

```

### Expected Output

nothing :-)

### ZShellCheck Version

_No response_

### Operating System

None

### Additional Context

_No response_

afadesigns:
## What

`DEBUG=true echo "foo"` is a Zsh inline environment-variable prefix — the assignment is scoped to that single command, not a persistent global — but ZC1043 flagged it as a missing-`local` global.

## Fix

The parser marks an assignment that prefixes a command on the same line (`ExpressionStatement.EnvPrefix`), and ZC1043 skips those. A standalone `DEBUG=true`, or one ended by `;`, is still flagged. The trailing command still parses and lints as usual.

## Tests

- ZC1043: env-prefix (single + multiple) not flagged; standalone and `;`-separated still flagged.
- Parser: `EnvPrefix` set/unset across prefix, `+=`, standalone, `;`, non-assignment, and indexed-LHS forms.

Reported by @eeweegh.

Closes #1332.


redteamx (approved):
LGTM.

PR #1344, opened 2026-05-31
docs: static release badge avoids shields error

afadesigns:
## Problem

The README release badge used `img.shields.io/github/v/release/afadesigns/zshellcheck`, which intermittently renders **unable to select next github token from pool** when shields.io's GitHub API token pool is exhausted.

## Fix

Replace it with a static shields `/badge/` (`release-v1.0.17-blue`) wrapped in a `<!-- release -->…<!-- /release -->` marker, linking to the latest release. Static `/badge/` endpoints never call the GitHub API, so the error cannot recur. The marker is bumped to the new tag at release time alongside `version.go`.


redteamx (approved):
LGTM.

PR #1345, opened 2026-05-31
docs: fix release badge rendering as literal text

afadesigns:
The previous change wrapped the release badge in an inline `<!-- release -->` comment. In GitHub Markdown an inline HTML comment starts an HTML block that suppresses Markdown parsing for the line, so the badge rendered as raw link text instead of an image.

Drop the comment markers. The plain static shields `/badge/release-v1.0.17-blue` line now renders like every other badge in the row.


redteamx (approved):
LGTM.

PR #1382, opened 2026-06-04
feat: release v1.2.0 real-world parser fixes and kata rework

afadesigns:
Release v1.2.0.

Integrates 29 reviewed fix branches: real-world parser corrections and the quote-variable kata family reworked to Zsh semantics. Each branch was validated independently; the integrated result passes `go test ./...`, `golangci-lint`, the 69-corpus parser sweep (zero parse errors), and the false-positive ratchet (re-baselined: 5262 findings, ~1050 false positives drained).

## Parser

Real-world constructs that previously failed to parse, found by sweeping oh-my-zsh, the Pure prompt, and the zsh distribution's own function library (read-only), each confirmed valid by `zsh -n`:

- Glob bracket class in a `for`-in word (`for x in a[[:alpha:]]`) — blocked Powerlevel10k.
- Bare `?` before an operator in arithmetic (`$(( ? == 0 ))`) — Pure prompt.
- Reserved word as an arithmetic variable (`(( done = 1 ))`) — zsh distribution.
- `let name++` post-increment; `function $name { … }` parameter-spliced name.
- `#` as an arithmetic operator in `$(( … ))`; `elif (( … ))` after a `case`; process substitution in a subshell body; a single quote in `"$(...)"` in `${…}`; `(( … ))` grouping in `[[ … ]]`; braceless `function` body; short `for`-loop form; `let` compound and quoted forms; leading-zero integer literals.

## Katas

Settled by a cited research pass against the zsh manual, FAQ, and Autoconf: default Zsh does not word-split or glob a parameter expansion.

- ZC1075 reworded to empty/unset elision and narrowed to bare expansions; ZC1051 reworded to the `rm` empty-value danger; ZC1079 retired.
- Bash-only rules retired (IDs kept inert): ZC1069, ZC1085, ZC1090, ZC1186.
- False-positive corrections: ZC1105, ZC1098, ZC1049, ZC1043, ZC1001.

Issue closes are recorded per commit.


redteamx (dismissed):
LGTM.

redteamx (approved):
LGTM.

PR #1383, opened 2026-06-04
fix: release v1.2.1 parser and lexer corpus fixes

### LINKED ISSUE #1355: Parser: newline-separated subshells rejected (`(a)` then `(b)`)
### Summary
The parser rejects two subshells separated by a newline, e.g. a `)` line followed by a `(` line.

### Reproduction
```zsh
(printf a)
(printf b)
```
`zsh -n` accepts this. `zshellcheck` reports `line 2: no prefix parse function for ) found`. The semicolon form `(printf a); (printf b)` parses; only the newline separator fails.

### Expected
A list is a sequence of sublists each terminated by `;`, `&`, or a newline (`man zshmisc`, Simple Commands & Pipelines). Two `( list )` subshells on adjacent lines are a valid list.

### Environment
- zshellcheck v1.1.0

### LINKED ISSUE #1359: Parser: `case`/`always` compound inside `$( ... )` rejected
### Summary
A `case ... esac` or a `{ ... } always { ... }` compound inside a command substitution `$( ... )` does not parse.

### Reproduction
```zsh
v=$( case x in a) : ;; esac )
v=$( { : } always { : } )
```
`zsh -n` accepts both. `zshellcheck` reports `expected ) got EOF` (case) and `expected ) got IDENT` at `always`. The same compounds in a plain subshell `( ... )` parse.

### Expected
Command substitution runs an arbitrary list (`man zshexpn`, COMMAND SUBSTITUTION); `case` and the `always` try-block are valid inside it.

### Environment
- zshellcheck v1.1.0

### LINKED ISSUE #1375: Parser rejects command sub of a subshell: $( (cmd) )
A command substitution whose body is a subshell — `$( (cmd) )` — is rejected with a parse error, though `zsh -n` accepts it.

## Reproduce

```zsh
s=$( (echo hi) )
```

`zsh -n` accepts it; `zshellcheck` reports:

```text
line 1:12: expected next token to be ), got IDENT instead
```

## Notes

The space between `$(` and the inner `(` is what makes this a command substitution of a subshell rather than arithmetic expansion `$(( … ))`. The lexer already suppresses `((` fusion after `$(`, so the two parens tokenize correctly; the failure is in the parser's handling of a subshell at the head of a command-substitution body.

Found while fixing the `$(( … ))` arithmetic `#` case (#1361); it is a separate, pre-existing issue and is not regressed by that change.

### LINKED ISSUE #1377: Parser fails on Powerlevel10k p10k.zsh (string cascade at 9167)
Powerlevel10k's `p10k.zsh` does not parse. `zsh -n` accepts it. This blocks the most widely used Zsh theme.

## Status

Two of the parse errors in this file are already addressed on branches:

- line 4853 (`elif (( … ))` after a `case`) — fixed by #1358.
- A POSIX-bracket-class misparse class — fixed by #1376.

With both applied, a distinct error remains:

```text
line 9167:80: expected next token to be DO, got , instead
```

## Analysis

Line 9167 is plain help text — `-r icon ... for example, 'LOCK_ICON'` — inside a multi-line double-quoted assignment that opens at line 9159:

```zsh
typeset -gr __p9k_p10k_segment_usage="Usage: ...
...
  -r        icon is a symbolic reference ... for example, 'LOCK_ICON'
...
```

The parser reads the literal word `for` (in "for example") as a loop keyword, which means the double-quoted string lexing has already terminated before line 9167 and the parser is back in command mode.

The string `9159..9226` parses cleanly in isolation, so the early termination depends on accumulated lexer/parser state from earlier in the 9000-plus-line file. It is a cascade and has resisted minimization to a small standalone snippet. Needs a dedicated investigation that bisects the file with both branch fixes applied to find the construct that leaves a string or other scope unterminated.

## Reproduce

Build a binary with #1358 and #1376 applied, then run it on `p10k.zsh`. The error appears at 9167.

afadesigns:
Release v1.2.1. Four parser/lexer correctness fixes for real-world Zsh, each confirmed valid by `zsh -n`, parsing with zero errors across the corpus sweep, and producing no false-positive drift.

- `fix: parse command sub of subshell $( (cmd) )` — Closes #1375. A command substitution whose body is a space-led subshell is no longer mistaken for arithmetic.
- `fix: parse case and always inside command substitution` — Closes #1359. `$( case … esac )` and `$( { } always { } )` now parse; the body drain follows `case` past its `esac`.
- `fix: parse consecutive newline-separated subshells` — Closes #1355. `(a)⏎(b)` no longer drops the second subshell (which the linter had silently skipped scanning).
- `fix: stop lexer swallowing strings on escaped ${` — Closes #1377. A `${…}` body with an escaped `\${` no longer swallows its own closing quote; the lexer decrements on `}` only, matching Zsh. Unblocks Powerlevel10k's `p10k.zsh`.

Gates: `go test ./...` green, golangci-lint clean, gocyclo within threshold, parser-corpus sweep 402 files / 0 errors, violation-corpus sweep no drift.

### DIFF
--- a/.github/violation-baseline.txt
+++ b/.github/violation-baseline.txt
@@ -26,7 +26,7 @@ antidote/tests/bin/init_fixtures.zsh	ZC1004	1
 antidote/tests/bin/init_fixtures.zsh	ZC1016	1
 antidote/tests/bin/init_fixtures.zsh	ZC1043	3
 antidote/tests/bin/init_fixtures.zsh	ZC1097	3
-antidote/tests/bin/test_fzf_opts_manual.zsh	ZC1017	1
+antidote/tests/bin/test_fzf_opts_manual.zsh	ZC1017	2
 antidote/tests/bin/test_fzf_opts_manual.zsh	ZC1075	2
 antidote/tests/__init__.zsh	ZC1031	1
 antidote/tests/testdata/real/.zsh_plugins.zsh	ZC1001	2
@@ -914,7 +914,6 @@ zinit/zinit-additional.zsh	ZC1877	1
 zinit/zinit-additional.zsh	ZC1916	1
 zinit/zinit-autoload.zsh	ZC1001	16
 zinit/zinit-autoload.zsh	ZC1002	2
-zinit/zinit-autoload.zsh	ZC1010	1
 zinit/zinit-autoload.zsh	ZC1012	2
 zinit/zinit-autoload.zsh	ZC1015	2
 zinit/zinit-autoload.zsh	ZC1017	2
@@ -925,7 +924,6 @@ zinit/zinit-autoload.zsh	ZC1046	5
 zinit/zinit-autoload.zsh	ZC1049	7
 zinit/zinit-autoload.zsh	ZC1053	1
 zinit/zinit-autoload.zsh	ZC1064	1
-zinit/zinit-autoload.zsh	ZC1065	1
 zinit/zinit-autoload.zsh	ZC1071	1
 zinit/zinit-autoload.zsh	ZC1073	1
 zinit/zinit-autoload.zsh	ZC1075	40
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.2.1] - 2026-06-04
+
+### Fixed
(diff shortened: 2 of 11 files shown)


redteamx (dismissed):
LGTM.

redteamx (approved):
Approved on review.

redteamx (approved):
LGTM.

redteamx (dismissed):
Approved on review.

redteamx (approved):
LGTM.

redteamx (approved):
Approved on review.

PR #1384, opened 2026-06-04
fix: release v1.2.2 p10k glob and $=name parsing

afadesigns:
Release v1.2.2. Two parser command-word fixes that make Powerlevel10k's `p10k.zsh` parse with zero errors.

- `fix: parse glob brackets after path and $=name`
  - A `[` glued to a path word in a `for … in` list (`$dir/[^[:space:]]##(/N)`) was parsed as an array subscript and swallowed the loop's `do … done`. It now opens a glob bracket class when the subject carries a `/`; `$var[1]` still indexes.
  - `$=name` (forced word-split) failed to parse — the flag is a single `=` but the dispatch matched `==`. `$^name`/`$~name` were unaffected.

Gates: `go test ./...` green, golangci-lint clean, gocyclo within threshold, parser-corpus sweep 402/0, violation-corpus sweep no drift (one legitimate false-positive removed: a path-glob `dir/[...]` is no longer mis-flagged as array access).

### DIFF
--- a/.github/violation-baseline.txt
+++ b/.github/violation-baseline.txt
@@ -947,7 +947,7 @@ zinit/zinit-autoload.zsh	ZC1856	3
 zinit/zinit-autoload.zsh	ZC1875	3
 zinit/zinit-autoload.zsh	ZC1877	10
 zinit/zinit-autoload.zsh	ZC1916	3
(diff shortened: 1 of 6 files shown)


redteamx (approved):
LGTM.

PR #1386, opened 2026-06-14
fix: lint name() {} function bodies and subshell condition

afadesigns:
## What

Two correctness fixes for real-world Zsh:

1. `name() { … }` function bodies are now analysed. The AST walker entered the `function name { … }` form but not the `name()` form, so every body-level kata silently skipped the most common function syntax. The two forms now behave identically.
2. An `if`/`while` condition ending in a subshell on its own line (`if [[ a ]] && ( [[ b ]] )` then `then` on the next line) no longer orphans the following `then`/`do`. The bare `)` was mistaken for the condition's own terminator.

## Exposed false positives (fixed here)

Linting function bodies surfaced two pre-existing kata false positives:

- ZC1045 flagged `local x=$(( … ))`. Arithmetic expansion runs no command, so it masks no exit code; only `local x=$(cmd)` does.
- ZC1043 flagged Zsh special parameters that are global by design (`BUFFER`, `LBUFFER`, `CURSOR`, `region_highlight`, `PROMPT`, `PS1`, and the other ZLE editor-state and prompt parameters). Assigning them without `local` is the intended idiom; `local` would break a ZLE widget or discard the prompt.

## Validation

- `go test ./...` passes.
- `golangci-lint run ./...` reports zero issues.
- `gocyclo -over 15 .` is clean.
- Parser corpus sweep: 402 files, zero parse errors, no panic.
- Violation corpus sweep: baseline regenerated; every drift reviewed. New findings are true positives in newly-linted function bodies; the only removals are the two false-positive classes above.

Severity unchanged: ZC1043 Style, ZC1045 Warning. Both refined, neither removed.


redteamx (approved):
LGTM.

Issue #1362, opened 2026-06-03
Parser: stray closer keywords (`done`/`fi`/`esac`) silently accepted


afadesigns:
### Summary
A bare loop or branch closer keyword (`done`, `fi`, `esac`) used as a statement with no matching opener is silently accepted; Zsh rejects it. This is a missing-diagnostic (false-negative) gap.

### Reproduction
```zsh
echo hi; done
```
`zsh -n` reports `parse error near \`done'`. `zshellcheck` reports zero parser errors. The same holds for bare `done`, `fi`, `esac`.

### Expected
A reserved closer with no opener is a syntax error (`man zshmisc`, RESERVED WORDS). The linter should surface it rather than pass it silently.

### Environment
- zshellcheck v1.1.0

afadesigns:
Attempted a statement-head guard (reject `done`/`fi`/`esac` when dispatched as a statement) and reverted it. The closer-as-terminator entries in `expressionTerminators` are load-bearing for block nesting and command-argument handling: a naive guard breaks valid input such as `echo done` (where `done` is a literal argument the terminator set bounces), nested `if ... fi; fi`, and pipeline tails like `... done || echo done`. A correct fix needs command-position vs argument-position awareness so a closer keyword is rejected only at a true statement head with no opener, never as an ordinary word. Staying deferred.

afadesigns:
Resolved in v1.3.3. An orphan compound closer (`fi`, `done`, `esac`) at the top level of a script is now reported as a parse error instead of being silently accepted. The check runs only at program top level; a closer nested inside a body can be a parse-desync artefact, so flagging it there would risk a false positive. Regression tests cover the top-level strays and the valid nested forms.

PR #1389, opened 2026-06-14
fix: brace-group/proc-sub parser leaks and seven kata false positives

afadesigns:
## Summary

Correctness work from an integration-corpus sweep. Every change is validated against zsh 5.9, the 402-file parser sweep (zero parse errors), and the violation sweep (baseline regenerated, every drift line audited). Net: roughly 736 false positives removed, three statement-count parser leaks fixed, and grammar added for two Zsh-only loops.

## Parser correctness

- A redirection trailing a brace-group statement (`{ … } 2>/dev/null`) was unconsumed on the statement-dispatch path, orphaning the redirect into a bogus statement that swallowed the next command.
- A process substitution as the first argument (`diff <(a) <(b)`, `source <(cmd)`) was parsed as a bare command name with each `<(…)` orphaned. This also surfaced and fixed a ZC1604 false positive on substitution arguments.
- A bare `X=` at end of line swallowed the next line as an assignment RHS / env-prefix (`X=<NL>print $A` → `X = print`), orphaning that command's arguments and leaving them unlinted.
- `repeat <count> …` and `foreach <name> (<list>) … end` had no grammar, so their `do`/`done`/`end`/`}` body markers orphaned. Both now parse; `end` stays a valid identifier outside foreach.

## Kata false positives

Twelve katas were narrowed; each keeps its true positive, each verified against zsh 5.9. ZC1001 `$+name[key]`, ZC1083 brace range, ZC1071 array prepend/transform, ZC1715 coprocess `read -p`, ZC1818 `del` name, ZC1856 associative `unset`, ZC1909 bare `exec`, ZC1098 `eval $cmd`, ZC1054 `[0-9]` range, ZC1012 `read -k`/`-q`/`-z`, ZC1094 mid-pipeline `sed`, ZC1017 `print` escape, ZC1037 single-quoted string, ZC1078 `$@` rationale.

## Known follow-ups (not in this PR)

- Issue #1362 (stray closer keywords): closer words are valid command arguments in zsh, so a statement-level guard misfires; needs a command-position grammar change.
- A brace-group body after `(( … )) && { … }` / `[[ … ]] && { … }` is not yet walked (low volume).


redteamx (dismissed):
LGTM.

redteamx (dismissed):
LGTM.

redteamx (dismissed):
LGTM.

redteamx (dismissed):
LGTM.

redteamx (dismissed):
LGTM.

redteamx (dismissed):
LGTM.

redteamx (dismissed):
LGTM.

redteamx (dismissed):
LGTM.

redteamx (dismissed):
LGTM.

redteamx (approved):
LGTM.

redteamx (approved):
Approved on review.

PR #1390, opened 2026-06-15
fix: parse reserved words as command arguments

afadesigns:
## What

In Zsh a reserved closer word (`done`, `fi`, `esac`, `then`, `do`, `else`, `elif`) is a literal word outside command position — `echo done` prints `done`. The parser treated these as command-word delimiters and non-argument peeks, so `echo done` stopped argument gathering at `done` and orphaned it into a second bogus top-level statement.

## Fix

Drop the closer words from the command-delimiter set and accept them as argument-prefix peeks. `echo done` now parses as one command with `done` as its argument. Compound terminators are still recognised through the block-terminator sets, so `if`/loop/`case` bodies are unaffected.

## Validation

- `go test ./...` passes; `golangci-lint` 0 issues; `gocyclo -over 15` clean.
- Parser corpus sweep: 402 files, zero parse errors.
- Violation corpus sweep: baseline unchanged (zero drift).

Groundwork for #1362. The stray-closer detection itself needs the separate compound-terminator-consumption change and is not in this PR.

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -305,9 +305,7 @@ var commandDelimiterTokens = map[token.Type]struct{}{
 	token.EOF: {}, token.SEMICOLON: {}, token.PIPE: {},
 	token.AND: {}, token.OR: {},
 	token.RPAREN: {}, token.RBRACE: {}, token.HASH: {},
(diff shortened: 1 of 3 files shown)


redteamx (approved):
LGTM.

redteamx (approved):
Approved on review.

PR #1391, opened 2026-06-15
fix: parser consumes nested compound terminators

afadesigns:
## What

Each compound command now consumes its own closing keyword, so an inner `fi`, `done`, or `esac` is never mistaken for the enclosing compound's terminator.

## Why

The previous parser left the closer as `curToken` for the caller to advance past. When two compounds nested and shared a closer keyword (`while …; do for …; do …; done; done`), the inner closer was taken as the outer loop's terminator and the real outer closer orphaned to statement-head, building a wrong AST that left the trailing statement unlinted.

## Scope

- `if`, `for`, `while`, `until`, `repeat`, `select` nested inside another loop or conditional.
- A `( … )` subshell on a pipeline right-hand side (`cmd | ( … )`) whose body ends in a compound command no longer swallows the following statement.
- `return` takes its optional value only on the same logical line.

## Verification

- `go test ./...` green; three new regression tests in `parser_corpus_fixes_test.go`.
- Parser-corpus sweep: 402 files, 0 parse errors.
- Violation-corpus sweep: zero drift (7398 findings, unchanged).
- golangci-lint 0 issues; gocyclo clean; coverage 92.1% local (new functions 100%).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.2.5] - 2026-06-15
+
+### Fixed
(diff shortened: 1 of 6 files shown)


redteamx (approved):
LGTM.

PR #1392, opened 2026-06-15
fix: brace-form for loop consumes its own terminator

afadesigns:
## What

The Zsh short-form `for x ( list ) { body }` brace loop now consumes its own closing `}`.

## Why

The loop left `curToken` on its `}` for the caller to advance past. Nested inside another brace block (`if (( c )) { for p ( … ) { … } } else { … }`, the zinit autoload path), the enclosing block mistook the loop's `}` for its own terminator and closed early, orphaning the trailing `else` and building a wrong AST. The `while` and `repeat` brace forms already consume their `}`; this aligns `for`.

## Verification

- `go test ./...` green; new regression test `TestParseShortFormForBraceTerminator`.
- Parser-corpus sweep 402/0; violation-corpus sweep zero drift (7398).
- golangci-lint 0 issues; gocyclo clean; coverage 92.1% local.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.2.6] - 2026-06-15
+
+### Fixed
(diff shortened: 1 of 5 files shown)


redteamx (approved):
LGTM.

PR #1393, opened 2026-06-15
chore: release v1.3.0

afadesigns:
### Added
- `-list-rules` prints every kata (ID, severity, title) and exits.
- `-explain ZC####` prints one kata's full description and exits (case-insensitive).

### Fixed
- ZC1075 no longer flags `${VAR:-word}` / `${VAR:=word}` / `${VAR:+word}` default-value expansions — they supply a value and never elide, so the canonical `: ${VAR:=default}` idiom is no longer warned. The scalar message no longer mislabels an expansion as an array element.
- ZC1107 no longer double-flags `[[ … -gt … ]]` arithmetic comparisons (owned by ZC1091); it keeps the `[ … ]` / `test` form. Each comparison is reported once.
- Documentation drift: auto-fix badge count, INSTALL anchors, SECURITY supported-versions table, info-tier severity example, and inline kata counts.

### Verification
- `go test ./...` green; new regression tests for the CLI flags and the ZC1075 / ZC1107 fixes.
- Parser-corpus sweep 402/0; violation-corpus sweep re-snapshotted (356 false positives removed, no new findings).
- golangci-lint 0 issues; gocyclo clean; coverage 92.1% local; KATAS.md in sync.


redteamx (approved):
LGTM.

PR #1394, opened 2026-06-15
fix: ZC1075 assignment-RHS and ZC1285 sort false positives

afadesigns:
### Fixed
- **ZC1075** no longer flags a concatenated path-style assignment RHS (`x=${a}/${b}`, `arr[k]=$H/$z`). The RHS split at the glued `/`, orphaning the tail into a bogus `/${b}` command that mis-flagged the trailing expansion.
- **ZC1285** no longer flags `sort <file>` with the `${(o)array}` suggestion. Sorting a file's lines cannot be replaced by the in-shell array flag; the idiom only applies to the pipe form.

### Verification
- `go test ./...` green; new regression test for the assignment RHS; ZC1285 test updated.
- Parser-corpus sweep 402/0; violation-corpus sweep no drift (baseline re-snapshotted for the assignment-RHS reductions).
- golangci-lint 0; gocyclo clean; coverage 92.1% local.

### DIFF
--- a/.github/violation-baseline.txt
+++ b/.github/violation-baseline.txt
@@ -36,7 +36,7 @@ antidote/antidote.zsh	ZC1059	1
 antidote/antidote.zsh	ZC1065	1
 antidote/antidote.zsh	ZC1071	1
 antidote/antidote.zsh	ZC1073	2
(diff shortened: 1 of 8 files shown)


redteamx (approved):
Approved on review.

PR #1395, opened 2026-06-16
fix: dedup kata overlap and ZC1075 split-flag false positives

afadesigns:
### Changed — one recommendation per construct
- **echo**: ZC1037 (`print -r --`, carries the auto-fix) is the single canonical recommendation. ZC1030 (printf) and ZC1092 (prefer print) defer — no more 2-3 divergent suggestions per echo.
- **single-bracket arithmetic** (`[ $a -eq $b ]`): ZC1003 (`(( … ))`, with fix) is canonical. ZC1107 defers; ZC1010 defers when an arithmetic operator is present.

### Fixed
- **ZC1075 split-flags**: no longer flags `${(f)x}` / `${(s:,:)x}` / `${(@)arr}` / `${(kv)map}` — they expand to multiple words; quoting defeats the idiom. The parser now preserves the `(flags)` group on `ast.ArrayAccess`.
- **glued ident path tail**: `x=${a}/init` is no longer split into two statements.

### Verification
- `go test ./...` green; new regression tests for each fix.
- Parser-corpus sweep 402/0; violation-corpus sweep re-snapshotted (net ~270 redundant/false findings removed this release; ~824 cumulative since v1.3.0).
- golangci-lint 0; gocyclo clean; coverage 92.1%; KATAS.md in sync.


redteamx (dismissed):
Approved on review.

redteamx (approved):
Approved on review.

PR #1396, opened 2026-06-16
fix: repair redirect chains and stray closers

afadesigns:
Parser correctness fixes for redirections, orphan closers, and expression-led pipelines.

What:
- A command followed only by redirects parses as one statement; a chain like `cmd >/dev/null 2>&1`, `cmd >> log`, `cmd >& fd`, or `cmd <& 3` no longer orphans its target word into a bogus second statement.
- `&>file` lexes as the combined stdout+stderr redirect instead of a background `&` plus an orphaned `>`.
- An expression-led pipeline (`$cmd && echo hello`) gathers the right-hand command arguments.
- A leading brace-group condition (`if { cmd }; then ...; fi`) is no longer mistaken for the brace-form body opener.
- An orphan `fi`/`done`/`esac` at the top level is reported as a parse error instead of being silently accepted.

Why: the expression path consumed only one redirect via the infix table and `&>` was mis-lexed, desyncing real scripts. Recovered findings on commands that carry a redirect; removed false positives on redirect targets (Zsh does not word-split them).

Severity: parser correctness.


redteamx (approved):
Approved on review.

PR #1397, opened 2026-06-16
fix: parse zsh always-block as one statement

afadesigns:
Zsh `{ try } always { always-list }` parsed the `always` clause as a separate `always { ... }` command and stranded its closing brace as its own statement.

What: fold the always-list into the preceding block so its commands are linted in place; no stray brace statement.
Why: `always` is a real Zsh construct used by integration corpora (zaw, zsh-defer). It is lexed as a plain identifier and is valid only after a brace group, so the keyword is unambiguous.

Severity: parser correctness.


redteamx (dismissed):
Approved on review.

PR #1398, opened 2026-06-16
fix: stop auto-fix from corrupting source

afadesigns:
A full-corpus auto-fix round-trip (apply `-fix`, re-parse) found 14 files the fixer corrupted into unparseable source and one it rewrote without converging.

What:
- Safety net in the fix loop: a pass that raises the parser-error count falls back to applying only individually safe edits, and skips a still-conflicting set rather than writing broken source. Auto-fix can no longer increase parse errors.
- ZC1001 emits one replacement spanning `$name[subscript]` instead of separate `{` / `}` insertions, so it overlaps (and the engine resolves) a competing edit on the same token. Fixes the broken `{name[subscript]}` produced when ZC1073 deleted the `$` inside `(( … ))`.
- ZC1040 `(N)` nullglob fix no longer stacks qualifiers on repeated runs.

Why: a linter that rewrites code must never corrupt it. Verified: 1095-corpus round-trip now reports broke=0 unstable=0 (was 14 + 1).

Severity: auto-fix correctness.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.3.5] - 2026-06-16
+
+### Fixed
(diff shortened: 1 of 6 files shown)


redteamx (approved):
Approved on review.

PR #1399, opened 2026-06-16
refactor: apply safe fixes incrementally

afadesigns:
Follow-up to the auto-fix safety net. The fallback that filtered individually-safe edits then re-applied them had an unreachable combined-recheck branch (an edit valid alone cannot fail the combined apply with identical offsets), which left a line uncovered.

Replace it with a single highest-offset-first pass: apply each edit to the accumulated source, keep it only when it does not raise the parser-error count. Highest-offset-first means accepting an edit never shifts the line/column of the edits not yet tried. Same guarantee (auto-fix never increases parse errors), simpler, fully covered.

Severity: internal refactor, no behavior change.

### DIFF
--- a/cmd/zshellcheck/main.go
+++ b/cmd/zshellcheck/main.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime/pprof"
+	"sort"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


redteamx (approved):
Approved on review.

PR #1400, opened 2026-06-16
build: drop all third-party dependencies

afadesigns:
ZShellCheck now compiles against the Go standard library alone — zero third-party dependencies.

What:
- `github.com/adrg/xdg` -> a small internal XDG config-path resolver (`$XDG_CONFIG_HOME` or `~/.config`, then `$XDG_CONFIG_DIRS`). This also drops the transitive `golang.org/x/sys`.
- `gopkg.in/yaml.v3` -> a parser for the existing flat `.zshellcheckrc` format: the `disabled_katas` sequence plus scalar colour/bool keys, with `#` comments, single/double quotes, and standard escapes.

Why: a static analyser that ships as a single binary should carry no supply-chain surface. The module graph is now one node; `go list -m all` reports only the project itself.

Compatibility: existing config files parse identically (verified end-to-end). New parser is 100% covered.

Severity: build / supply-chain.


redteamx (approved):
Approved on review.

redteamx (dismissed):
Approved on review.

redteamx (approved):
Approved on review.

PR #1401, opened 2026-06-16
fix: emit one document for multi-file json and sarif

afadesigns:
Dogfooding a real config surfaced this: `zshellcheck --format=json a.zsh b.zsh` printed one array per file, concatenated into invalid JSON. SARIF had the same flaw and additionally carried no locations — useless for GitHub code scanning.

What:
- JSON/SARIF now accumulate findings across every scanned file and emit a single valid document.
- JSON gains a `File` field per finding (single-file consumers unaffected).
- SARIF results now carry a physical location (file URI + 1-based line/column) and a per-severity level.
- A clean scan emits an empty document (`[]` / empty run) instead of nothing.
- The per-file JSONReporter/SarifReporter are replaced by aggregating ReportJSON/ReportSARIF (100% covered).

Why: a linter whose machine output cannot be parsed by more than one file at a time breaks every CI integration, SARIF code scanning included.

Severity: output correctness.


redteamx (approved):
Approved on review.

PR #1402, opened 2026-06-16
feat: add fixable markers and -statistics

afadesigns:
Two of Ruff's volume-management UX wins, the answer to "thousands of findings is overwhelming".

What:
- Findings from auto-fixable katas are tagged `[*]` in the text report; the run ends with `[*] N fixable with the -fix option.` So you see how much `-fix` resolves before running it.
- `-statistics` prints a per-kata count sorted by frequency (with the `[*]` marker) instead of individual reports — triage which checks dominate.

Backed by `KatasRegistry.IsFixable` and `TextReporter.MarkFixable`. New code covered; docs + help + CHANGELOG updated.

Severity: feature (UX).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.4.0] - 2026-06-16
+
+### Added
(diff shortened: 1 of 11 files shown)


redteamx (dismissed):
Approved on review.

redteamx (approved):
Approved on review.

PR #1403, opened 2026-06-16
feat: split auto-fixes into safe and unsafe tiers

afadesigns:
Answers "are all autofixes semantically verified": no — most change behavior, and `-fix` was applying them by default.

What:
- Audited all 122 fix functions. Only 8 are purely value-preserving (safe): ZC1001, ZC1002, ZC1073, ZC1086, ZC1411, ZC1502, ZC1637, ZC1643. The rest swap commands/flags, change scope, or add glob qualifiers — unsafe.
- `-fix` now applies safe fixes only; `-unsafe-fixes` applies the rest.
- Each FixEdit carries its kata ID so edits can be filtered. The report counts the tiers separately: `[*] N fixable with -fix` then `M more fixable with -fix -unsafe-fixes`, and `[*]` marks only what the current command applies.

Why: `-fix` was silently rewriting `which`->`whence`, `netstat`->`ss`, injecting `-y`, etc. Behavior-changing rewrites must be opt-in.

Note: this changes `-fix` default behavior (fewer fixes applied), hence the minor bump to v1.5.0.

Severity: feature + safety.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.5.0] - 2026-06-16
+
+### Added
+- `-unsafe-fixes` applies fixes that may change runtime behavior. Without it, `-fix` now applies only safe (value-preserving) fixes.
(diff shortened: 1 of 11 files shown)


redteamx (approved):
Approved on review.

PR #1404, opened 2026-06-16
feat: embed rule metadata in SARIF output

afadesigns:
Clears task #37. The SARIF run now carries a `rules` array in its tool driver: each kata that produced a finding contributes a descriptor (title, full description, default level, help URL), and every result references its rule by index.

GitHub code scanning renders rule metadata, so a finding shows what the rule means and links to the kata catalog instead of showing a bare ID.

Severity: feature (SARIF completeness).


redteamx (approved):
Approved on review.

redteamx (approved):
Approved on review.

PR #1405, opened 2026-06-16
feat: add a baseline ratchet

redteamx:
Clears task #35. `-baseline-write <path>` snapshots current findings; `-baseline <path>` suppresses those and reports only findings new since the snapshot, exiting non-zero only on new ones.

Lets a large existing codebase adopt ZShellCheck and fail CI on regressions without fixing every pre-existing finding first.

Fingerprint = kata + file + trimmed source line (not line number), so unrelated line moves do not resurrect a suppressed finding. New code 100% covered; docs + help + CHANGELOG updated.

Severity: feature (adoption).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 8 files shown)


afadesigns (approved):
Approved on review.

PR #1407, opened 2026-06-16
feat: rule re-grade, add-noka, stale-noka detection

afadesigns:
Completes the volume-management surface with three suppression/severity tools.

- `-rule-severity ZC####:level,...` re-grades chosen katas (error/warning/info/style), independent of the `-severity` floor.
- `-add-noka` appends a `# noka: ZC####` directive to every line with a finding and writes the files — bulk-silence an existing codebase for review.
- `-detect-stale-noka` reports `# noka` directives that suppress no finding and exits non-zero, so stale suppressions can be pruned.

New code covered; docs, help, and CHANGELOG updated; zero doc drift verified. Uses `# noka` (no-katas) — never noqa.

Severity: feature (UX).

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -77,8 +77,9 @@ linters:
           - gosec
         text: "G101"
       # Lint pipeline writes the rewritten file at the original path; the
-      # path comes from the user's CLI arg, validated upstream.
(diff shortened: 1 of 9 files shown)


redteamx (approved):
Approved on review.

PR #1410, opened 2026-06-25
docs: document -no-banner flag in CLI reference

afadesigns:
## What

- Add the `-no-banner` flag to the USER_GUIDE CLI reference table. The flag is defined and used in a combos example, but was missing from the flag list.
- Remove duplicate SPDX header lines from `pkg/fix/fix.go` and `internal/tools/gen-katas-md/main.go`. The header now appears once, at the top of each file.
- Group the Sublime editor ignores with the existing editor and IDE ignores in `.gitignore`.

## Why

A reader of the CLI reference table could not find `-no-banner` even though the same guide uses it in an example. The SPDX duplication and ignore placement are housekeeping.

## Verification

- `go build ./...`, `go vet ./...`, `go test -count=1 ./...` pass.
- `golangci-lint run ./...` reports 0 issues.
- `gocyclo -over 15 .` clean.
- KATAS.md in sync; version parity at v1.7.0.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,8 @@ zshellcheck-windows-amd64.exe
 # Editor and IDE files
 .vscode/
 .idea/
+*.sublime-project
+*.sublime-workspace
 *.swp
 *~
 .#*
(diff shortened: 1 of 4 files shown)


redteamx (approved):
Approved on review.

PR #1411, opened 2026-06-25
ci: correct stale action-pin version comments

afadesigns:
## What

Correct version comments on pinned action SHAs that no longer matched the tag the SHA points to. The SHAs are unchanged; only the trailing `# vX.Y.Z` comments are corrected.

| Action | Pinned SHA | Was | Now |
| --- | --- | --- | --- |
| `actions/checkout` | `9c091bb` | `# v4.1.7` | `# v7.0.0` |
| `actions/dependency-review-action` | `a1d282b` | `# v4` | `# v5.0.0` |
| `github/codeql-action` (init/autobuild/analyze) | `5c8a8a6` | `# v3` | `# v3.35.1` |
| `ludeeus/action-shellcheck` | `00cae50` | `# v2.0.0` | `# 2.0.0` |

## Why

The pin comment is the human-readable record of which release a SHA represents. After prior SHA bumps the comments were left stale, so the `actions/checkout` pin read `# v4.1.7` while pinning the v7.0.0 commit, and `dependency-review-action` read `# v4` while pinning v5.0.0. A comment that misstates the version defeats the purpose of the pin annotation.

## Verification

- Each comment was resolved against the upstream tag with `git ls-remote --tags`; every pin now matches.
- Diff is comment-only: no SHA changed, no step behavior changed.
- All workflow YAML parses; the actionlint job validates syntax.

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -27,6 +27,6 @@ jobs:
       - uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
         with:
           egress-policy: audit
-      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v4.1.7
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       - name: Lint workflow files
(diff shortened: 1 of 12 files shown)


redteamx (approved):
Approved on review. Comment-only pin corrections; SHAs verified unchanged against upstream tags.

PR #1413, opened 2026-06-25
test: add autofix-safety corpus round-trip gate

afadesigns:
## What

Make autofix safety a permanent, enforced invariant and close the per-kata fix-test gap.

- `scripts/fix-corpus-sweep.sh` — the canonical autofix-safety gate, wired into the corpus CI job. For every pinned-corpus file it applies `-fix` (safe) and `-fix -unsafe-fixes` (every tier), then asserts:
  1. No corruption — the rewrite never raises the file's own parser-error count.
  2. Idempotent — a second `-fix` pass produces no further change.
  3. Panic-free.
  No baseline: fix safety is an absolute invariant, so any violation fails.
- `pkg/fix/corpus_roundtrip_test.go` — an always-on (`go test`) round-trip over the historically risky fix interactions: the subscript-vs-arithmetic collision (ZC1001/ZC1073), glob-qualifier stacking (ZC1040), the echo cluster, and nested backticks.
- `pkg/fix/integration_gap_test.go` — isolated golden fix-output tests for the seven fixable katas that previously had only detection coverage: `ZC1006`, `ZC1008`, `ZC1020`, `ZC1022`, `ZC1036`, `ZC1037`, `ZC1217`.

## Why

The full-corpus fix round-trip is the method that found the three destructive-fix bugs in v1.3.5 (a subscript/arithmetic edit collision, a non-converging glob qualifier, and the engine safety net). It was run by hand and never gated, so the class could regress silently. This makes it a permanent gate and adds an always-on in-repo layer plus the missing per-kata goldens.

## Verification

- Local sweep over all 1095 corpus files: zero corruptions, zero non-idempotent rewrites, zero panics.
- `go test ./...` passes; `golangci-lint run ./...` reports 0 issues.
- The new CI step runs the sweep against the pinned corpora.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,6 +101,10 @@ jobs:
         env:
           ZSHELLCHECK_BIN: ${{ github.workspace }}/zshellcheck
         run: bash scripts/violation-corpus-sweep.sh
+      - name: Sweep auto-fixes for corruption and non-convergence
+        env:
+          ZSHELLCHECK_BIN: ${{ github.workspace }}/zshellcheck
+        run: bash scripts/fix-corpus-sweep.sh
 
   lint-shell:
     name: Lint Shell Scripts
--- a/README.md
+++ b/README.md
@@ -128,10 +128,11 @@ The full catalog with file counts lives in [INTEGRATIONS.md](INTEGRATIONS.md).
 
 ## Quality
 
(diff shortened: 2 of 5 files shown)


redteamx (approved):
Approved on review. Fix-safety gate verified clean across 1095 corpus files; in-repo round-trip and isolated goldens added.

PR #1414, opened 2026-06-25
test: guard every fixable kata has a golden test

afadesigns:
## What

Add a coverage guard: `TestEveryFixableKataHasGoldenTest` enumerates every kata that ships a `Fix` and fails if it has no golden fix-output test.

## Why

The per-kata golden tests pin each fix's exact rewrite, but nothing stopped a new fixable kata from shipping without one. The guard reads the registry (the source of truth for what is fixable) and the golden test files (the source of truth for what is covered), so it cannot drift from a hand-maintained list and cannot pass on a kata whose test was deleted. This extends the project's "no new public API without a test" rule to the auto-fixer.

## Verification

- Passes now: all 132 fixable katas have a golden test.
- Bites when coverage is incomplete: removing a golden file from the scan reports the exact uncovered kata IDs and fails.
- `go test ./...` passes; `golangci-lint run ./...` reports 0 issues.

### DIFF
--- a/pkg/fix/fixable_coverage_test.go
+++ b/pkg/fix/fixable_coverage_test.go
@@ -1,0 +1,58 @@
+// SPDX-License-Identifier: MIT
+// Copyright the ZShellCheck contributors.
+package fix
+
+import (
+	"os"
+	"regexp"
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/katas"
+)
+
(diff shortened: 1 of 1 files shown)


redteamx (approved):
Approved on review. Coverage guard verified: passes at 132/132, fails listing exact gaps when a golden is missing.

PR #1415, opened 2026-06-25
fix: stop auto-fixes emitting broken code

afadesigns:
## What

Stop three auto-fixes from rewriting valid Zsh into broken or wrong source. Found by auditing real-corpus findings against runnable Zsh 5.9.

| Kata | Tier | Before | After |
| --- | --- | --- | --- |
| `ZC1086` | **safe** | `function name"${1:-}"() {}` -> `name()"${1:-}"() {}` (defines a different function) | bails when the name continues past the parsed identifier |
| `ZC1010` | unsafe | `[ x -o y ]` -> `[[ x -o y ]]` (syntax error: `[[ ]]` has no binary `-o`/`-a`) | bails on binary `-a`/`-o`; unary `[ -o opt ]` still converts |
| `ZC1076` | unsafe | `autoload -U …` -> `autoload -Uz -U …` (duplicate flag) | inserts only the absent flag |

## Why

`ZC1086` is the most serious: it is in the **safe** (value-preserving) tier, yet it silently changed which function a definition creates. With `$1=X`, `gitstatus_query"${1:-}"` defines `gitstatus_queryX`; the fix made it define `gitstatus_query`, breaking gitstatus's per-instance namespacing. A safe fix must never change behavior. `ZC1010` produced source that Zsh rejects outright (`condition expected`). `ZC1076` was merely sloppy but still wrong.

These are parseable-but-wrong rewrites, so the parser-level fix-safety sweep does not catch them — they need semantic review, which is how they surfaced.

## Verification

- Each rewrite confirmed against real `zsh`: the old output defines the wrong function / errors / duplicates; the new output is correct.
- Detection is unchanged for all three — the violation baseline holds.
- New regression tests in `pkg/fix/fix_regression_test.go`.
- `go test ./...` passes, `golangci-lint run ./...` reports 0 issues, `gocyclo` clean, fix-corpus sweep stays 0/0/0 across 402 files.

### DIFF
--- a/pkg/fix/fix_regression_test.go
+++ b/pkg/fix/fix_regression_test.go
@@ -1,0 +1,58 @@
+// SPDX-License-Identifier: MIT
+// Copyright the ZShellCheck contributors.
+package fix
+
+import (
+	"strings"
+	"testing"
+)
+
+// These regressions pin three auto-fixes that previously emitted broken
+// or wrong source. Each asserts the fix no longer corrupts its input.
+// They were found by auditing real-corpus findings against runnable Zsh.
+
+func TestFixRegression_ZC1086_NamespacedFunctionName(t *testing.T) {
+	// `function name"${1:-}"() { ... }` builds the function name from a
+	// parameter expansion (gitstatus does this for per-instance
+	// namespacing). The fix used to insert `()` after the parsed
+	// identifier, splitting the real name so a different function was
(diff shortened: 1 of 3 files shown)


redteamx (dismissed):
Approved on review. Three corrupting auto-fixes confirmed against real zsh; detection unchanged, regression tests added.

PR #1416, opened 2026-06-25
fix: drop false positives in ZC1097 and ZC1049

afadesigns:
## What

Stop two katas from flagging code where the advice is wrong or impossible. Found by auditing real-corpus findings against runnable Zsh.

- **ZC1097** suggested `local` for a positional loop variable (`for 1 in …`). `local 1` is a zsh error (`not an identifier: 1`), and positionals are already function-local. Now skips all-numeric loop-variable names.
- **ZC1049** flagged the alias listing/query forms (`alias`, `alias -L`, `alias name`, `alias | …`) with "prefer functions" — impossible, since they print existing aliases rather than define one. Now fires only on a real `name=value` definition, including a parameter-expansion name (`alias ${CMD}=val`) and a dashed name (`alias -- -='cd -'`).

## Why

Both produced noise that a user cannot act on: ZC1097's suggestion errors when applied, and ZC1049 cannot be satisfied for a listing command. The ZC1049 fix was iterated twice against the corpus drift — the first cut wrongly suppressed `alias -- -='cd -'` and `alias ${CMD}=val`, which the baseline review caught.

## Verification

- Every behavior confirmed against real `zsh`.
- ZC1049 removes 12 listing-form findings from the violation baseline across six corpora; every removed line is a listing/query form and every kept line is a real definition (reviewed individually).
- Regression tests in `pkg/katas/katatests/fp_round2_test.go` and `pkg/katas/zc1097_fp_test.go`.
- `go test ./...` passes, `golangci-lint` 0 issues, `gocyclo` clean, fix-corpus sweep stays 0/0/0.

### DIFF
--- a/.github/violation-baseline.txt
+++ b/.github/violation-baseline.txt
@@ -226,7 +226,6 @@ fzf/shell/completion.zsh	ZC1003	3
 fzf/shell/completion.zsh	ZC1010	10
 fzf/shell/completion.zsh	ZC1043	4
 fzf/shell/completion.zsh	ZC1046	4
-fzf/shell/completion.zsh	ZC1049	1
 fzf/shell/completion.zsh	ZC1064	1
 fzf/shell/completion.zsh	ZC1074	1
 fzf/shell/completion.zsh	ZC1075	8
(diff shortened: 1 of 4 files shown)


redteamx (approved):
Approved on review. Two false positives fixed, each confirmed against real zsh; baseline drift reviewed line-by-line, regression tests added.

PR #1417, opened 2026-06-25
fix: drop false positives in ZC1043

afadesigns:
## What

Stop ZC1043 ("use `local` for function variables") from flagging code where `local` is wrong, found by auditing real-corpus findings against runnable Zsh.

- A variable declared with a flagged `declare`/`typeset`/`local` (e.g. `declare -a versions`) parses as a `DeclarationStatement`, not a command, so the local-harvester missed it and a later `versions=(…)` was reported as an accidental global. The harvester now reads `DeclarationStatement` names too.
- Special tied and shell-state parameters (`path`/`PATH`, `fpath`, `cdpath`, `manpath`, `options`, the alias/function hashes, `dirstack`, …) are global by nature. Assigning them inside a function mutates the shell-wide value on purpose; `local fpath` scopes it away and silently breaks plugin load/unload. They are now skipped.

## Why

Both produced unactionable noise. The `declare -a` case is a plain miss in the scope tracker. The special-parameter case is worse than noise — following the advice (`local fpath`) breaks plugin managers, which is exactly why it must not be suggested.

## Verification

- Confirmed against real `zsh`: a `local fpath` provably loses the intended global mutation (`fpath` reverts on function return).
- Removes 62 findings from the violation baseline across 18 corpora; every removed line is a declared local (`declare`/`typeset`) or a special parameter, reviewed file by file. Every kept finding is a real bare unscoped assignment.
- Regression tests in `pkg/katas/katatests/fp_round3_test.go`; `go test ./...` passes, `golangci-lint` 0 issues, `gocyclo` clean, fix-corpus sweep stays clean.

### DIFF
--- a/.github/violation-baseline.txt
+++ b/.github/violation-baseline.txt
@@ -2,7 +2,6 @@
 # Snapshot of kata findings on the pinned, zero-parser-error corpora. Any drift is reviewed.
 # Format: <relpath>\t<ZCID>\t<count>
 agkozak-prompt/agkozak-zsh-prompt.plugin.zsh	ZC1017	1
-agkozak-prompt/agkozak-zsh-prompt.plugin.zsh	ZC1043	1
 agkozak-prompt/agkozak-zsh-prompt.plugin.zsh	ZC1075	6
 agkozak-prompt/agkozak-zsh-prompt.plugin.zsh	ZC1083	1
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review. ZC1043 declare/typeset + special-tied-param FPs fixed; 62 baseline removals reviewed file by file, confirmed against real zsh.

PR #1418, opened 2026-06-25
fix: ZC1098 skips the eval "$@" dispatch idiom

afadesigns:
## What

Stop ZC1098 ("use `(q)` for variables in `eval`") from flagging the `eval "$@"` argument-dispatch idiom.

`eval "$@"` and `eval "$*"` run the positional parameters as the command. The kata already skips the var-as-command forms (`eval "$cmd"`, `eval "$REPLY"`), but `zc1098IsBareName` rejected the `@`/`*` sigils, so `eval "$@"` slipped through and was flagged.

## Why

Following the advice breaks the code. `eval "${(q)@}"` collapses the arguments into one quoted literal, so a worker dispatched with `set -- "echo a && echo b"` fails with `command not found`. Confirmed against real zsh. The idiom is used by the zsh-async library and every prompt that vendors it.

A variable used as data inside a command string (`eval "echo $userdata"`, `eval "rm $path"`) still warrants the `(q)` advice and continues to fire.

## Verification

- Confirmed against real `zsh`: `eval "$@"` runs the command; `eval "${(q)@}"` errors.
- Removes 5 findings from the violation baseline (zsh-async and its forks: agkozak, purer, spaceship, typewritten); every removed line is an `eval "$@"` dispatch.
- Regression test in `pkg/katas/katatests/fp_round4_test.go`; `go test ./...` passes, `golangci-lint` 0 issues, sweeps clean.

### DIFF
--- a/.github/violation-baseline.txt
+++ b/.github/violation-baseline.txt
@@ -11,7 +11,6 @@ agkozak-prompt/lib/async.zsh	ZC1001	5
 agkozak-prompt/lib/async.zsh	ZC1043	2
 agkozak-prompt/lib/async.zsh	ZC1046	1
 agkozak-prompt/lib/async.zsh	ZC1075	30
-agkozak-prompt/lib/async.zsh	ZC1098	1
 agkozak-prompt/lib/async.zsh	ZC1134	1
 agkozak-prompt/lib/async.zsh	ZC1176	1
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review. eval "$@" dispatch FP confirmed against real zsh; 5 baseline removals, regression test added.

PR #1419, opened 2026-06-25
fix: ZC1188 gives direction-aware PATH advice

afadesigns:
## What

Make ZC1188's `path` advice match the direction of the PATH edit.

The kata flagged every `export PATH=…` with one hardcoded suggestion, `path+=(dir)`. That is right only for an **append** (`$PATH:dir`). For a **prepend** (`dir:$PATH`) — used to give a directory precedence — `path+=(dir)` appends instead, moving the directory to the end and changing which binary runs.

The message now adapts:

| Edit | Advice |
| --- | --- |
| `export PATH=dir:$PATH` (prepend) | `path=(dir $path)` |
| `export PATH=$PATH:dir` (append) | `path+=(dir)` |
| `export PATH=dir` (other) | both forms |

## Why

Following the old advice on a prepend silently reverses command precedence — confirmed against real zsh, where a prepend and a `path+=` resolve to different binaries first. Twenty-six of the corpus hits are prepends.

## Verification

- Direction-aware messages verified on prepend, append, and plain assignments.
- ZC1188 is advisory-only (no auto-fix), so detection counts and the violation baseline are unchanged.
- Regression test in `pkg/katas/katatests/fp_round5_test.go`; the existing ZC1188 message test updated; `go test ./...` passes, `golangci-lint` 0 issues, sweeps clean.

### DIFF
--- a/pkg/katas/katatests/fp_round5_test.go
+++ b/pkg/katas/katatests/fp_round5_test.go
@@ -1,0 +1,38 @@
+// SPDX-License-Identifier: MIT
+// Copyright the ZShellCheck contributors.
+package katas
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/afadesigns/zshellcheck/pkg/testutil"
+)
+
+// TestZC1188DirectionAwareAdvice pins the PATH-direction advice. A
+// prepend (`dir:$PATH`) must be rewritten as `path=(dir $path)`; the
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review. Wrong-direction PATH advice confirmed against real zsh; message-only change, baseline unchanged, regression test added.

PR #1420, opened 2026-06-25
fix: ZC1044 skips cd= variable assignment

afadesigns:
## What

Stop ZC1044 ("use `cd … || return`") from flagging an assignment to a variable named `cd`.

`(( flag )) && cd=${cd/#$HOME/~}` assigns to a variable `cd`. On the right of `&&`/`||` the parser reads `cd=value` as a `cd` command whose first argument is the `=value` tail, so the kata flagged it as an unguarded directory change. The cd builtin never takes an argument beginning with `=`, so this shape is now recognised as the assignment and skipped.

## Why

The advice is meaningless on an assignment — there is no directory change to guard. The root cause is a parser quirk: an assignment word is recognised at statement start but parsed as a command after `&&`/`||`. The full parser fix is broader and risk-bearing (it shifts how every `cmd && var=value` parses), so this guards the kata precisely while the parser fix is tracked separately.

## Verification

- AST confirmed: `foo && cd=$bar` parses to `SimpleCommand{Name: cd, Args: ["=$bar"]}`; a real `cd /tmp` has no `=`-prefixed argument.
- Removes 1 finding from the violation baseline (zsh-z); a real unguarded `cd`, including `foo && cd /tmp`, still fires.
- Regression test in `pkg/katas/katatests/fp_round6_test.go`; `go test ./...` passes, `golangci-lint` 0 issues, sweeps clean.

### DIFF
--- a/.github/violation-baseline.txt
+++ b/.github/violation-baseline.txt
@@ -1602,7 +1602,6 @@ zsh-vi-mode/zsh-vi-mode.zsh	ZC1477	2
 zsh-z/zsh-z.plugin.zsh	ZC1001	1
 zsh-z/zsh-z.plugin.zsh	ZC1004	2
 zsh-z/zsh-z.plugin.zsh	ZC1043	3
-zsh-z/zsh-z.plugin.zsh	ZC1044	1
 zsh-z/zsh-z.plugin.zsh	ZC1049	1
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review. cd= assignment FP confirmed via AST; kata-level guard, parser quirk noted separately. Regression test added.

PR #1421, opened 2026-06-25
fix: ZC1075 skips flag-led expansions

afadesigns:
## What

Stop ZC1075 ("quote this expansion") from flagging a flag-led parameter expansion the parser cannot fully model.

`${(%):-default}` parses to an expansion with a `%` flag, a nil subject (`Left`), and `${ }` as rendered text — the parser drops the `:-default`. ZC1075 then flagged it as an elision hazard. But a `(%)` prompt expansion with a `:-` default never yields an empty word, so the advice is wrong.

When the parser could not resolve the subject (`Left` is nil), ZC1075 now skips the expansion instead of guessing.

## Why

The flagged forms (`${(%):-…}`, `${(P)…}`) carry a default or width modifier that guarantees a non-empty result, so quoting against elision is meaningless. Without the parsed subject, the kata has no basis to judge elision, so skipping is the conservative correct choice. Confirmed against real zsh: `${(%):-fallback}` always prints `fallback`.

## Verification

- Flag-led forms now skip; ordinary scalars (`$plain`) and array elements (`${arr[1]}`) still fire; normal `${var:-default}` and split flags `${(f)x}` keep their existing skips.
- Removes 1 finding from the violation baseline (zinit, `read -q ${(%):-"?…"}`).
- Regression test in `pkg/katas/katatests/fp_round7_test.go`; `go test ./...` passes, `golangci-lint` 0 issues, sweeps clean.

### DIFF
--- a/.github/violation-baseline.txt
+++ b/.github/violation-baseline.txt
@@ -1248,7 +1248,7 @@ zinit/zinit-autoload.zsh	ZC1049	1
 zinit/zinit-autoload.zsh	ZC1053	1
 zinit/zinit-autoload.zsh	ZC1064	1
 zinit/zinit-autoload.zsh	ZC1073	3
-zinit/zinit-autoload.zsh	ZC1075	41
+zinit/zinit-autoload.zsh	ZC1075	40
 zinit/zinit-autoload.zsh	ZC1083	2
 zinit/zinit-autoload.zsh	ZC1091	32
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review. Flag-led-expansion FP confirmed; nil-subject skip, baseline drift reviewed (zinit ${(%):-} prompt-default), regression test added.

PR #1422, opened 2026-06-25
fix: ZC1075 skips bare expansions of arrays

afadesigns:
## What

Stop ZC1075 from flagging a bare expansion of a variable that holds an **array**. The kata's premise — an unquoted empty value elides and drops the word — is a scalar concern. For an array, quoting `$arr` as `"$arr"` *joins* the elements into one word; the correct form is `"${arr[@]}"`, which the message already states.

ZC1075 moves from a per-command check to a program-level one:
1. Harvest every array name in the file (from `name=(…)` array literals, `local -a`/`typeset -A`/`declare -a`/`readonly -a` declarations, and `local opts=(…)`).
2. Flag only bare scalar expansions; skip a bare `$arr` / `${arr}` whose name was harvested as an array.

## Why this is careful

Array detection reads the **AST**, not rendered text. The renderer prints a scalar value `$1` as `name=($1)` (spurious parens), so a string-based `=(` check would misclassify scalars as arrays. The harvest inspects the parsed value node (an `ArrayLiteral` part) and the declaration flags instead, so a scalar `local worker=$1` is never mistaken for an array.

This was prototyped, found to over-suppress via the renderer trap, reverted, and re-done against the AST — the over-suppression is gone (`local worker=$1` still fires).

## Verification

- Removes 121 findings across 33 corpora; spot-checked the harvested set in the biggest-drop files (`async.zsh`, `zsh-vi-mode`, `gitstatus`) — every removed expansion names a genuinely array-declared variable.
- Scalars (`local worker=$1`, `local x=$(cmd)`, `rm $file`), array elements `${arr[i]}`, and the flag-led / default-modifier skips are all unchanged.
- `go test ./...` passes, `golangci-lint` 0 issues, `gocyclo` clean, the new harvest helpers are 100% covered, fix-corpus sweep stays clean.

### DIFF
--- a/.github/violation-baseline.txt
+++ b/.github/violation-baseline.txt
@@ -10,7 +10,7 @@ agkozak-prompt/agkozak-zsh-prompt.plugin.zsh	ZC1604	2
 agkozak-prompt/lib/async.zsh	ZC1001	5
 agkozak-prompt/lib/async.zsh	ZC1043	2
 agkozak-prompt/lib/async.zsh	ZC1046	1
-agkozak-prompt/lib/async.zsh	ZC1075	30
+agkozak-prompt/lib/async.zsh	ZC1075	24
 agkozak-prompt/lib/async.zsh	ZC1134	1
 agkozak-prompt/lib/async.zsh	ZC1176	1
 agkozak-prompt/lib/async.zsh	ZC1509	2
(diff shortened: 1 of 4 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
Approved on review. ZC1075 array rework: AST-based harvest (avoids the renderer paren trap), over-suppression verified gone, 121 baseline removals spot-checked, helpers 100% covered.

PR #1423, opened 2026-06-25
fix: ZC1075 skips width-pad and typeset arrays

afadesigns:
## What

Two ZC1075 elision refinements.

- A width modifier `${(l:n:)x}` / `${(r:n:)x}` pads to a fixed width — never an empty word. The `l`/`r` flag is not a word-splitting flag, so it slipped past the existing skip. Now skipped, distinguished from `(r)` (reverse, can be empty) by the `:n:` width argument.
- `typeset arr=(a b)` declares an array, but the parser does not model its value as an `ArrayLiteral` here. The harvest now also reads a value that renders with a leading `(`; a quoted scalar `q="(literal)"` keeps its opening quote and is not mistaken for an array.

## Verification

- `${(l:5:)x}` confirmed against real zsh: an unset value yields five spaces, never empty.
- Removes 2 findings from the violation baseline (gitstatus `${(l:20:)pgid}`, spaceship `typeset -U sections=(...)`); `(r)` reverse and quoted scalars still fire.
- `go test ./...` passes, `golangci-lint` 0 issues, new helpers 100% covered, sweeps clean.

### DIFF
--- a/.github/violation-baseline.txt
+++ b/.github/violation-baseline.txt
@@ -324,7 +324,7 @@ git-open/git-open.plugin.zsh	ZC1275	1
 gitstatus/gitstatus.plugin.zsh	ZC1001	7
 gitstatus/gitstatus.plugin.zsh	ZC1043	3
 gitstatus/gitstatus.plugin.zsh	ZC1046	1
-gitstatus/gitstatus.plugin.zsh	ZC1075	61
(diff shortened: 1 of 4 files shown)


redteamx (approved):
Approved on review. Width-pad + typeset-array skips verified against zsh; 2 baseline removals, helpers 100% covered.

redteamx (dismissed):
Approved on review.

PR #1424, opened 2026-06-25
fix: retire ZC1104, duplicate of ZC1188

afadesigns:
## What

ZC1104 and ZC1188 both fired on `export PATH=…` with the same advice — every such line was double-reported. ZC1188 now owns that detection (its message adapts to a prepend or an append), so ZC1104 is stubbed to a no-op, following the retired-duplicate pattern already used for ZC1018/ZC1019. Legacy `disabled_katas: [ZC1104]` lists keep parsing.

ZC1103 (direct `PATH=$PATH:…` assignment, no `export`) is a **distinct** case and still fires.

## Verification

- Confirmed ZC1104 and ZC1188 fire on an identical input set; ZC1103 covers the non-export form separately.
- Removes 33 redundant findings from the violation baseline; ZC1188 still fires on every one of those lines.
- KATAS.md regenerated (ZC1104 -> "retired duplicate"); the ZC1104 test asserts no finding.
- `go test ./...` passes, `golangci-lint` 0 issues, gocyclo clean, sweeps clean.

### DIFF
--- a/.github/violation-baseline.txt
+++ b/.github/violation-baseline.txt
@@ -54,17 +54,12 @@ antidote/tests/bin/test_fzf_opts_manual.zsh	ZC1017	1
 antidote/tests/bin/test_fzf_opts_manual.zsh	ZC1075	2
 antidote/tests/__init__.zsh	ZC1031	1
 antidote/tests/testdata/real/.zsh_plugins.zsh	ZC1001	1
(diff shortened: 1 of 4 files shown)


redteamx (approved):
Approved on review. ZC1104≡ZC1188 confirmed; retired-duplicate pattern (ZC1018/19); ZC1103 distinct; 33 redundant findings removed.

redteamx (approved):
Approved on review.

PR #1425, opened 2026-06-25
fix: absorb glued quote tail in assignment RHS

afadesigns:
What: the assignment right-hand-side absorber now consumes a quoted segment or bare word glued with no separating space, so `expansion='b'cd` parses as one concatenated word (`bcd`) instead of orphaning `cd` into a second statement.

Why: in zsh, adjacent words with no space concatenate. The parser stopped at the closing quote, so the trailing `cd` parsed as a real command and drew a spurious ZC1044 cd-failure warning on a plain string assignment. The zsh-abbr quote-handling tests exercise every glue position.

Guard: the leading `HasPrecedingSpace` check preserves the env-prefix form `x='b' cd` — a space still marks the following word as a real command (two statements).

Tests: new `TestParseAssignmentGluedQuoteTail` pins statement counts for the four glue forms and the two spaced forms. Violation baseline drops the two `zsh-abbr/tests/_add.ztr.zsh` ZC1044 false positives; no other corpus drift across 402 files.

Severity: Warning (false positive removed).


redteamx (approved):
Approved on review.

PR #1426, opened 2026-06-25
fix: lint inside logical-chain compound bodies

afadesigns:
What: a compound statement on the right of a `&&` / `||` chain (`cmd && { ... }`, `cmd && if ...; fi`, `cmd && for ...; done`, `cmd && case ... esac`) is now walked, so katas see the code inside the block.

Why: `keywordStmtToExpression` wrapped the compound in a stub identifier, so the AST walker never descended into the body. Dangerous code inside the block, such as an unquoted `rm -rf $x`, went unlinted. These statement types already implement `Expression`, so the fix returns the real node, exactly as the `while` head already does.

Latent false positives surfaced by the wider coverage, fixed here:
- ZC1071 fired on a single-element rebuild (`arr=(${arr[@]:#p})`, a filter, not an append). Now requires the whole-array reference plus at least one appended element.
- ZC1075 fired on a `$var` glued to a literal command-name prefix (`_clear$fsuf`), which cannot elide the word.
- ZC1075 fired on a `(q)`-quoted expansion, which renders an empty value as a quoted empty string and never elides.
- ZC1149 fired on an error message already routed to stderr (`print -u2`, `>&2`).

Net corpus effect: 28 real findings inside compound bodies now visible; 21 false positives removed. Parser-error sweep stays 402/0; fix-corpus sweep clean; full suite, gocyclo, and golangci-lint green.

Tests: `TestParseLogicalChainCompoundBodyWalked` (parser), `TestLogicalChainCompoundBodyLinted` plus per-kata FP regressions in `fp_round10_test.go`.

Severity: mixed (false negatives closed, false positives removed).


redteamx (approved):
Approved on review.

PR #1427, opened 2026-06-25
fix: parse command-position assignments correctly

afadesigns:
What: a `name=value` after `&&` / `||` or in a pipeline is now parsed as an assignment, not a command.

Why: the single-command parser captured it as a command whose name was the variable, with `=value` as an argument. A variable named for a builtin was mis-flagged as that builtin -- `[[ cond ]] && test=$(x)` drew the test-command katas, `timeout=-t$2` drew the timeout kata, `seq=$mac` drew the seq katas. Assignment katas also missed these, so a real `fpath=($fpath $dir)` append after `&&` went unreported.

How: `reshapeCommandAssignment` converts the fully-parsed command node into an `=` InfixExpression when it is a lone glued `=value` argument. It runs after the command is parsed, so it never changes token consumption -- the caller's cursor is untouched, which is what makes it safe where an early re-parse desynced the dispatch loop. The env-prefixed form (`FOO=bar cmd`) and the spaced form (`cd =$x`) stay commands.

Corpus effect: 10 false positives removed across 8 lines (variables named test/seq/timeout/GEOMETRY_PROMPT assigned in a logical chain); one real array append now reported. Parser-error sweep stays 402/0; fix-corpus and full suite green.

Tests: `TestParseCommandPositionAssignment` (parser), `TestCommandPositionAssignmentNotMisflagged` and `TestCommandPositionAssignmentKeepsRealCommands` (katas).

Severity: false positives removed, one false negative closed.

### DIFF
--- a/.github/violation-baseline.txt
+++ b/.github/violation-baseline.txt
@@ -257,9 +257,6 @@ fzf-tab/test/runtests.zsh	ZC1073	2
 fzf-tab/test/runtests.zsh	ZC1075	1
 fzf-tab/test/ztst.zsh	ZC1001	3
 fzf-tab/test/ztst.zsh	ZC1004	1
-fzf-tab/test/ztst.zsh	ZC1006	1
-fzf-tab/test/ztst.zsh	ZC1020	1
-fzf-tab/test/ztst.zsh	ZC1036	1
 fzf-tab/test/ztst.zsh	ZC1037	1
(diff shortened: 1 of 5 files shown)


redteamx (dismissed):
Approved on review.

redteamx (approved):
Approved on review.

PR #1438, opened 2026-08-02
fix: [FEAT] add `latest` tag

### LINKED ISSUE #1437: [FEAT] add `latest` tag
### Request Type

New Kata (New Check)

### Description

Adding a `latest` tag would allow CI to run the latest release automatically.

```zsh

```

```zsh

```

### Rationale

_No response_

### Additional Context

_No response_

webbrain-one:
Closes #1437

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -91,6 +91,7 @@ dockers:
     goarch: amd64
     image_templates:
       - "ghcr.io/afadesigns/zshellcheck:{{ .Version }}-amd64"
+      - "ghcr.io/afadesigns/zshellcheck:latest"
       - "ghcr.io/afadesigns/zshellcheck:latest-amd64"
     build_flag_templates:
       - "--platform=linux/amd64"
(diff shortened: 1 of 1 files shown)


afadesigns:
Thanks for taking the time to put this together. Closing because the `latest` tag already ships: `.goreleaser.yml` declares a `docker_manifests` entry named `ghcr.io/afadesigns/zshellcheck:latest` built from `latest-amd64` and `latest-arm64`, and the registry confirms `latest` currently resolves to the same digest as `1.7.1`.

On the diff itself: adding `latest` to both single-arch `image_templates` would make the amd64 and arm64 builds overwrite each other's `latest`, leaving a single-arch image for whichever pushed last — the multi-arch manifest entry is the right place for it, and it is already there. The third hunk also duplicates the `latest-arm64` line inside the manifest list.

Appreciated regardless — reports like #1437 are useful even when the answer turns out to be already covered.

PR #1440, opened 2026-08-21
ci: correct stale checkout pin comment

afadesigns:
The composite `action.yml` pins `actions/checkout` at the v7.0.1 commit (`3d3c42e5`), but the version comment still read `v4.1.7` — left behind by successive dependabot SHA bumps that update the pin without touching the comment. Verified via `git ls-remote --tags`: the SHA resolves to `v7.0.1`.

Comment-only change; the executed SHA is unchanged. Every other pin in the tree (30 checked, workflows plus the composite action) matches its comment.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -25,7 +25,7 @@ runs:
   using: 'composite'
   steps:
     - name: Checkout repository under test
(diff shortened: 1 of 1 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
Approved on review.

PR #1441, opened 2026-08-21
chore: bump Go toolchain to 1.27.0

afadesigns:
Moves the `go.mod` directive and the `release.yml` `setup-go` pin from `1.25.9` to `1.27.0`, the latest stable Go release. The other workflows use `go-version: stable` and already resolve the same toolchain.

Verified locally under 1.27.0: full test suite green, `golangci-lint` 2.13.1 (built with go1.27.0) reports 0 issues, `gocyclo` clean, parser-corpus sweep 402/0, violation-corpus sweep zero drift, fix-corpus sweep 0 corruptions / 0 non-idempotent / 0 panics.

Zero dependencies in `go.mod`, so this is a pure toolchain move; the release build stays reproducible against the pinned version.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
-          go-version: '1.25.9'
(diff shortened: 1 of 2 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
Approved on review.

PR #1442, opened 2026-08-21
fix: emit valid URIs in SARIF artifact locations

### LINKED ISSUE #1435: [BUG] URIs in SARIF output not actually URIs
### Describe the bug

The URI fields in SARIF output seem to be regular file paths, not actually in URI format. This is important for tools that expect a proper URI.

### Reproduction Code

```zsh
echo hi
```

### Actual Output

```text
{
  "$schema": "https://json.schemastore.org/sarif-2.1.0.json",
  "version": "2.1.0",
  "runs": [
    {
      "tool": {
        "driver": {
          "name": "zshellcheck",
          "informationUri": "https://github.com/afadesigns/zshellcheck",
          "version": "1.7.1",
          "rules": [
            {
              "id": "ZC1037",
              "name": "Use 'print -r --' for variable expansion",
              "shortDescription": {
                "text": "Use 'print -r --' for variable expansion"
              },
              "fullDescription": {
                "text": "Using 'echo' to print strings containing variables can lead to unexpected behavior if the variable contains special characters or flags. A safer, more reliable alternative is 'print -r --'."
              },
              "helpUri": "https://github.com/afadesigns/zshellcheck/blob/main/KATAS.md",
              "defaultConfiguration": {
                "level": "note"
              }
            }
          ]
        }
      },
      "results": [
        {
          "ruleId": "ZC1037",
          "ruleIndex": 0,
          "level": "note",
          "message": {
            "text": "Use `print -r --` instead of `echo`. `echo` behaviour varies with options and escape handling; `print -r --` is the reliable Zsh builtin."
          },
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "/path/to/test.zsh"
                },
                "region": {
                  "startLine": 1,
                  "startColumn": 1
                }
              }
            }
          ]
        }
      ]
    }
  ]
}
```

### Expected Output

The `"uri"` field should be something like `file:///path/to/test.zsh`

### ZShellCheck Version

v1.7.1

### Operating System

Linux

### Additional Context

_No response_

afadesigns:
Closes #1435.

What: `artifactLocation.uri` in SARIF output carried the raw filesystem path. An absolute path (`/path/to/test.zsh`) is not a valid URI, and strict consumers rejected it.

How: absolute paths become `file://` URIs with percent-encoded segments (`file:///path/to/test.zsh`). Relative paths stay relative URI references with forward slashes — the form GitHub code scanning requires to map results onto repository files — and a first segment containing a colon gains a `./` prefix so it cannot be misread as a URI scheme. Windows drive paths form `file:///C:/dir/file.zsh`.

Tests: `TestSarifFileURI` covers relative, absolute, space-encoding, and colon-segment cases; `TestReportSARIF_AbsolutePathURI` pins the issue repro end to end. Existing relative-path expectations are unchanged.

Severity: Warning (spec conformance).

### DIFF
--- a/pkg/reporter/aggregate.go
+++ b/pkg/reporter/aggregate.go
@@ -215,9 +215,11 @@ func buildSarifRule(v katas.Violation, meta func(string) RuleMeta) sarifRule {
 // which broke consumers that parse the field strictly.
 func sarifFileURI(path string) string {
 	s := filepath.ToSlash(path)
-	if filepath.IsAbs(path) {
-		// A Windows drive path (C:/dir) needs a leading slash to form the
-		// file:///C:/dir shape; POSIX absolute paths already have one.
+	// A slash-rooted path is treated as absolute on every platform, so the
+	// same input renders identically on Windows, where filepath.IsAbs
+	// rejects a rooted path without a drive letter. A Windows drive path
+	// (C:/dir) gains the leading slash the file:///C:/dir shape needs.
+	if filepath.IsAbs(path) || strings.HasPrefix(s, "/") {
 		if !strings.HasPrefix(s, "/") {
 			s = "/" + s
 		}
--- a/pkg/reporter/aggregate.go
+++ b/pkg/reporter/aggregate.go
@@ -5,6 +5,9 @@ package reporter
 import (
 	"encoding/json"
 	"io"
+	"net/url"
+	"path/filepath"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


redteamx (approved):
Approved on review.

redteamx (dismissed):
Approved on review.

PR #1443, opened 2026-08-21
ci: bump govulncheck to v1.7.0

afadesigns:
The vulnerability scan pins `govulncheck@v1.1.4`, whose vendored `x/tools` v0.29.0 SSA builder panics with `unexpected expr: *ast.KeyValueExpr` under Go 1.27.0 — which `setup-go`'s `go-version: stable` now resolves. The scan job crashes with exit 2 on every run, failing the required `security` check (seen on #1440 and #1441).

Bumps the pin to `v1.7.0`, the latest `x/vuln` release (verified via `git ls-remote --tags`), which carries a current `x/tools`. This PR's own scan run exercises the new pin under the same resolved toolchain.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -242,7 +242,7 @@ jobs:
           go-version: 'stable'
           cache: true
       - name: Install govulncheck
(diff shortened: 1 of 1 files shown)


redteamx (approved):
Approved on review.

PR #1446, opened 2026-08-25
fix: align ZC1044 scope with where a failed cd can do harm

afadesigns:
## What

ZC1044 ("Use \`cd ... || return\`") walks the tree itself, because the guard depends on whether the exit status is already consumed — an `if` condition, the left of `||`, a negation — and the shared walker cannot hand that state to one child but not its sibling. That private walker had drifted from the node set the parser produces.

**Parity bug.** `function name { … }` builds a `FunctionLiteral`; `name() { … }` builds a `FunctionDefinition`. Only the second was visited:

```
function cdf() { cd /etc; ls }   -> silent
cdf()          { cd /etc; ls }   -> ZC1044
```

Identical Zsh semantics, opposite verdicts. Case clause bodies and select bodies were also never visited, though both run in the current shell with the same blast radius as the top level.

**Exemption.** A `cd` that ends a function no longer reports, directly or through a trailing `if`. Nothing runs after it, a failed `cd` leaves the directory unchanged, and the failure surfaces as the function's exit status for the caller to test. Verified against zsh 5.9: `f() { cd /nonexistent }` leaves the caller's directory untouched and returns 1.

**Fork boundaries stay out of scope.** A subshell, `$( … )`, `<( … )`, or `coproc` cannot move the caller's directory. Measured on 933 real Zsh files, extending into subshells produced 13 warnings and zero genuine bugs — 10 of them on `( cd -- "$ROOT" && cmd )`, already guarded.

## Corpus effect

26 findings drop, all of the newly exempt last-command form, e.g. `autojump/bin/autojump.zsh` (`cd "${output}"` inside `if [[ -d "${output}" ]]` ending the function) and 23 spaceship test files (`cd $SHUNIT_TMPDIR` ending `setUp()`). Nothing is lost from a context that can still do harm.

Parser sweep 402 files / 0 errors; fix-corpus sweep 0 corruptions, 0 non-idempotent, 0 panics; full suite, `golangci-lint`, and `gocyclo` all clean (`walkZC1044Compound` sits at 11 of the 15 ceiling).

## Tests

Six cases added to `TestZC1044`: firing for a function-keyword body, a case clause body, and a select body; silence for a `cd` ending a function, one ending it through a trailing `if`, and one inside a subshell. Existing assertions — including `cd /tmp && echo ok` and the loop-body cases — are unchanged.

Severity: Warning (false negatives closed, false positives removed).

### DIFF
--- a/.github/violation-baseline.txt
+++ b/.github/violation-baseline.txt
@@ -83,7 +83,6 @@ atuin/crates/atuin/src/shell/atuin.zsh	ZC1353	1
 atuin/crates/atuin/src/shell/atuin.zsh	ZC1509	1
 autojump/bin/autojump.zsh	ZC1010	1
 autojump/bin/autojump.zsh	ZC1037	9
-autojump/bin/autojump.zsh	ZC1044	1
 autojump/bin/autojump.zsh	ZC1045	2
 autojump/bin/autojump.zsh	ZC1075	8
 autojump/bin/autojump.zsh	ZC1120	2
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
Approved on review.

PR #1447, opened 2026-08-25
fix: stop ZC1053 reporting an already-silenced grep

afadesigns:
## What

ZC1053 asks you to silence a `grep` used as a condition and names two ways to do it — `grep -q`, or a redirect to `/dev/null`. It recognised only the first, so it reported the very form it recommends:

```
zsh/Src/mkmakemod.sh:110  if grep '^#define DYNAMIC ' config.h >/dev/null; then
                          ^ ZC1053: "Use \`grep -q\` or redirect to \`/dev/null\`"
```

Across the pinned corpora, **every** ZC1053 finding was a grep that was already redirected — 7 findings, 7 false positives, 0 true positives.

## Why it happened

The parser folds a redirection into the argument list rather than building a redirection node, so the operator arrives glued to its target (`>/dev/null`, one argument) or as two adjacent words (`> /dev/null`). The existing redirection-node branch was therefore never reached in condition position, and its `/dev/null` comparison could not read an unquoted target anyway, since that parses as a bare word.

## What still reports

Verified against zsh 5.9 — these keep printing to stdout, so they remain findings:

- `grep p f 2>/dev/null` — stderr only, stdout still prints
- `grep p f >/dev/null.txt` — an ordinary file, not the null device
- `grep '>/dev/null' f` — the redirect text as a quoted search pattern
- `grep -c p f` — prints a count

## Quiet-flag scan

It accepted any short option containing the letter `q`, which silently swallowed real findings: `grep -fq file` is `-f q` (read patterns from a file named `q`), and a quoted `"-quiet"` is the pattern. Clusters are now walked left to right and stop at the first option that consumes a value (`-e -f -m -A -B -C -d -D`), scanning stops after `--`, and quoted words are treated as operands.

## Gates

Corpus drift: 7 removed, 0 added. Parser sweep 402 files / 0 errors; fix-corpus sweep 0 corruptions, 0 non-idempotent, 0 panics; suite, `golangci-lint`, and `gocyclo` clean — `checkCommandZC1053` was at the 15 ceiling and is now well under it, since the scans moved into helpers.

Tests: 10 cases added to `TestZC1053` covering each silencing form and each look-alike that must still report; the `isDevNull` helper test now pins bare-word targets.

Known, unchanged, and out of scope: `if ( grep p f )` is silent (a subshell condition is not walked; 4 hits across 2533 files), and `grep -eq` never reaches the kata because the lexer tokenises `-eq` as the numeric-comparison operator.

Severity: Style (false positives removed, false negatives closed).

### DIFF
--- a/.github/violation-baseline.txt
+++ b/.github/violation-baseline.txt
@@ -418,7 +418,6 @@ prezto/modules/git/alias.zsh	ZC1049	212
 prezto/modules/gnu-utility/init.zsh	ZC1046	1
 prezto/modules/gnu-utility/init.zsh	ZC1098	1
 prezto/modules/gpg/init.zsh	ZC1046	1
-prezto/modules/gpg/init.zsh	ZC1053	1
 prezto/modules/gpg/init.zsh	ZC1086	1
 prezto/modules/gpg/init.zsh	ZC1098	1
 prezto/modules/gpg/init.zsh	ZC1125	1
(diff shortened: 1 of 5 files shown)


redteamx (approved):
Approved on review.

redteamx (dismissed):
Approved on review.

redteamx (approved):
Approved on review.

PR #1448, opened 2026-08-26
fix: report ZC1273 without rewriting the redirect

afadesigns:
## The corruption

`zshellcheck -fix -unsafe-fixes` produced unparseable Zsh. A spaced redirect occupies two argument slots; the fix deleted only the `/dev/null` one:

```
grep p f > /dev/null    ->  grep -q p f >      zsh -n: parse error
grep p f >> /dev/null   ->  grep -q p f >>     zsh -n: parse error
```

On real corpus code, `prezto/modules/gpg/init.zsh:33` became `if grep -q '^enable-ssh-support' -- "$_gpg_agent_conf" &>; then`.

## Why the rewrite is dropped rather than repaired

Even performed correctly, `grep -q` is not a drop-in for a discarded redirect:

- **It inverts conditions under `pipefail`.** `-q` exits on the first match, SIGPIPE-ing the producer: `seq 1 8000000 | grep -q '^1$'` exits **141**, while `seq 1 8000000 | grep '^1$' >/dev/null` exits **0**.
- **It leaks stderr.** `grep -q p /nonexistent` prints "No such file or directory"; `grep p /nonexistent &> /dev/null` printed nothing.
- **Exit codes differ** with a missing file among several: `2` versus `0`.

So the kata now reports and explains, and the description states why the forms are not interchangeable. The advice itself stands — `grep -q` is ~31x faster on an early match — it just cannot be applied mechanically.

## Detection was wrong in both directions

It matched any `/dev/null` argument, so it fired on `grep pattern file /dev/null` — where `/dev/null` is a **second search file**, the idiom that forces grep to print file-name prefixes — and the fix deleted it, silently changing output from `file:match` to `match`. It also missed the attached spelling `>/dev/null`, which is the more common one in the corpus.

Detection is now operator-aware: a plain stdout redirect (`>` or `1>`) to `/dev/null`, attached or spaced, and never a form that also discards stderr (`&>`, `2>`, `2>&1`, `>>`).

## Why it shipped

Both tests only covered `grep PAT file /dev/null` — the argument form, no redirection operator anywhere. The fixture never exercised the case the kata's own title describes. Tests now cover each redirect spelling and each form that must stay silent, and the fix-integration test asserts the source comes back **unchanged**.

## Gates

Corpus drift: 2 added (attached-form redirects in `zinit-install.zsh` and `zsh-256color`, both genuine), 3 removed (all `&> /dev/null`, where `-q` would leak stderr). Parser sweep 402/0; fix-corpus sweep clean; suite, lint, gocyclo green. KATAS.md regenerated (auto-fix count 132 to 131).

Severity: Warning — a corrupting autofix removed, false positives removed, false negatives closed.

### DIFF
--- a/.github/violation-baseline.txt
+++ b/.github/violation-baseline.txt
@@ -421,7 +421,6 @@ prezto/modules/gpg/init.zsh	ZC1046	1
 prezto/modules/gpg/init.zsh	ZC1086	1
 prezto/modules/gpg/init.zsh	ZC1098	1
 prezto/modules/gpg/init.zsh	ZC1125	1
-prezto/modules/gpg/init.zsh	ZC1273	1
 prezto/modules/gpg/init.zsh	ZC1502	1
 prezto/modules/helper/init.zsh	ZC1076	1
 prezto/modules/helper/init.zsh	ZC1086	9
@@ -603,7 +602,6 @@ spaceship/sections/haxe.zsh	ZC1091	1
 spaceship/sections/hg_branch.zsh	ZC1045	1
 spaceship/sections/hg_status.zsh	ZC1037	4
 spaceship/sections/hg_status.zsh	ZC1045	1
-spaceship/sections/hg_status.zsh	ZC1273	4
 spaceship/sections/hg.zsh	ZC1010	1
 spaceship/sections/hg.zsh	ZC1045	2
 spaceship/sections/hg.zsh	ZC1097	1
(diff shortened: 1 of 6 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
Approved on review.

redteamx (approved):
Approved on review.

PR #1449, opened 2026-08-26
fix: apply a violation's edits as one unit

afadesigns:
## The bug

A rewrite can span several edits. Turning `[[ a -eq b ]]` into `(( a == b ))` replaces the opening bracket, the operator, and the closing bracket — the three only make sense together.

The safety net that rejects a pass introducing parse errors retried the edits **one at a time**, keeping whichever parsed alone. For a condition that cannot move into arithmetic — a pattern match beside a numeric test — the bracket edits failed individually while the operator edit succeeded:

```
[[ "$w" = -* && $c -eq 0 ]]   ->   [[ "$w" = -* && $c == 0 ]]
```

That parses cleanly, so **no gate objected**. But `-eq` compares numbers and `==` matches a glob pattern:

```
zsh -c '[[ "01" -eq 1 ]]'   # true
zsh -c '[[ "01" == 1 ]]'    # false
```

The safety net was manufacturing the corruption it exists to prevent, in the one form no parser check can catch.

## The fix

Each edit now carries the identity of the violation it repairs. The retry keeps or discards a **group** whole, and a group whose full application still breaks the parse is skipped rather than partially applied.

The two filters in the edit collector had the same confusion — they indexed the edit slice by *violation* position, so any run with `-severity` or a `# noka` directive tore multi-edit rewrites apart the same way. They now select by group.

## Effect

Applying every unsafe fix across the pinned corpora previously left **11 files that Zsh itself rejects**. That count is now **0** (218 files re-checked with `zsh -n`), with **no change to reported findings** — violation drift is 0 added, 0 removed.

Two of the known corruption classes are resolved by atomicity alone:

| Input | Before | After |
|---|---|---|
| `[[ "$w" = -* && $c -eq 0 ]]` | `[[ … == 0 ]]` (silent semantic change) | unchanged |
| `[ -f file ] \| cat` | `[[ -f file ]]] \| cat` (syntax error) | unchanged |

Conversions that were always correct still happen: `[[ $a -eq 0 ]]` → `(( a == 0 ))`, `[ 1 -eq 1 ]` → `(( 1 == 1 ))`, `test -f file` → `[[ -f file ]]`.

Where a group cannot be applied safely the source is now left untouched. That is a deliberate trade: no fix beats a broken fix.

## Not covered here

`test -f f 2>/dev/null` → `[[ -f f 2>/dev/null ]]` is still wrong — the fix wraps the redirect inside the brackets as a complete group, so it is a detection/rewrite bug rather than a partial-application one. It gets its own change.

## Gates

Suite, `golangci-lint`, and `gocyclo` green; parser sweep 402/0; fix-corpus sweep clean; violation sweep unchanged. `TestApplySafeEdits_Bailouts` gains a case pinning that a group containing one breaking edit is dropped whole, and its existing fixtures now state explicitly which edits are independent.

Severity: Error — silent semantic corruption of user code removed.

### DIFF
--- a/cmd/zshellcheck/main.go
+++ b/cmd/zshellcheck/main.go
@@ -560,26 +560,51 @@ func parseErrorCount(src string) int {
 // stays valid against the growing source. A failed apply or a fix that
 // would break the parse is simply skipped, so a broken file is never
 // written.
+// applySafeEdits retries a rejected pass one violation at a time, keeping
+// only the groups that parse. A group is all-or-nothing: the edits that
+// rewrite `[[ a -eq b ]]` into `(( a == b ))` are meaningless apart, and
+// applying just the operator would leave `[[ a == b ]]`, which parses but
+// compares as a glob pattern instead of a number.
 func applySafeEdits(base string, edits []katas.FixEdit) (string, int) {
 	baseErrs := parseErrorCount(base)
-	ordered := append([]katas.FixEdit(nil), edits...)
-	sort.SliceStable(ordered, func(i, j int) bool {
-		if ordered[i].Line != ordered[j].Line {
-			return ordered[i].Line > ordered[j].Line
-		}
-		return ordered[i].Column > ordered[j].Column
-	})
+	groups := groupEdits(edits)
 	acc := base
 	applied := 0
-	for _, e := range ordered {
-		trial, err := fix.Apply(acc, []katas.FixEdit{e})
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review.

redteamx (dismissed):
Approved on review.

redteamx (approved):
Approved on review.

PR #1450, opened 2026-08-26
fix: keep a redirection outside the rewritten brackets

afadesigns:
## What

Rewriting `test EXPR` into `[[ EXPR ]]` closed the bracket after the **last argument**. Since the parser folds a redirection into the argument list rather than building a redirection node, the redirect ended up inside:

```
test -f f 2>/dev/null   ->   [[ -f f 2>/dev/null ]]
```

Zsh rejects that — a redirection may follow a conditional expression but not sit inside one:

```
zsh -n -c '[[ -f x ]] 2>/dev/null'    # valid
zsh -n -c '[[ -f x 2>/dev/null ]]'    # parse error near '>'
```

This was the last of the four autofix-corruption classes; unlike the others it was not a partial-application problem, so the atomic-group change in #1449 did not address it. The fix emitted a complete, wrong rewrite.

## The fix

The closing bracket now goes after the last argument that belongs to the expression, leaving any trailing redirection untouched:

```
test -f f 2>/dev/null    ->  [[ -f f ]] 2>/dev/null
test -f f 2> /dev/null   ->  [[ -f f ]] 2> /dev/null
test -f f >/dev/null     ->  [[ -f f ]] >/dev/null
```

Plain rewrites are unchanged (`test -f file` → `[[ -f file ]]`, `test "$a" = "$b"` → `[[ "$a" = "$b" ]]`), and the `[` form already behaved because its own `]` marks where the expression ends.

The fix is shared by ZC1006, ZC1020, ZC1036, and ZC1293, so all four are corrected together.

## Gates

Suite, `golangci-lint`, `gocyclo` green; parser sweep 402/0; fix-corpus sweep clean; violation drift 0 added, 0 removed. New `TestFixIntegration_ZC1293_RedirectStaysOutsideBrackets` covers the attached, spaced, and stdout spellings.

Severity: Error — a corrupting autofix removed.

### DIFF
--- a/pkg/fix/integration_test.go
+++ b/pkg/fix/integration_test.go
@@ -1571,6 +1571,9 @@ func TestFixIntegration_ZC1293_RedirectStaysOutsideBrackets(t *testing.T) {
 		"test -f f 2>/dev/null\n":  "[[ -f f ]] 2>/dev/null\n",
 		"test -f f 2> /dev/null\n": "[[ -f f ]] 2> /dev/null\n",
 		"test -f f >/dev/null\n":   "[[ -f f ]] >/dev/null\n",
+		// Nothing but a redirection follows the command name, so there is
+		// no expression to wrap and the source is left alone.
+		"test 2>/dev/null\n": "test 2>/dev/null\n",
 	}
 	for src, want := range cases {
 		if got := runFix(t, src); got != want {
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review.

redteamx (dismissed):
Approved on review.

redteamx (approved):
Approved on review.

PR #1452, opened 2026-08-26
ci: judge rewritten files with zsh, not only our parser

afadesigns:
## The hole

The autofix gate measured corruption as an increase in **this linter's own** parser-error count. Our parser is more permissive than zsh — it accepts a dangling `>`, a stray `]]`, an unterminated block — so a fix producing any of those scored zero errors and passed. The corrupted output was also idempotent, so the convergence invariant stayed quiet as well.

That is how four separate corrupting autofixes reached the corpus while the gate reported `corruptions=0` on every run.

## The check

A file zsh accepted before the rewrite must still be accepted after it.

Two traps it accounts for:

- **`zsh -n` exits 0 while still printing some diagnostics.** `zinit/share/rpm2cpio.zsh` prints "redirection with no command" and returns 0. The check reads stderr, not the exit status.
- **`$(<file)` makes `zsh -n` open the file**, reporting a missing input. That is not a syntax error, so the message is ignored — otherwise the legitimate `$(cat f)` → `$(<f)` rewrite would look like damage.

Where zsh is absent the check is skipped with a warning rather than failing, so the sweep still runs off CI and on other platforms. The job that runs it is ubuntu-only, where zsh is present; the summary line now ends with `zsh=zsh` or `zsh=none` so a silent degradation is visible.

## Proof it works

Same script, two binaries:

```
current main:                 files=402 corruptions=0 ... zsh=zsh
binary from before the fixes: files=402 corruptions=9 ... zsh=zsh
```

The nine are the files this work repaired — fzf/completion.zsh, prezto/gpg/init.zsh, spaceship ×3, zgen, zinit-autoload, zsh-utils/editor, zsh-vimode-visual — each reported with its line and zsh's own reason, for example:

```
+ zgen/zgen.zsh [unsafe]: zsh rejects the rewrite at line 316: parse error near `}'
```

Before this change that same run reported `corruptions=0`.

Severity: Error — the gate that was supposed to catch destructive fixes could not see them.

### DIFF
--- a/scripts/fix-corpus-sweep.sh
+++ b/scripts/fix-corpus-sweep.sh
@@ -5,8 +5,13 @@
 # never corrupts the source and always converges:
 #
 #   1. No corruption — the rewritten file must not raise the parser-error
-#      count above the original file's own count. A fix that turns
-#      parseable source into unparseable source is a destructive bug.
+#      count above the original file's own count, and, where a real zsh is
+#      available, must not turn a file zsh accepts into one it rejects.
+#      A fix that turns parseable source into unparseable source is a
+#      destructive bug. The second check matters because this linter's
+#      parser is more permissive than zsh: it accepts a dangling `>` or a
+#      stray `]]`, so measuring corruption with the linter alone is blind
+#      exactly where a fix has gone wrong.
 #   2. Idempotent — a second `-fix` pass over the already-fixed file must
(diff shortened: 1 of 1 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
Approved on review.

redteamx (approved):
Approved on review.

PR #1454, opened 2026-08-26
fix: reject a redirection with no target

afadesigns:
## What

Zsh rejects `cmd >` outright. This parser accepted it — and that silence had consequences beyond a missing diagnostic. The auto-fix safety gate measures corruption with this parser, so a rewrite that deleted a redirect target while leaving its operator behind scored **zero errors** and passed every run. That is how the ZC1273 corruption reached the corpus unnoticed.

## Where the check goes

A redirection in command position is folded into the argument list rather than parsed as a redirection node, so the check belongs where a command finishes gathering arguments: a final argument that is nothing but a redirection operator — with or without a file-descriptor prefix — now records a parser error. The expression-path redirection parser gained the same guard.

## The one subtlety

`>|` and `>>|` override `NO_CLOBBER`, and the lexer splits them into an operator plus a pipe token. A pipe **glued** to the operator therefore means the target is still to come, while a spaced `> | cat` really is dangling — which matches zsh:

```
zsh -n -c ': >>| /dev/null'      # valid
zsh -n -c 'echo x > | cat'       # parse error near '|'
```

`zaw/sources/bookmark.zsh` uses `: >>| "${BOOKMARKFILE}"`; the corpus sweep caught my first attempt flagging it, which is what surfaced this distinction.

## Agreement with zsh

Checked on eighteen forms — dangling `>`, `>>`, `1>`, `2>`, `> | cat`; and valid `> file`, `>/dev/null`, `>> log`, `2>&1`, `exec 3>&1`, `>|`, `>>|`, heredoc, here-string, process substitution, a quoted `">"`, and `(( a > b ))`. The linter and zsh agree on every one.

## Gates

Parser sweep 402 files / **0** errors; violation drift 0 added, 0 removed; fix-corpus sweep clean with the zsh oracle; suite, `golangci-lint`, `gocyclo` green.

Severity: Warning — a syntax error the linter used to accept.

### DIFF
--- a/pkg/parser/parser_corpus_fixes_test.go
+++ b/pkg/parser/parser_corpus_fixes_test.go
@@ -681,6 +681,13 @@ func TestParseDanglingRedirection(t *testing.T) {
 		"grep p f 1>\n",
 		"grep p f 2>\n",
 		"echo x > | cat\n",
+		// The compound operators reach the expression-path redirection
+		// parser rather than the command one.
+		"x=1 >>\n",
+		"( cmd >> )\n",
+		"a && b >>\n",
+		"cmd | b <&\n",
+		"f() { cmd >& }\n",
 	}
 	for _, src := range dangling {
 		p := New(lexer.New(src))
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review.

redteamx (dismissed):
Approved on review.

redteamx (approved):
Approved on review.

PR #1455, opened 2026-08-26
fix: recognise git clone --depth=1 and friends

afadesigns:
## What

Every `git clone` option that truncates history takes a value, and git accepts both spellings — `--depth 1` and `--depth=1`. ZC1231 compared each argument against a bare `--depth`, so the inline form read as a full clone.

That spelling is not the rare one. Across the pinned corpora, clone lines that set a depth use `--depth=N` over `--depth N` by **2.6 to 1**; long options in general carry an inline value in roughly a third of real usages. The kata was blind to the majority form.

## Why it matters more than a stray warning

The auto-fix acted on the bad reading. On a clone that limits history by date it inserted `--depth 1` alongside `--shallow-since`, and git rejects that pair outright:

```text
fatal: git upload-pack: deepen and deepen-since (or deepen-not) cannot be used together
```

Exit 128 — a working shallow clone rewritten into a command that cannot run. The second failure mode is quieter: git keeps the **last** value of a repeated option, so inserting `--depth 1` ahead of an existing `--depth=50` accomplishes nothing while leaving a duplicate flag behind.

## The change

- Arguments split on an inline `=` before the name is compared, through a new `LongFlagName` helper in `fixutil.go`. It is the reusable primitive for this class.
- `--shallow-exclude` joins the set it always belonged to.
- The fix refuses to add a second history-limiting option, so the two failures above cannot recur even if the check is bypassed.
- `--single-branch` **leaves** the set. It narrows a clone to one branch but still downloads that branch entire, which is precisely the cost the kata warns about. Verified against git 2.55.0: `git clone --single-branch` of a six-commit repository fetches all six.

## Behaviour

| clone | before | after |
|---|---|---|
| `--depth 1` | clean | clean |
| `--depth=1`, `--depth=50`, `--depth="1"` | **flagged, fix breaks or no-ops** | clean |
| `--shallow-since=<date>` | **flagged, fix exits 128** | clean |
| `--shallow-since <date>` | clean | clean |
| `--shallow-exclude=<ref>` | **flagged** | clean |
| `--single-branch` | silent | flagged, fix valid |
| `--filter=blob:none` | flagged | flagged, fix valid |
| plain | flagged | flagged |

`--filter` stays a finding on purpose: a blobless partial clone still fetches the full commit history, and `--depth 1 --filter=blob:none` is a valid combination.

## Gates

Violation drift is one line, and it is the bug itself: `zgen/zgen.zsh:136` clones with `--depth=1 --recursive` and was being told to make it shallow. Nothing appeared. Parser sweep 402/0; fix-corpus sweep clean under the zsh oracle; suite, `golangci-lint` (0 issues), `gocyclo` green.

Severity stays Style.

### DIFF
--- a/.github/violation-baseline.txt
+++ b/.github/violation-baseline.txt
@@ -1125,7 +1125,6 @@ zgen/zgen.zsh	ZC1075	3
 zgen/zgen.zsh	ZC1097	7
 zgen/zgen.zsh	ZC1098	1
 zgen/zgen.zsh	ZC1099	1
-zgen/zgen.zsh	ZC1231	1
 zgen/zgen.zsh	ZC1830	1
 zhooks/zhooks.plugin.zsh	ZC1075	2
 zimfw/zimfw.zsh	ZC1040	1
--- a/pkg/fix/integration_test.go
+++ b/pkg/fix/integration_test.go
@@ -653,6 +653,23 @@ func TestFixIntegration_ZC1231_GitCloneShallow(t *testing.T) {
 	}
 }
 
+// A clone that already limits its history is left alone. Inserting a second
+// history-limiting option is never an improvement: git refuses `--depth`
+// together with `--shallow-since`, and a second `--depth` silently overrides
+// the first.
+func TestFixIntegration_ZC1231_ShallowCloneUntouched(t *testing.T) {
+	for _, src := range []string{
+		"git clone --depth=1 https://github.com/x/y\n",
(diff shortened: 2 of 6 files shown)


redteamx (approved):
Approved on review.

redteamx (dismissed):
Approved on review.

redteamx (approved):
Approved on review.

PR #1456, opened 2026-08-26
fix: read a long flag with an inline value

afadesigns:
## What

Most GNU-style tools accept `--flag=value` wherever they accept `--flag value`, and roughly a third of real long-flag usage takes the inline form. Five katas compared an argument against the bare spelling only, so the inline one slipped past.

## Both directions were wrong

Sometimes in the same command. On `xargs --max-procs=4 cmd`:

| kata | before | after |
|---|---|---|
| ZC1111 | fires — sees a plain xargs, suggests a Zsh loop | silent |
| ZC1406 | silent — misses the parallel form it exists to catch | fires |

The other three are false positives on code that already does the right thing:

| kata | input that misfired |
|---|---|
| ZC1521 | `strace --trace=openat ls` read as an unscoped strace |
| ZC1579 | `curl --retry-all-errors --max-time=5` read as having no timeout |
| ZC1494 | `tcpdump -w f --relinquish-privileges=user` read as an unguarded capture |

## Spellings verified against the tools on this machine

```text
xargs     --max-procs=max-procs, --replace[=replace-str], --max-lines[=max-lines]
strace    --trace=syscall_set, --trace-path=path
curl      --max-time=5      (bad value gives the same diagnostic as the spaced form)
tcpdump   --relinquish-privileges=user
```

Boolean options are left alone — `xargs --null=x` exits with `option '--null' doesn't allow an argument`, so the inline form cannot occur.

## Deliberately excluded

`sudo --preserve-env=list` names the variables to keep rather than preserving the whole environment. It is a different flag from bare `-E`, not a spelling of it, and reporting it with ZC1584's wholesale wording would misdescribe the code. It needs its own treatment.

## Gates

Corpus drift zero — no pinned file writes these options inline, so this is a pure correctness gain with no observable change on the corpus. The new cases are non-vacuous: eight subtests fail against the previous code. Parser sweep 402/0; fix-corpus sweep clean under the zsh oracle; suite, `golangci-lint` (0 issues), `gocyclo` green.

Builds on the `LongFlagName` helper added in #1455.

### DIFF
--- a/pkg/katas/katatests/zc1100s_test.go
+++ b/pkg/katas/katatests/zc1100s_test.go
@@ -557,6 +557,23 @@ func TestZC1111(t *testing.T) {
 			expected: []katas.Violation{},
 		},
 		{
+			// xargs accepts the value inline for every option that takes
+			// one, and that spelling must count the same as the spaced one.
+			name:     "valid xargs with parallel inline value",
+			input:    `xargs --max-procs=4 cmd`,
+			expected: []katas.Violation{},
+		},
+		{
+			name:     "valid xargs with replace inline value",
+			input:    `xargs --replace=@ mv @ /dest`,
+			expected: []katas.Violation{},
+		},
+		{
+			name:     "valid xargs with max-lines inline value",
+			input:    `xargs --max-lines=1 cmd`,
+			expected: []katas.Violation{},
+		},
+		{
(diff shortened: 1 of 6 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
Approved on review.

PR #1457, opened 2026-08-26
fix: read a flag value written inline

afadesigns:
## What

Three katas read an option's value from the *following* argument, so the inline spelling hid it. Each of the three guards something that matters:

```zsh
tar --directory=/ -xf a.tar          # extracts into the filesystem root
curl --cacert=/dev/null https://h    # hands curl an empty trust store
smartctl --smart=off /dev/sda        # turns off the drive health source
```

All three were silent. The spaced form of each was caught.

ZC1523 carries **Error** severity, and writing `--directory=/` instead of `-C /` was enough to walk past it — while tar extracts to exactly the same place.

## The helper

`FlagValueAt` returns an option's value together with the number of arguments it spans, so a caller no longer keeps its own previous-argument state machine. ZC1523 loses one:

```go
for i := range cmd.Arguments {
    if dir, _ := FlagValueAt(cmd.Arguments, i, "-C", "--directory"); dir == "/" {
```

The span return is there for callers that need to skip a consumed value; `--directory=/` spans one argument, `--directory /` spans two.

## Spellings verified against the tools on this machine

```text
tar        -C, --directory=DIR
smartctl   -s VALUE, --smart=VALUE
curl       --cacert=path      (same parser that takes --max-time=5)
```

## Behaviour

| command | before | after |
|---|---|---|
| `tar --directory=/ -xf a.tar` | silent | ZC1523 |
| `tar -C / -xf a.tar` | ZC1523 | ZC1523 |
| `tar --directory=/opt -xf a.tar` | silent | silent |
| `curl --cacert=/dev/null …` | silent | ZC1661 |
| `curl --cacert=/etc/ca.pem …` | silent | silent |
| `smartctl --smart=off …` | silent | ZC1835 |
| `smartctl --smart=on …` | silent | silent |

## Gates

Corpus drift zero — no pinned file writes these options inline. The new cases are non-vacuous: three subtests fail against the previous code. Parser sweep 402/0; fix-corpus sweep clean under the zsh oracle; suite, `golangci-lint` (0 issues), `gocyclo` green.

Completes the sweep started in #1455 and #1456.

### DIFF
--- a/pkg/katas/fixutil.go
+++ b/pkg/katas/fixutil.go
@@ -136,3 +136,34 @@ func LongFlagName(word string) (string, bool) {
 	}
 	return word, false
 }
+
+// FlagValueAt returns the value carried by the option at args[i] when that
+// option is one of names, together with the number of arguments it spans.
+// `--directory=/` spans one argument, `--directory /` spans two. A word that
+// names no listed option, or an option whose value never arrives, yields
+// ("", 0).
+func FlagValueAt(args []ast.Expression, i int, names ...string) (string, int) {
+	if i < 0 || i >= len(args) {
+		return "", 0
+	}
+	word := args[i].String()
+	name, inline := LongFlagName(word)
+	matched := false
+	for _, want := range names {
+		if name == want {
+			matched = true
+			break
+		}
+	}
+	if !matched {
+		return "", 0
+	}
+	if inline {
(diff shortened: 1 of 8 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
Approved on review.

PR #1458, opened 2026-08-26
fix: read the last inline flag values

afadesigns:
## What

Seven more katas matched only the spaced spelling of an option, so the inline one walked past them. Five of the seven guard something specific:

```zsh
env --unset=LD_PRELOAD cmd        # clears a security-relevant variable
usermod --password=$6$hash bob    # puts the hash in ps and /proc
date --set=2025-01-01             # sets the system clock
pip install --trusted-host=h pkg  # skips TLS verification for an index
stat --format=%s file             # a shell-out Zsh does with zstat
date --date=@0                    # a shell-out Zsh strftime does
diff --unified=3 a b              # a diff Zsh can produce itself
```

All seven were silent. Every spaced equivalent was caught.

## One of them was already trying

ZC1402 meant to handle its inline form:

```go
if v == "-d" || v == "--date" ||
    (len(v) > 6 && v[:6] == "--date=") {
```

`v[:6]` is six characters — `--date` — and `"--date="` is seven. The comparison could never be true, so the branch was dead from the start. It now goes through the shared helper, which cannot drift that way.

ZC1562 and ZC1514 each also lose a previous-argument state machine, replaced by `FlagValueAt`.

## Every spelling confirmed by running the tool

```text
env --unset=PATH sh -c …            exit 0
date --date=@0 -u +%Y               1970
stat --format=%s /etc/hostname      9
diff --unified=1 f f                exit 0
pip install --trusted-host=host     parses; --bogus-host=x is rejected
usermod --password=x <nouser>       "user does not exist" — parsed
usermod --bogus-opt=x <nouser>      "unrecognized option"
```

The last pair is the decisive one for `usermod`, which cannot be run for real: the option is accepted far enough to reach the user lookup, while an invented one is refused outright.

## Gates

Corpus drift zero — no pinned file writes these options inline, so this adds capability without changing a single finding on the corpus. The new cases are non-vacuous: seven subtests fail against the previous code. Parser sweep 402/0; fix-corpus sweep clean under the zsh oracle; suite, `golangci-lint` (0 issues), `gocyclo` green.

Closes the sweep opened in #1455, #1456 and #1457. `sudo --preserve-env=list` stays out: it names the variables to keep rather than preserving the whole environment, so it is a different flag from bare `-E` and needs its own kata treatment.

### DIFF
--- a/pkg/katas/katatests/zc1100s_test.go
+++ b/pkg/katas/katatests/zc1100s_test.go
@@ -3262,6 +3262,18 @@ func TestZC1184(t *testing.T) {
 			expected: []katas.Violation{},
 		},
 		{
+			name:  "invalid diff --unified=3",
+			input: `diff --unified=3 old.txt new.txt`,
+			expected: []katas.Violation{
+				{
+					KataID:  "ZC1184",
+					Message: "Consider `git diff` instead of `diff -u` when working in a repository. `git diff` provides better context and integration.",
+					Line:    1,
+					Column:  1,
+				},
+			},
+		},
+		{
 			name:  "invalid diff -u",
 			input: `diff -u old.txt new.txt`,
 			expected: []katas.Violation{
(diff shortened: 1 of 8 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
Approved on review.

PR #1459, opened 2026-08-26
chore: release v1.7.3

afadesigns:
Patch release. Fourteen commits since v1.7.2, all fixes and CI work — no new katas, no feature surface, so the patch digit moves.

## What is in it

The bulk is the auto-fix corruption family, closed end to end. Applying every fix across the pinned corpora used to leave eleven files that Zsh itself rejects; it now leaves none, and the sweep proves it by asking `zsh -n` rather than trusting this project's own parser.

The root cause was the safety net. When a full pass raised the parser-error count, the fallback retried edits one at a time and kept whichever parsed alone — which tore a multi-edit rewrite apart and could turn `[[ $x -eq 0 ]]` into `[[ $x == 0 ]]`: valid Zsh, different meaning, invisible to every gate. Edits from one violation now apply as a unit or not at all.

The rest is detection accuracy: ZC1053 stops reporting greps that are already silenced, ZC1044 reports a `cd` only where a failed one can do harm, a `( … )` in command position parses as a subshell so katas can see inside it, and sixteen katas learned to read `--flag=value`.

## Version surfaces

- `pkg/version/version.go` → `1.7.3`
- README release badge → `v1.7.3`
- `CHANGELOG.md` → `[1.7.3] - 2026-08-26`

Tag follows the merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.7.3] - 2026-08-26
+
+### Fixed
+- Auto-fix no longer rewrites a file into code Zsh cannot parse. A violation's edits now apply as one unit, so a rewrite that only partly succeeds is dropped whole rather than left half-applied. Applying every fix across the pinned corpora previously broke eleven files; it now breaks none.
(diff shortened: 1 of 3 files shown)


redteamx (approved):
Approved on review.

redteamx (approved):
Approved on review.

